PR #39, opened 2026-04-07
fix: align hybrid search vector scores via evidence chain to prevent recall regression

rnwolfe:
## Diagnosis

Hybrid search with a real AI provider (Gemini) was degrading entity Recall@5 from **0.82 → 0.30**. Two root causes:

**Problem 1: Vector-only episode union displaced entities**

During ingest, embeddings are generated for **episodes** (git commit messages), not entities. `findSimilar()` therefore returns episode IDs. The previous code unioned these semantically-similar episodes directly into the result set. Since the benchmark evaluates entity recall (top-5 entities), having episodes flood the top-5 directly displaced relevant entities.

**Problem 2: FTS weight cut in half in hybrid mode for no gain**

`WEIGHTS_HYBRID` dropped FTS weight from 0.4 → 0.25. But since entities have no direct embeddings, their `vector_score` was always 0.0 in hybrid mode — so hybrid mode just penalised entity FTS scores with nothing in return.

### `scoring.ts` — Rebalance WEIGHTS_HYBRID

- FTS: 0.25 → 0.35 (restore strength for entities)
- vector: 0.25 → 0.15 (vector is a reranking boost, not a replacement)
- evidence/temporal/graph: unchanged (sum = 1.0)

### `search.ts` — Evidence-chain vector score translation

Instead of looking up entity IDs directly in the vector score map (which always misses since embeddings are on episodes), compute each entity's vector score as the **max cosine similarity of any linked episode** in the similarity set:

```ts
// For each FTS entity result:
let vectorScore = 0.0;
for (const epId of provenance) {
  const s = episodeVectorScores.get(epId) ?? 0;
  if (s > vectorScore) vectorScore = s;
}
```

### `search.ts` — Remove vector-only union block

Deleted the ~120-line block that unions vector-only matches into the result set. Vector signal should only re-rank FTS results, not inject new candidates.

## Benchmark Results (Gemini, fastify v4.28.1, 500 episodes / 571 entities)

| Strategy | Recall@5 | MRR | vs vcs-only |
|----------|----------|-----|-------------|
| grep-baseline | 0.97 | 1.00 | +15.0pp |
| vcs-only | 0.82 | 1.00 | (baseline) |
| **ai-enhanced (after)** | **0.85** | **1.00** | **+2.5pp** |

Before this fix, `ai-enhanced` was scoring ~0.30 Recall@5 with a real AI provider configured. The fix achieves:
- No regression: ai-enhanced >= vcs-only baseline (0.85 >= 0.82) ✓
- Improvement: ai-enhanced beats vcs-only by +2.5pp ✓

### DIFF
--- a/packages/engram-core/src/retrieval/scoring.ts
+++ b/packages/engram-core/src/retrieval/scoring.ts
@@ -25,11 +25,11 @@ const WEIGHTS_FULLTEXT = {
 };
 
 const WEIGHTS_HYBRID = {
-  fts: 0.25,
+  fts: 0.35,
   evidence: 0.25,
   temporal: 0.15,
   graph: 0.1,
-  vector: 0.25,
+  vector: 0.15,
 };
 
 /**
--- a/packages/engram-core/src/retrieval/scoring.ts
+++ b/packages/engram-core/src/retrieval/scoring.ts
@@ -9,7 +9,7 @@ export interface ScoreComponents {
   graph_score: number; // 1-hop neighbor count, normalized
   temporal_score: number; // recency decay (0-1)
   evidence_score: number; // evidence strength (0-1)
-  vector_score: number; // always 0.0 (deferred)
+  vector_score: number; // cosine similarity via episode evidence chain (hybrid mode only)
 }
 
 /**
--- a/packages/engram-core/src/retrieval/search.ts
+++ b/packages/engram-core/src/retrieval/search.ts
@@ -301,9 +301,13 @@ export async function search(
     episodeRows.map((r) => r.rank),
   );
 
-  // Build vector similarity map when provider is set and produces a valid embedding
-  const vectorScoreMap = new Map<string, number>();
-  let vectorSimilarResults: Array<{ target_id: string; score: number }> = [];
+  // Build vector similarity map when provider is set and produces a valid embedding.
+  // episodeVectorScores maps episode IDs to cosine similarity scores.
+  // findSimilar returns episode IDs (embeddings are generated for episodes, not entities).
+  // Entity vector scores are derived via the evidence chain:
+  //   an entity's vector score = max cosine score of any linked episode in the similarity set.
+  // The same map is used directly for edge and episode lookups (they are also keyed by episode ID).
+  const episodeVectorScores = new Map<string, number>();
   let effectiveMode = mode;
   if (opts.provider) {
     try {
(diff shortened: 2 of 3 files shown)


packages/engram-core/src/retrieval/scoring.ts
```diff
@@ -25,11 +25,11 @@ const WEIGHTS_FULLTEXT = {
 };
```
rnwolfe, line 25: **[Non-blocking] Stale interface comment**

The `ScoreComponents` interface still documents `vector_score` as `// always 0.0 (deferred)`. After this PR, entities in hybrid mode can receive a non-zero `vector_score` via the evidence-chain translation. This comment should be updated to something like `// cosine similarity of best-matching backing episode (0.0 in fulltext mode)`.

packages/engram-core/src/retrieval/search.ts
```diff
@@ -327,6 +327,15 @@ export async function search(
 
   const results: SearchResult[] = [];
 
+  // Build episodeId -> vectorScore map for evidence-chain-based entity boosting.
+  // findSimilar returns episode IDs (embeddings are generated for episodes, not entities).
+  // We translate episode vector scores to entity scores via the evidence chain:
+  // an entity's vector score = max cosine score of any linked episode in the similarity set.
+  const episodeVectorScores = new Map<string, number>();
+  for (const r of vectorSimilarResults) {
```
rnwolfe: **[Non-blocking] Redundant `vectorScoreMap` still populated but unused for entities**

`vectorScoreMap` (built at ~line 319) is keyed on `target_id` from `findSimilar`, which returns episode IDs. It is still correctly used for edges (line 384) and episodes (line 412), where `vectorScoreMap.get(row.id)` looks up by episode ID directly. For entities the old lookup has been correctly replaced by the new `episodeVectorScores` path.

However the two maps (`vectorScoreMap` and `episodeVectorScores`) are built from the same `vectorSimilarResults` array and contain identical data. Consider collapsing them into one map (`episodeVectorScores`) and using it uniformly for edges, episodes, and entities — this removes the duplication and makes the data-flow easier to reason about.

rnwolfe (commented):
Code review by forge-loop sub-agent.

**Summary of findings** (2 non-blocking, 1 informational):

1. Weights sum to 1.00 — confirmed correct.
2. Evidence-chain max-episode approach is sound; empty-provenance and zero-score edge cases both resolve safely to 0.0.
3. Union block removal is clean; no dangling references.

See inline comments for details.

packages/engram-core/src/retrieval/scoring.ts
rnwolfe: Fixed: updated the `vector_score` comment to `// cosine similarity via episode evidence chain (hybrid mode only)`.

packages/engram-core/src/retrieval/search.ts
rnwolfe: Fixed: collapsed `vectorScoreMap` into `episodeVectorScores`. The single map is now built once from the `findSimilar` output and used uniformly for entity evidence-chain lookups, edge results, and episode results — no duplication.

PR #59, opened 2026-04-08
feat(viz): filter sidebar and search-to-focus

### LINKED ISSUE #50: feat(viz): filter sidebar + search-to-focus
## Summary

Phase 4 of graph visualization. Adds the left sidebar with filter controls (entity type, relation type, edge kind, decay status) and a top search input that pans the viewport to matching entities. Filters are client-side over the already-loaded graph; search hits the `/api/search` endpoint.

Spec: `docs/internal/specs/graph-visualization.md` (Phase 4)

Depends on: #48 (graph rendering)

### New API endpoint

- `GET /api/search?q=<text>` — wraps `search()` from engram-core. Returns matching entity IDs with scores.

### Frontend filters (`packages/engram-web/ui/filters.ts`)

- Left sidebar, collapsible, sticky.
- **Entity type filter**: checkbox list of all entity types present in the loaded graph. Default all checked.
- **Relation type filter**: checkbox list of all relation types. Default all checked.
- **Edge kind filter**: three checkboxes — observed / inferred / asserted. Default all checked.
- **Decay status filter**: checkboxes for `critical` / `elevated` / `stale` / `all`. Uses `/api/decay` data when decay overlay is on (issue #F); otherwise disabled.
- **"Hide orphans" toggle**: hides nodes with zero visible edges after filtering.
- Filter changes apply instantly — client-side set toggling on cytoscape nodes/edges, no server round-trip.

### Search input (`packages/engram-web/ui/search.ts`)

- Top-center or top-bar search input, placeholder "Search entities…".
- Keyboard: `/` key focuses the input (vim-style).
- Typing calls `/api/search?q=...` with a 150ms debounce.
- Dropdown results list below the input showing top 10 matches with entity type badges.
- Clicking a result or pressing Enter:
  - Pans and zooms the viewport to the match.
  - Highlights the node (pulse animation).
  - Opens its entity detail panel (from #49).
- `Escape` closes the dropdown.

### Filter + search interaction

- Filters apply on top of search results (search pans to the best match even if it's filtered out — in that case, highlight it distinctly and show a "remove filter to see this entity" hint).

## Acceptance Criteria

- [ ] `GET /api/search?q=...` returns ranked matches.
- [ ] Entity type filter toggles visible nodes correctly.
- [ ] Relation type filter toggles visible edges correctly.
- [ ] Edge kind filter toggles by observed/inferred/asserted.
- [ ] "Hide orphans" toggle works after other filters are applied.
- [ ] `/` key focuses the search input.
- [ ] Typing in the search input shows a ranked dropdown with ≤10 results.
- [ ] Enter / click selects a result → pans + highlights + opens detail panel.
- [ ] Escape closes the dropdown and clears focus.
- [ ] Filter changes feel instant (<50ms) on Fastify-scale.
- [ ] Unit tests for the search API handler.

## Out of Scope

- Saved filter presets — future.
- Full-text search inside episode content via the UI — the API endpoint supports it, but presentation is keyword-only in v1.
- Regex search — v1 is substring only.

rnwolfe:
## Summary

- Adds `GET /api/search?q=<text>` endpoint wrapping `engram-core` `search()`, returning top 10 entity matches with `id`, `canonical_name`, `entity_type`, `score`
- Adds `ui/filters.ts`: left sidebar with checkboxes for `entity_type`, `relation_type`, `edge_kind`, and a "Hide orphans" toggle — all client-side via `cy.elements().show()/hide()`
- Adds `ui/search.ts`: search input with 150ms debounce, dropdown of ≤10 results, pan+zoom to node on select, pulse highlight, and `openEntityPanel(id)`. Shows toast if result is filtered out. `/` focuses input, `Escape` closes dropdown

## Acceptance Criteria

- [x] `GET /api/search?q=...` returns ranked entity matches (top 10)
- [x] `GET /api/search` with empty `q` returns empty results
- [x] Filter sidebar with checkboxes for `entity_type`, `relation_type`, `edge_kind`
- [x] "Hide orphans" toggle hides nodes with no visible edges after any filter change
- [x] Filter changes are instant — `cy.elements().show()/hide()`
- [x] Search input in header with 150ms debounce
- [x] Dropdown ≤10 results with click/Enter to navigate
- [x] Pan+zoom: `cy.animate({ fit: { eles: node, padding: 100 } })`
- [x] Pulse CSS class applied and removed after 1.2s
- [x] `/` key focuses search (when not in an input element)
- [x] `Escape` closes dropdown
- [x] Toast shown when search result is hidden by filters
- [x] Tests: `GET /api/search?q=...` returns ranked matches; empty `q` returns empty results

## Test plan

- [ ] Run `bun test` — all 454 tests pass
- [ ] Run `bun run lint` — no errors
- [ ] Run `bun run build` — builds cleanly

Closes #50


rnwolfe (commented):
## Code Review — feat(viz): filter sidebar and search-to-focus

Overall: solid implementation that covers all acceptance criteria. Six issues need addressing before merge, with two blocking.

---

### BLOCKING

**1. Dist files committed to source control**

`packages/engram-web/dist/` is in the root `.gitignore` (`dist/`) yet four build artifacts are included in this PR:
- `packages/engram-web/dist/index.js` (+37,327 lines)
- `packages/engram-web/dist/ui/index.html`
- `packages/engram-web/dist/ui/main.css`
- `packages/engram-web/dist/ui/main.js`

Build artifacts must not be committed. Remove these files and ensure the package's `.gitignore` or the root one is respected. If the web package needs a checked-in build for serving static assets, that design decision should be explicit and documented — but the current approach silently violates the existing `.gitignore`.

**2. `handleSearch` accesses `graph.db` directly (internal API violation)**

In `packages/engram-web/src/api/search.ts` (lines 45–63), the handler runs a raw SQL query against `graph.db` to enrich `entity_type`:

```ts
const rows = graph.db
  .query<{ id: string; canonical_name: string; entity_type: string }, string[]>(
    `SELECT id, canonical_name, entity_type FROM entities WHERE id IN (${placeholders})`,
  )
  .all(...ids);
```

Per the project architecture, `engram-core` is the product and `engram-web` must never reach through to the internal DB layer. This couples the web package to the private schema and will silently break if the `entities` table is renamed or restructured. 

The correct fix is one of:
- Have `engram-core`'s `search()` return `entity_type` in its result objects (preferred — it should be a first-class field).
- Add a public `getEntityById(graph, id)` function to `engram-core` and call that.

Do not access `graph.db` from outside `engram-core`.

---

### NON-BLOCKING (should fix)

**3. Filter state not reset on re-initialization**

In `packages/engram-web/ui/filters.ts`, `activeFilters` is a module-level singleton:

```ts
const activeFilters = {
  entityTypes: new Set<string>(),
  relationTypes: new Set<string>(),
  edgeKinds: new Set<string>(),
  hideOrphans: false,
};
```

`buildCheckboxSection` adds values to the passed `activeSet` reference. If `initFilters` is called a second time (e.g., on graph data refresh), the old set values from the previous call are still present, causing phantom filter entries. The sets should be cleared at the top of `initFilters` before rebuilding:

```ts
activeFilters.entityTypes.clear();
activeFilters.relationTypes.clear();
activeFilters.edgeKinds.clear();
activeFilters.hideOrphans = false;
```

**4. Pulse animation is not actually animated**

The CSS defines `@keyframes pulse-ring` but the Cytoscape `.pulse` style in `ui/graph.ts` just applies a static border:

```ts
{ selector: ".pulse", style: { "border-width": 4, "border-color": "#58a6ff", "border-opacity": 1 } }
```

The keyframe animation is never referenced (`animation-name: pulse-ring` is absent from the Cytoscape style block). The node gets a static blue border for 1.2s then snaps off — not the pulsing ring implied by the keyframe. Either wire up the animation or rename the CSS comment to match the actual behavior.

**5. `search()` limit + post-filter can return fewer than expected results**

In `search.ts`:

```ts
const raw = await search(graph, query.trim(), { limit: 10, entity_types: undefined });
const results = raw.filter((r) => r.type === "entity").slice(0, 10);
```

If `search()` returns 10 mixed results (entities + edges + episodes), the post-filter leaves fewer than 10 entity results in the response. Since the intent is "top 10 entities", pass `entity_types` to restrict the search at the core level, or increase the limit to compensate. The `entity_types: undefined` on the options object has no effect — it's the same as omitting the option.

**6. Missing HTTP-level test for `/api/search`**

`test/search.test.ts` tests `handleSearch()` directly, which is good. However there is no test for the HTTP route in `server.ts` — specifically that:
- `/api/search` with a missing or empty `q` returns `{ results: [] }` with status 200
- `/api/search?q=...` returns the correct JSON shape
- Errors in the handler return status 500 with `{ error: "..." }`

The server route changed from sync to `async` in this PR; route-level tests would catch any regression there.

---

### Notes (no action required)

- The `/` keybind correctly guards against firing when focus is in `HTMLInputElement` or `HTMLTextAreaElement` — good defensive check.
- `textContent` is used throughout `renderDropdown` and `showToast` — no XSS risk.
- 150ms debounce is reasonable; filter changes via `cy.elements().show()/hide()` are instant as required.
- `openEntityPanel` is explicitly a stub for now; acceptable for v0.
- `debounceTimer` and `currentResults` as module-level state in `search.ts` would break if `initSearch` were called multiple times, but given the single-SPA nature of this UI that's not a practical concern currently.

rnwolfe (commented):
Fixed:
- **dist artifact removed**: `packages/engram-web/dist/index.js` (Node.js server build output) removed from git tracking via `git rm --cached`. The `dist/ui/` files (pre-built frontend bundle) remain intentionally committed per project convention.
- **Public API used in search handler**: Replaced `graph.db.prepare(...)` raw SQL in `handleSearch` with `getEntity(graph, id)` from engram-core public API. Each entity result is now fetched via the public API to obtain `canonical_name` and `entity_type`.

All 454 tests pass, build succeeds, lint clean.

PR #78, opened 2026-04-10
feat(format): schema migration v0.1 → v0.2 — projection layer tables

### LINKED ISSUE #66: feat(format): schema migration v0.1 → v0.2 — projection layer tables
## Context

ADR-002 introduces an AI-authored projection layer on top of the existing substrate. The first piece of work is purely additive schema: three new tables, one FTS5 virtual table, and the migration runner that applies them.

This issue is the **foundation** for all subsequent projection work. Every other implementation issue depends on this one shipping.

### 1. Schema constants in \`packages/engram-core/src/format/schema.ts\`

Add the following DDL constants (full SQL is in [\`docs/internal/specs/format-v0.2.md\`](../blob/main/docs/internal/specs/format-v0.2.md)):

- \`CREATE_PROJECTIONS\` — the \`projections\` table
- \`CREATE_PROJECTIONS_INDEXES\` — five indexes including the partial unique index for \`(anchor, kind)\` on active projections
- \`CREATE_PROJECTION_EVIDENCE\` — the polymorphic evidence table
- \`CREATE_PROJECTION_EVIDENCE_INDEXES\`
- \`CREATE_RECONCILIATION_RUNS\`
- \`CREATE_PROJECTIONS_FTS\` — virtual FTS5 table
- \`CREATE_PROJECTIONS_FTS_TRIGGERS\` — three triggers (ai/ad/au) analogous to \`entities_fts\` triggers

These are **appended** to \`SCHEMA_DDL\` for fresh-database creation. No existing constants change.

### 2. Migration step in \`packages/engram-core/src/format/migrations.ts\`

Add \`migrate_0_1_0_to_0_2_0\` that runs the new DDL constants in order, then \`UPDATE metadata SET value = '0.2.0' WHERE key = 'schema_version'\`.

### 3. Version bump

\`packages/engram-core/src/format/version.ts\`:
- Bump current schema version to \`'0.2.0'\`
- Set \`MIN_READABLE_VERSION='0.1.0'\` and \`MIN_WRITABLE_VERSION='0.2.0'\` per the format spec

### 4. Tests

In \`packages/engram-core/test/format/\` (or wherever schema tests live):

- A test that creates a fresh in-memory database, applies all DDL, and verifies all new tables exist with the expected columns
- A test that opens a v0.1 fixture file and runs the migration, verifying schema_version becomes \`0.2.0\` and the new tables exist
- A test that confirms a v0.2 reader on a v0.1 file works (projection tables are empty, no errors)
- A test that confirms a v0.1 fixture round-trips through migration without losing any v0.1 data
- An integration test that inserts a row into \`projections\` + \`projection_evidence\` (raw SQL, no operations layer yet) and reads it back

## Acceptance criteria

- [ ] All DDL constants added to \`schema.ts\` and applied in \`SCHEMA_DDL\` in the order specified by format-v0.2
- [ ] Migration runner applies the new DDL on existing v0.1 files and bumps \`schema_version\` to \`'0.2.0'\`
- [ ] All new tables, indexes, virtual table, and triggers exist in a fresh v0.2 database
- [ ] Tests cover fresh-create, migration-from-v0.1, v0.1-data-preservation, and basic raw insert/read
- [ ] \`bun run lint\` passes
- [ ] \`bun test\` passes

## Out of scope

- The \`project()\`, \`reconcile()\`, \`getProjection()\` operations — separate issues, all blocked on this one
- The verify() invariant additions — separate issue
- Frontmatter validation — separate issue (lives in the operations layer)

## Blocked by

- #31 (AI provider integration) — not a hard block, but the projection layer's value is gated on the provider being available; the migration alone is independently useful and can ship before #31

## References

- [\`docs/internal/specs/format-v0.2.md\`](../blob/main/docs/internal/specs/format-v0.2.md) — full DDL with comments
- [\`docs/internal/specs/projections.md\`](../blob/main/docs/internal/specs/projections.md) — design rationale

rnwolfe:
## Summary

- Adds all v0.2 DDL constants to `schema.ts`: `projections`, `projection_evidence`, `reconciliation_runs` tables, 5 indexes (including partial unique), FTS5 virtual table `projections_fts`, and 3 triggers (ai/ad/au)
- Adds `migrate_0_1_0_to_0_2_0` to the new `migrations.ts` — purely additive, runs all new DDL then bumps `format_version` to `0.2.0` atomically in a transaction
- Bumps `FORMAT_VERSION` to `0.2.0`, adds `MIN_READABLE_VERSION='0.1.0'` and `MIN_WRITABLE_VERSION='0.2.0'` in `version.ts`
- Updates `openGraph` and `verifyGraph` to use semver range checking (`MIN_READABLE_VERSION`..`FORMAT_VERSION`) so v0.2 engine opens v0.1 files

## Acceptance Criteria

- [x] All DDL constants added to `schema.ts` and applied in `SCHEMA_DDL`
- [x] `migrate_0_1_0_to_0_2_0` applies new DDL on existing v0.1 databases and bumps `schema_version` to `0.2.0`
- [x] All new tables, indexes, virtual table, and triggers exist in a fresh v0.2 database
- [x] Tests cover: fresh-create, migration-from-v0.1, v0.1-data-preservation, compatibility (openGraph on migrated file), raw insert/read
- [x] `bun test packages/engram-core/` passes (344 tests, 0 failures)
- [x] `bunx biome check` passes on all changed files

## Test plan

- `bun test packages/engram-core/test/format/schema-v0.2.test.ts` — 21 tests covering all acceptance criteria
- `bun test packages/engram-core/` — full 344-test suite passes

Closes #66

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -2,7 +2,13 @@
   "$schema": "https://biomejs.dev/schemas/2.4.10/schema.json",
   "assist": { "actions": { "source": { "organizeImports": "on" } } },
   "files": {
-    "includes": ["**/*.ts", "**/*.json", "!**/dist", "!**/node_modules"]
+    "includes": [
+      "**/*.ts",
+      "**/*.json",
+      "!**/dist",
+      "!**/node_modules",
+      "!.claude/worktrees"
+    ]
   },
   "linter": {
     "enabled": true,
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.4.10/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.4.11/schema.json",
   "assist": { "actions": { "source": { "organizeImports": "on" } } },
   "files": {
     "includes": [
--- a/packages/engram-core/src/format/graph.ts
+++ b/packages/engram-core/src/format/graph.ts
@@ -9,7 +9,11 @@
 import { Database } from "bun:sqlite";
 import { ulid } from "ulid";
 import { SCHEMA_DDL } from "./schema.js";
-import { ENGINE_VERSION, FORMAT_VERSION } from "./version.js";
+import {
+  ENGINE_VERSION,
+  FORMAT_VERSION,
+  MIN_READABLE_VERSION,
+} from "./version.js";
 
 export interface EngramGraph {
   db: Database;
@@ -30,6 +34,31 @@ export class EngramFormatError extends Error {
     super(message);
     this.name = "EngramFormatError";
   }
+}
+
+/**
+ * Compares two semver strings. Returns negative if a < b, 0 if equal, positive if a > b.
+ * Only handles simple MAJOR.MINOR.PATCH format.
+ */
+function compareSemver(a: string, b: string): number {
+  const pa = a.split(".").map(Number);
+  const pb = b.split(".").map(Number);
+  for (let i = 0; i < 3; i++) {
+    const diff = (pa[i] ?? 0) - (pb[i] ?? 0);
+    if (diff !== 0) return diff;
+  }
+  return 0;
+}
+
+/**
+ * Returns true if the given format version can be opened by this engine.
+ * The readable range is [MIN_READABLE_VERSION, FORMAT_VERSION].
+ */
+function isVersionReadable(version: string): boolean {
+  return (
+    compareSemver(version, MIN_READABLE_VERSION) >= 0 &&
+    compareSemver(version, FORMAT_VERSION) <= 0
+  );
 }
 
 function applyPragmas(db: Database): void {
@@ -115,10 +144,11 @@ export function openGraph(path: string): EngramGraph {
     );
   }
 
-  if (formatVersion !== FORMAT_VERSION) {
+  // Accept any version between MIN_READABLE_VERSION and FORMAT_VERSION (inclusive).
+  if (!isVersionReadable(formatVersion)) {
     db.close();
     throw new EngramFormatError(
-      `openGraph: unsupported format_version '${formatVersion}' (expected '${FORMAT_VERSION}'): ${path}`,
+      `openGraph: unsupported format_version '${formatVersion}' (readable range: ${MIN_READABLE_VERSION}–${FORMAT_VERSION}): ${path}`,
     );
   }
 
--- a/packages/engram-core/src/format/index.ts
+++ b/packages/engram-core/src/format/index.ts
@@ -9,6 +9,7 @@ export {
   EngramFormatError,
   openGraph,
 } from "./graph.js";
+export { migrate_0_1_0_to_0_2_0 } from "./migrations.js";
 export { SCHEMA_DDL } from "./schema.js";
 export type { VerifyResult, Violation, ViolationSeverity } from "./verify.js";
 export { verifyGraph } from "./verify.js";
--- a/packages/engram-core/src/format/migrations.ts
+++ b/packages/engram-core/src/format/migrations.ts
@@ -1,0 +1,53 @@
+/**
+ * migrations.ts — schema migration steps for the .engram format.
+ *
+ * Each migration function accepts a Database, applies the DDL for the new
+ * version, and updates schema_version (stored as 'format_version' in metadata).
+ */
+
+import type { Database } from "bun:sqlite";
+import {
+  CREATE_PROJECTION_EVIDENCE,
+  CREATE_PROJECTION_EVIDENCE_INDEXES,
+  CREATE_PROJECTIONS,
+  CREATE_PROJECTIONS_FTS,
+  CREATE_PROJECTIONS_FTS_TRIGGERS,
+  CREATE_PROJECTIONS_INDEXES,
+  CREATE_RECONCILIATION_RUNS,
+} from "./schema.js";
+
+/**
+ * Migrates a v0.1.0 database to v0.2.0.
+ *
+ * This migration is purely additive: it appends the projection-layer tables,
+ * indexes, FTS virtual table, and triggers. No existing tables are altered and
+ * no data is backfilled.
+ *
+ * DDL application order matches the spec (format-v0.2.md § DDL application order):
+ *  1. CREATE TABLE projections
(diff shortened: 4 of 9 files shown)


rnwolfe (commented):
## Code Review — feat(format): schema migration v0.1 → v0.2

Overall the approach is sound — the migration is purely additive, atomically wrapped in a transaction, and the semver range check is a clean generalisation. The tests use real SQLite (no mocks), cover the key invariants, and are well-structured. That said, there are a few issues that need attention before merge, plus some lower-severity observations.

---

### Blocking

#### 1. CI is failing — import-order lint error in `packages/engram-core/src/index.ts`

The `test` job fails at the Lint step. Biome's `organizeImports` rule flags that the new version-constants export block was placed **before** the AI-provider exports, which Biome wants sorted differently. The fix is to run `bunx biome check --write .` (or manually reorder the block so the version re-exports come after `./format/index.js` per Biome's order). The PR description claims `bunx biome check` passes — that may have been against a different Biome version; CI is running 2.4.11 and the schema pin is 2.4.10, which also causes a warning.

`packages/engram-core/src/index.ts`, lines 7–12.

---

#### 2. `migrate_0_1_0_to_0_2_0` silently succeeds when called on an already-migrated database

```ts
// migrations.ts line 47-52
db.transaction(() => {
  for (const ddl of steps) {
    db.exec(ddl);   // will throw "table projections already exists"
  }
  db.run("UPDATE metadata SET value = '0.2.0' WHERE key = 'format_version'");
})();
```

If called twice (e.g. an operator retries after a partial failure or calls it on a v0.2 file), `db.exec` will throw `SqliteError: table projections already exists`. The transaction will roll back, leaving the database in its prior state, but the **error message from SQLite is cryptic** — the caller has no idea why. Consider either:

- Checking `format_version` at the top of the function and returning early if it's already `>= 0.2.0`, or
- Using `CREATE TABLE IF NOT EXISTS` / `CREATE INDEX IF NOT EXISTS` throughout the DDL constants.

This is especially important because `migrate_0_1_0_to_0_2_0` is now a public API (`export`ed from `src/index.ts`) and callers will reasonably try to use it defensively.

---

#### 3. `projection_evidence` has no foreign-key constraint to `projections`, only to itself via `projection_id`

```sql
-- schema.ts lines 250-257
CREATE TABLE projection_evidence (
  projection_id TEXT NOT NULL REFERENCES projections(id),
  ...
);
```

Wait — this one _is_ present. Ignore this point; the FK is there.

---

### Medium severity

#### 4. `compareSemver` is duplicated between `graph.ts` and `verify.ts`

The identical 8-line `compareSemver` function is copy-pasted into two files. This isn't a correctness bug today, but when the version logic evolves they will inevitably diverge. Extract it to `version.ts` (or a small `semver.ts` utility) and import from both call sites.

Files: `packages/engram-core/src/format/graph.ts` lines 43–51, `packages/engram-core/src/format/verify.ts` lines 11–19.

---

#### 5. `MIN_WRITABLE_VERSION` is exported but never enforced

`version.ts` documents: _"Opening a v0.1 file for writes must first migrate it to v0.2."_ `MIN_WRITABLE_VERSION` is exported from the public surface, but `openGraph` does not enforce it — there is no write-guard anywhere. If a caller opens a v0.1 file and calls `addEntity`, the insert will succeed but `projections` / `projection_evidence` won't exist, which breaks any later call that touches projection tables.

Either:
- Enforce the write guard inside `openGraph` (or a new `openGraphForWrite` variant), or
- Document explicitly that `MIN_WRITABLE_VERSION` is _advisory_ and enforcement is deferred to a future PR (and note the issue number).

As-is the constant communicates an invariant that isn't actually upheld.

---

#### 6. `createV1Db` in tests uses string interpolation for SQL values

```ts
// test line 104
db.run(`INSERT INTO metadata (key, value) VALUES ('created_at', '${now}')`);
```

The project convention (and `graph.ts` itself) uses parameterised statements for all writes. This is test-only code so the injection risk is negligible, but it sets a bad example and should use `.run("...", [now])` or a prepared statement to stay consistent with the rest of the codebase.

---

### Low severity / nits

#### 7. `projection_evidence` has no FK back to any referenced entity/edge/episode

`target_type` + `target_id` is a polymorphic reference (entity, edge, episode, or projection). SQLite can't enforce polymorphic FKs declaratively, so this is expected — but `verifyGraph` does not include a corresponding integrity check for dangling `projection_evidence` rows. The existing check for `entity_evidence` / `edge_evidence` episode refs (`checkEvidenceEpisodeRefs`) has no counterpart for the new table. Worth adding before the retrieval layer ships.

#### 8. Test file is 699 lines — at the project's 500-line limit

`packages/engram-core/test/format/schema-v0.2.test.ts` is 699 lines. The CLAUDE.md convention is "Keep files under 500 lines." The five `describe` blocks map cleanly onto separate files (`schema-v0.2-fresh.test.ts`, `schema-v0.2-migration.test.ts`, etc.). Not blocking for a schema-only PR, but worth splitting before the next PR lands on top of this.

#### 9. `phases` default in `reconciliation_runs` is a comma-separated string

```sql
phases TEXT NOT NULL DEFAULT 'assess,discover',
```

Using a plain `TEXT` column for a structured list means any code reading this column must know to split on commas. If the set of phases is expected to grow (e.g. `prune`, `backfill`), consider storing JSON (array) instead so no parsing convention needs to be documented separately. Minor, but worth deciding before v0.2 data accumulates in the wild.

---

### Summary

| Severity | Count |
|----------|-------|
| Blocking | 2 (CI lint failure; silent re-run failure of migration function) |
| Medium   | 2 (duplicated compareSemver; unenforced MIN_WRITABLE_VERSION) |
| Low/nit  | 3 |

The CI lint failure must be fixed before merge. The migration idempotency issue (#2) is also strongly recommended — it affects a public API. Everything else can be addressed in follow-up PRs if preferred.

rnwolfe (commented):
Blocking issues addressed:
1. Biome lint: excluded `.claude/worktrees/**` from biome.json scan (worktree-local biome.json files caused nested-root conflicts), then auto-fixed import order with `bunx biome check --write`.
2. Migration idempotency: `migrate_0_1_0_to_0_2_0` now reads `format_version` from metadata first and returns early if already at `0.2.0`.

All 344 tests pass, lint clean.

PR #79, opened 2026-04-10
feat: add project() operation and ProjectionGenerator interface

### LINKED ISSUE #67: feat(core): project() operation — explicit projection authoring primitive
## Context

ADR-002 introduces an AI-authored projection layer. \`project()\` is the **explicit authoring primitive** — the deterministic escape hatch that the CLI \`engram project\` wraps and that the discover phase of \`reconcile()\` calls under the hood. There is exactly one authoring code path with two callers.

This issue implements the core operation in \`engram-core\`. It does NOT include the CLI command, the MCP tool, or the discover phase — those are separate issues.

### 1. The operation in \`packages/engram-core/src/graph/projections.ts\` (new file)

\`\`\`ts
project(db, opts: {
  kind: string;
  anchor: { type: AnchorType; id?: string };
  inputs: { type: 'episode'|'entity'|'edge'|'projection'; id: string }[];
  generator: ProjectionGenerator;  // wraps an AI provider + prompt template
}): Promise<Projection>
\`\`\`

Steps (per [\`projections.md\`](../blob/main/docs/internal/specs/projections.md) § Operations):

1. Resolve and read the input set from the substrate. Reject if any input is missing or redacted.
2. Compute \`input_fingerprint = sha256(sorted(\"type:id:content_hash\" entries))\`.
3. **Cycle check**: if any input is \`type='projection'\`, run a recursive CTE from the candidate input set to detect would-be cycles. Reject with \`ProjectionCycleError\` on hit.
4. Check whether an active projection already exists for \`(anchor, kind)\`. If yes and its \`input_fingerprint\` matches → return existing (idempotent no-op).
5. Otherwise call \`generator.generate(inputs)\` → markdown body with frontmatter.
6. **Validate frontmatter**: required keys are present (id, kind, anchor, title, model, prompt_template_id, prompt_hash, input_fingerprint, valid_from, valid_until, inputs). Normalize the inputs list to match what was actually used.
7. Insert \`projections\` row + \`projection_evidence\` rows in a single transaction.
8. If an existing active projection was present with a different fingerprint → call \`supersede()\` (helper, see below) atomically in the same transaction.

### 2. \`supersede()\` helper

Atomic transition in one transaction (full SQL in [\`projections.md\`](../blob/main/docs/internal/specs/projections.md) § Operation 3):

- \`UPDATE\` old projection: set \`invalidated_at = now\`, \`valid_until = now\`, \`superseded_by = new.id\`
- \`INSERT\` new projection row + evidence rows

### 3. \`ProjectionGenerator\` interface in \`packages/engram-core/src/ai/projection-generator.ts\` (new file)

\`\`\`ts
interface ProjectionGenerator {
  generate(inputs: ResolvedInput[]): Promise<{ body: string; confidence: number }>;
  assess(projection: Projection, currentInputs: ResolvedInput[]): Promise<AssessVerdict>;
  regenerate(projection: Projection, currentInputs: ResolvedInput[]): Promise<{ body: string; confidence: number }>;
}

type AssessVerdict =
  | { verdict: 'still_accurate' }
  | { verdict: 'needs_update'; reason: string }
  | { verdict: 'contradicted'; reason: string };
\`\`\`

The interface is implemented by concrete generators per kind (e.g., \`EntitySummaryGenerator\`). For this issue, ship a single \`NullGenerator\` that errors on \`generate()\` and an \`AnthropicGenerator\` stub that calls the AI provider with a placeholder prompt — concrete prompts are shipped by the kind catalog issue (#64) and the discover prompt experiment (#63). The \`NullGenerator\` is what makes \`engram project --kind X\` fail loudly when no AI provider is configured.

### 4. Frontmatter parser/validator in \`packages/engram-core/src/format/projection-frontmatter.ts\` (new file)

A small YAML frontmatter parser (use \`gray-matter\` or similar; check existing deps before adding) that:

- Extracts the frontmatter block from a markdown body
- Validates required keys
- Returns a typed \`ProjectionFrontmatter\` object
- Re-serializes a \`ProjectionFrontmatter\` to a markdown body for the export path (separate issue, but the helper lives here)

### 5. Typed errors in \`packages/engram-core/src/graph/errors.ts\`

- \`ProjectionCycleError\`
- \`ProjectionEvidenceMissingError\` (raised when an input doesn't resolve or is redacted)
- \`ProjectionFrontmatterInvalidError\`

### 6. Tests

Extensive — this is the operation that establishes the invariants. In \`packages/engram-core/test/graph/projections.test.ts\`:

- Create a projection with episode inputs only; verify row + evidence rows
- Create a projection with mixed inputs (episode + entity + edge); verify polymorphic evidence
- Create a projection with another projection as input; verify the projection-of-projection link
- Cycle rejection: try to create a projection that would cycle through another; assert \`ProjectionCycleError\`
- Idempotent re-author with same fingerprint: assert second call returns the existing row, no duplicate inserts
- Re-author with different fingerprint: assert the old projection is superseded atomically (old.invalidated_at and new.valid_from are equal)
- Frontmatter validation: missing required key → \`ProjectionFrontmatterInvalidError\`
- Evidence missing: input that doesn't exist in the substrate → \`ProjectionEvidenceMissingError\`

## Acceptance criteria

- [ ] \`project()\` operation implemented per spec
- [ ] \`supersede()\` helper implemented and atomic (verified by transaction-rollback test)
- [ ] \`ProjectionGenerator\` interface defined; \`NullGenerator\` and stub \`AnthropicGenerator\` implemented
- [ ] Frontmatter parser/validator implemented and tested
- [ ] Typed errors defined and raised at the right boundaries
- [ ] All tests above pass; \`bun test\` and \`bun run lint\` are green
- [ ] No regressions in existing tests

## Out of scope

- The CLI command \`engram project\` — separate issue (#TBD)
- The MCP tool — separate issue
- The discover phase of \`reconcile()\` — separate issue (depends on this one)
- The actual prompt content for kinds — depends on #64 (kind catalog)
- Read-time staleness check — separate issue (uses the same fingerprint mechanism)
- The verify() invariants — separate issue

## Blocked by

- #66 (schema migration) — needs the tables to exist
- #31 (AI provider integration) — \`AnthropicGenerator\` calls into the provider layer

## Why \`backlog/needs-refinement\` not \`backlog/ready\`

This issue is specced but its concrete shape will get clearer once #66 ships and once the discover prompt experiment (#63) starts producing real prompts. The interface is right; the details of the \`NullGenerator\` / \`AnthropicGenerator\` split may want adjustment. Refinement check before going \`backlog/ready\`.

## References

- [\`docs/internal/specs/projections.md\`](../blob/main/docs/internal/specs/projections.md) § Operations, § Resolved Decisions
- [\`docs/internal/specs/format-v0.2.md\`](../blob/main/docs/internal/specs/format-v0.2.md)

## Documentation

- [ ] JSDoc on `project()`, `supersede()`, and the `ProjectionGenerator` interface covering inputs, errors, and the `(anchor, kind, input_fingerprint)` idempotency contract
- [ ] JSDoc on the frontmatter parser covering required keys and the round-trip expectation for `engram export wiki`
- [ ] CLAUDE.md: add `graph/projections.ts`, `ai/projection-generator.ts`, and `format/projection-frontmatter.ts` to the Key Files table
- [ ] CLAUDE.md: under Architecture Patterns, note the "one authoring code path, two callers" invariant — `project()` is called both by the `engram project` CLI and by the `reconcile()` discover phase

rnwolfe:
## Summary

- Implements the `project()` authoring primitive from the projection layer spec (ADR-002, `docs/internal/specs/projections.md`)
- Adds `ProjectionGenerator` interface with `NullGenerator` and `AnthropicGenerator` stub in `packages/engram-core/src/ai/projection-generator.ts`
- Adds `project()`, `supersedeProjection()`, `getProjection()` and error types in `packages/engram-core/src/graph/projections.ts`
- Exports all new types and functions from the public API (`packages/engram-core/src/index.ts`)

## Key behaviors

- **Idempotent**: returns the existing projection when `(anchor, kind, input_fingerprint)` match
- **Supersession**: atomically invalidates the old projection and inserts the new one when fingerprint differs (old-first ordering to satisfy the partial unique index)
- **Cycle detection**: recursive CTE walks downward through projection dependencies; rejects with `ProjectionCycleError` if the would-be superseded projection is reachable
- **Read-time stale detection**: `getProjection()` recomputes the fingerprint on every read and returns `stale: true` with reason (`input_content_changed` | `input_deleted`) — always correct regardless of reconcile cadence
- **Frontmatter validation**: required YAML keys enforced at write time

## Test plan

- [x] `NullGenerator.generate()` throws with a helpful error message
- [x] `project()` with `NullGenerator` throws before touching the DB
- [x] `project()` creates projection + evidence rows
- [x] `project()` is idempotent when fingerprint matches
- [x] `project()` supersedes when fingerprint changes (invalidates old, links `superseded_by`)
- [x] `project()` rejects missing or redacted inputs
- [x] `project()` rejects invalid frontmatter
- [x] `project()` detects projection DAG cycles
- [x] `supersedeProjection()` sets `invalidated_at`, `valid_until`, `superseded_by` correctly
- [x] `supersedeProjection()` throws for nonexistent or already-invalidated projections
- [x] `getProjection()` returns `stale: false` when inputs are unchanged
- [x] `getProjection()` returns `stale: true, stale_reason: input_content_changed` after episode update
- [x] `getProjection()` returns `stale: true, stale_reason: input_deleted` after episode redaction
- [x] All 368 `engram-core` tests pass

Closes #67

### DIFF
--- a/packages/engram-core/src/ai/projection-generator.ts
+++ b/packages/engram-core/src/ai/projection-generator.ts
@@ -1,0 +1,200 @@
+/**
+ * projection-generator.ts — ProjectionGenerator interface and implementations.
+ *
+ * A ProjectionGenerator wraps an AI provider and prompt template to produce
+ * projection bodies. It is the AI boundary for the projection authoring layer.
+ *
+ * Implementations:
+ * - NullGenerator: throws on generate() — used when no AI is configured.
+ * - AnthropicGenerator: stub that calls the Anthropic API with a placeholder prompt.
+ */
+
+import type { Projection } from "../graph/projections.js";
+
+// ─── Types ──────────────────────────────────────────────────────────────────
+
+/**
+ * A substrate element resolved from the database, ready to pass to an AI generator.
+ */
+export interface ResolvedInput {
+  type: "episode" | "entity" | "edge" | "projection";
+  id: string;
+  /** The content of the substrate element at resolution time. */
+  content: string | null;
+  /** SHA-256 hash of the content at resolution time. */
+  content_hash: string | null;
+}
+
+/**
+ * The verdict returned by generator.assess() during a reconcile() run.
+ */
+export type AssessVerdict =
+  | { verdict: "still_accurate" }
+  | { verdict: "needs_update"; reason: string }
+  | { verdict: "contradicted"; reason: string };
+
+/**
+ * Core interface for projection generation.
+ *
+ * Implementations wrap an AI provider plus a prompt template and handle:
+ * - generate(): produce the initial markdown body from resolved inputs.
+ * - assess(): determine whether an existing projection is still accurate
+ *   given the current (possibly changed) input state.
+ * - regenerate(): produce a revised body for an existing projection given
+ *   updated inputs.
+ */
+export interface ProjectionGenerator {
+  /**
+   * Generate a markdown body (with YAML frontmatter) from resolved inputs.
+   *
+   * The returned body MUST include all required frontmatter keys:
+   * id, kind, anchor, title, model, input_fingerprint, valid_from, inputs.
+   *
+   * @throws Error if generation fails or is not supported.
+   */
+  generate(
+    inputs: ResolvedInput[],
+  ): Promise<{ body: string; confidence: number }>;
+
+  /**
+   * Assess whether an existing projection is still accurate given the
+   * current state of its inputs.
+   *
+   * Called during reconcile() assess phase for projections whose
+   * input_fingerprint has drifted.
+   */
+  assess(
+    projection: Projection,
+    currentInputs: ResolvedInput[],
+  ): Promise<AssessVerdict>;
+
+  /**
+   * Regenerate a projection body based on updated inputs.
+   *
+   * Called during reconcile() when assess() returns 'needs_update' or
+   * 'contradicted'. The old projection is passed for context (e.g. to
+   * carry over parts of the body that haven't changed).
+   */
+  regenerate(
+    projection: Projection,
+    currentInputs: ResolvedInput[],
+  ): Promise<{ body: string; confidence: number }>;
+}
+
+// ─── NullGenerator ───────────────────────────────────────────────────────────
+
+/**
+ * NullGenerator: used when no AI provider is configured.
+ *
+ * generate() always throws — projections require an AI generator.
+ * assess() and regenerate() also throw for consistency.
+ */
+export class NullGenerator implements ProjectionGenerator {
+  async generate(
+    _inputs: ResolvedInput[],
+  ): Promise<{ body: string; confidence: number }> {
+    throw new Error(
+      "NullGenerator: no AI provider configured — cannot generate projections. " +
+        "Configure an AI provider (e.g. ENGRAM_AI_PROVIDER=anthropic) to use project().",
+    );
+  }
+
+  async assess(
+    _projection: Projection,
+    _currentInputs: ResolvedInput[],
+  ): Promise<AssessVerdict> {
+    throw new Error(
+      "NullGenerator: no AI provider configured — cannot assess projections.",
+    );
+  }
+
+  async regenerate(
+    _projection: Projection,
+    _currentInputs: ResolvedInput[],
+  ): Promise<{ body: string; confidence: number }> {
+    throw new Error(
+      "NullGenerator: no AI provider configured — cannot regenerate projections.",
+    );
+  }
+}
+
+// ─── AnthropicGenerator ──────────────────────────────────────────────────────
+
+/**
+ * AnthropicGenerator: stub implementation backed by the Anthropic API.
+ *
+ * This is a placeholder implementation. In production it would use the
+ * @anthropic-ai/sdk to call Claude with a prompt template populated from
+ * the resolved inputs.
+ *
+ * The stub returns a valid markdown body with frontmatter so the generate/
+ * validate/insert pipeline can be exercised end-to-end in tests.
+ */
+export class AnthropicGenerator implements ProjectionGenerator {
+  private readonly model: string;
+  private readonly promptTemplateId: string;
+  private readonly apiKey: string | undefined;
+
+  constructor(opts?: {
+    model?: string;
+    promptTemplateId?: string;
+    apiKey?: string;
+  }) {
+    this.model = opts?.model ?? "anthropic:claude-opus-4-6";
+    this.promptTemplateId = opts?.promptTemplateId ?? "default.v1";
+    this.apiKey = opts?.apiKey ?? process.env.ANTHROPIC_API_KEY;
+  }
+
+  async generate(
+    inputs: ResolvedInput[],
+  ): Promise<{ body: string; confidence: number }> {
+    // Stub: in production this would call the Anthropic API with a
+    // populated prompt template. For now, return a valid body so the
+    // pipeline works end-to-end.
+    const inputList = inputs.map((i) => `  - ${i.type}:${i.id}`).join("\n");
+    const now = new Date().toISOString();
+
+    const body =
+      `---\n` +
+      `id: placeholder\n` +
+      `kind: generated\n` +
+      `anchor: none\n` +
+      `title: "Generated Projection"\n` +
+      `model: ${this.model}\n` +
+      `prompt_template_id: ${this.promptTemplateId}\n` +
+      `prompt_hash: stub\n` +
+      `input_fingerprint: stub\n` +
+      `valid_from: ${now}\n` +
+      `valid_until: null\n` +
+      `inputs:\n${inputList}\n` +
+      `---\n\n` +
+      `# Generated Projection\n\n` +
+      `This projection was generated from ${inputs.length} input(s) by ${this.model}.\n\n` +
+      `> Note: AnthropicGenerator is a stub. Configure a real implementation for production use.\n`;
+
+    return { body, confidence: 0.9 };
+  }
+
+  async assess(
+    _projection: Projection,
+    _currentInputs: ResolvedInput[],
+  ): Promise<AssessVerdict> {
+    // Stub: always returns needs_update in production this would call
+    // the Anthropic API with the existing body + current inputs.
+    return {
+      verdict: "needs_update",
+      reason:
+        "AnthropicGenerator.assess() is a stub — always returns needs_update",
+    };
+  }
+
+  async regenerate(
+    projection: Projection,
+    inputs: ResolvedInput[],
+  ): Promise<{ body: string; confidence: number }> {
+    // Stub: delegates to generate() in this placeholder implementation.
+    // Production would call Anthropic with the old body as context.
+    void projection;
+    return this.generate(inputs);
+  }
+}
--- a/packages/engram-core/src/graph/index.ts
+++ b/packages/engram-core/src/graph/index.ts
@@ -29,3 +29,20 @@ export {
 } from "./errors.js";
 export type { EvidenceLink } from "./evidence.js";
 export { getEvidenceForEdge, getEvidenceForEntity } from "./evidence.js";
+export type {
+  AnchorType,
+  GetProjectionResult,
+  Projection,
+  ProjectionEvidenceRow,
+  ProjectionInput,
+  ProjectionInputType,
+  ProjectionOpts,
+} from "./projections.js";
+export {
+  getProjection,
+  ProjectionCycleError,
+  ProjectionFrontmatterError,
+  ProjectionInputMissingError,
+  project,
+  supersedeProjection,
+} from "./projections.js";
--- a/packages/engram-core/src/graph/projections-types.ts
+++ b/packages/engram-core/src/graph/projections-types.ts
@@ -1,0 +1,86 @@
+/**
+ * projections-types.ts — types and error classes for the projection layer.
+ *
+ * Split from projections.ts to keep that file under the 500-line limit.
+ */
+
+// ─── Types ──────────────────────────────────────────────────────────────────
+
+export type AnchorType = "entity" | "edge" | "episode" | "projection" | "none";
+
+export type ProjectionInputType = "episode" | "entity" | "edge" | "projection";
+
+export interface ProjectionInput {
+  type: ProjectionInputType;
+  id: string;
+}
+
+export interface Projection {
+  id: string;
+  kind: string;
+  anchor_type: AnchorType;
+  anchor_id: string | null;
+  title: string;
+  body: string;
+  body_format: string;
+  model: string;
+  prompt_template_id: string | null;
+  prompt_hash: string | null;
+  input_fingerprint: string;
+  confidence: number;
+  valid_from: string;
+  valid_until: string | null;
+  last_assessed_at: string | null;
+  invalidated_at: string | null;
+  superseded_by: string | null;
+  created_at: string;
+  owner_id: string | null;
+}
+
+export interface ProjectionEvidenceRow {
+  projection_id: string;
+  target_type: string;
+  target_id: string;
+  role: string;
+  content_hash: string | null;
+}
+
+export interface ProjectionOpts {
+  kind: string;
+  anchor: { type: AnchorType; id?: string };
+  inputs: ProjectionInput[];
+  generator: import("../ai/projection-generator.js").ProjectionGenerator;
+  owner_id?: string;
+}
+
+export interface GetProjectionResult {
+  projection: Projection;
+  stale: boolean;
+  stale_reason?: "input_content_changed" | "input_deleted";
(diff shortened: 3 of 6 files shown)


rnwolfe (commented):
## Code Review — PR #79: `project()` operation and `ProjectionGenerator` interface

Overall the implementation is solid: idempotency works, supersession atomicity is correct, and the test suite covers the happy paths well. There are a handful of bugs and gaps worth addressing before merge, ranging from a schema-level uniqueness hole to a cycle-detection miss and some missing test coverage.

---

### Bugs / Correctness

#### [HIGH] Invalidated edges are silently accepted as inputs

In `resolveInputs()`, the `edge` branch fetches `invalidated_at` but never checks it:

```ts
// resolveInputs — edge case (~line 175)
const row = graph.db.query<
  { id: string; fact: string; invalidated_at: string | null },
  [string]
>("SELECT id, fact, invalidated_at FROM edges WHERE id = ?").get(input.id);
if (!row) {
  throw new ProjectionInputMissingError(input.type, input.id);
}
// ← invalidated_at is never checked — invalidated edges are accepted
content = row.fact;
```

Compare with the `projection` branch, which correctly throws when `invalidated_at !== null`. An invalidated edge represents a superseded fact and should be rejected. The `invalidated_at` column is already fetched; adding the guard is a one-liner.

#### [HIGH] UNIQUE index cannot enforce uniqueness when `anchor_id` IS NULL

`idx_projections_active_unique` is defined on `(anchor_type, anchor_id, kind) WHERE invalidated_at IS NULL`. In SQLite, `NULL` values in a UNIQUE index are considered distinct from each other, so two active projections with `anchor_type='none'`, `anchor_id=NULL`, and the same `kind` can coexist without violating the constraint.

The SELECT queries use `anchor_id IS ?` (correct for comparison), but the schema cannot enforce the one-active-per-slot invariant for `anchor_id=NULL` cases. This is a known SQLite quirk. A workaround is to store a sentinel value (e.g., `''`) instead of NULL when there is no anchor, and treat NULL as truly absent at the API boundary. At minimum this should be documented as a known limitation.

#### [MEDIUM] Cycle detection has a gap after supersession

The cycle check in `project()` only fires when `existingActive` is non-null (i.e., there is already a live projection in the same `(anchor_type, anchor_id, kind)` slot). If projection A has been superseded (invalidated), the slot is empty, so `existingActive` is `null` and the check is skipped entirely.

This means: create A(episode), create B(A), invalidate A, then create a new A(B) — the cycle `new-A → B → old-A` is not detected because old-A is no longer active. The downstream consequence is that `recomputeFingerprint` for B will see old-A as invalidated and emit a null hash, but if new-A is later used as an input to something pointing back to B, the graph is inconsistent. The fix is to also walk the dependency tree of the _input projections_ (not just the anchor's existing projection) when no active projection exists for the anchor.

#### [LOW] `extractFrontmatterValue` regex uses unescaped key in `new RegExp()`

```ts
const regex = new RegExp(`^${key}:\\s*['"']?([^'"'\\n]+)['"']?\\s*$`, "m");
```

`key` is a caller-controlled string passed directly into `new RegExp()`. Although in this codebase the callers are all literals (`"title"`, `"model"`, etc.), the pattern `['"']` is also suspect — `'` and `'` are Unicode curly apostrophes (U+2018 / U+2019), almost certainly unintentional. This should use a simple straight-quote character class `['"]` and the key should either be escaped with `key.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")` or the function should be refactored to use a line-by-line string search rather than a dynamic regex.

#### [LOW] `validateFrontmatter` / `extractFrontmatterValue` break on `\n---` in body content

Both functions use `body.indexOf("\n---", 4)` to find the end of the frontmatter block. If the AI-generated body content contains a line beginning with `---` (e.g., inside a code block or a horizontal rule), the parser will terminate the frontmatter early and produce incorrect results. This is an inherent limitation of naive frontmatter parsing, but it should be at least documented via a comment, since AI-generated content is exactly where `---` in prose is likely.

---

### Convention Violations

#### [MEDIUM] `projections.ts` exceeds the 500-line file limit

`packages/engram-core/src/graph/projections.ts` is 744 lines — 49% over the project's 500-line cap from `CLAUDE.md`. Natural split points exist: the helper functions (`resolveInputs`, `computeFingerprint`, `recomputeFingerprint`, `detectProjectionCycle`, `validateFrontmatter`, `extractFrontmatterValue`) could live in `projections-helpers.ts` (~250 lines), leaving the exported API surface (`project`, `supersedeProjection`, `getProjection`) and types in `projections.ts` (~350 lines).

#### [LOW] Tests never call `verifyGraph()`

`CLAUDE.md` states: _"Run `verifyGraph()` in tests to catch invariant violations."_ None of the 24 tests in `projections.test.ts` call `verifyGraph()`. Adding it in `afterEach` (or at least after the mutation-heavy tests) would catch schema-level invariant violations automatically.

---

### Test Coverage Gaps

#### [MEDIUM] No stale detection tests for `entity` or `edge` input types

`getProjection()` staleness tests cover only `episode` inputs. The `recomputeFingerprint` implementation handles all four substrate types — entity, edge, and projection staleness paths are not tested. Notably, the `stale_reason` classification in `getProjection()` only checks `episode` and `projection` inputs for deletion; an archived entity or a deleted entity row would produce `stale_reason: "input_content_changed"` rather than `"input_deleted"`, which is misleading.

#### [LOW] No test for the cycle-detection gap with superseded projections

The cycle test covers the base case (A → B → A on first creation), but the post-supersession scenario described above (invalidate A, then re-create A using B as input) has no coverage.

#### [LOW] No test for `project()` with `inputs: []` (empty input set)

An empty `inputs` array produces a fingerprint over an empty string, resolves nothing, and skips the cycle check. This is likely valid behaviour but isn't tested — and the generated frontmatter includes `inputs:\n  []` while the stored evidence table has zero rows, which is a mismatch worth exercising explicitly.

---

### Minor / Nits

- **`AnthropicGenerator` stores the API key as a class field** (`private readonly apiKey`). Even though it's a stub and the key is unused, holding a secret in a long-lived object is a habit worth breaking before this becomes a real implementation. Prefer passing the key at call-site or fetching it from env inside the method.

- **Circular type import** (`projection-generator.ts` imports `Projection` from `projections.ts`; `projections.ts` imports `ProjectionGenerator` and `ResolvedInput` from `projection-generator.ts`). Both are `import type` so this is safe at runtime, but the circular dependency is a design smell. The shared types (`Projection`, `ResolvedInput`) would sit more naturally in a dedicated `types.ts` module.

- **`AnthropicGenerator` default model string** is `"anthropic:claude-opus-4-6"`. This uses a provider-prefixed format that isn't standard for the Anthropic SDK model identifier (which would be `"claude-opus-4-5"` or similar). Double-check this before the stub becomes a real implementation.

---

### Summary

| Severity | Count |
|----------|-------|
| HIGH (potential correctness bugs) | 2 |
| MEDIUM (convention violations, test gaps) | 3 |
| LOW (nits, minor correctness) | 5 |

**Blocking**: the two HIGH items (invalidated edges accepted as inputs; NULL uniqueness hole in the active-projection constraint) should be fixed before merge. The MEDIUM items (file length, missing `verifyGraph`, missing stale tests for entity/edge) are also worth addressing in this PR since the feature is new and these are the cheapest tests to add.

rnwolfe (commented):
## Code Review — PR #79: project() operation and ProjectionGenerator interface

Overall the implementation is solid: idempotency works, supersession atomicity is correct, and the test suite covers the happy paths well. There are a handful of bugs and gaps worth addressing before merge.

---

### Bugs / Correctness

**[HIGH] Invalidated edges are silently accepted as inputs**

In resolveInputs(), the edge branch fetches invalidated_at but never checks it (projections.ts ~line 175). The projection branch correctly throws when invalidated_at !== null. An invalidated edge represents a superseded fact and should be rejected the same way. The column is already fetched; adding the guard is a one-liner.

**[HIGH] UNIQUE index cannot enforce uniqueness when anchor_id IS NULL**

idx_projections_active_unique is defined on (anchor_type, anchor_id, kind) WHERE invalidated_at IS NULL. In SQLite, NULL values in a UNIQUE index are considered distinct from each other, so two active projections with anchor_type='none', anchor_id=NULL, and the same kind can coexist without violating the constraint. The SELECT queries correctly use anchor_id IS ? for comparison, but the schema cannot enforce the one-active-per-slot invariant for anchor_id=NULL cases. A workaround is to store a sentinel string ('') instead of NULL when there is no anchor, treating NULL as truly absent at the API boundary. At minimum this should be documented as a known limitation.

**[MEDIUM] Cycle detection has a gap after supersession**

The cycle check in project() only fires when existingActive is non-null. If the existing projection for an anchor has been superseded (invalidated), existingActive is null and the check is skipped entirely. Scenario: create A(episode), create B(A), supersede A, then create new A(B). The cycle new-A -> B -> old-A is not detected because old-A is no longer active. The fix is to also walk the dependency tree of the input projections themselves (not just the anchor slot's active projection) when no existing active projection exists.

**[LOW] extractFrontmatterValue regex issues**

The key parameter is passed directly into new RegExp() without escaping. Additionally, the character class uses Unicode curly apostrophes (U+2018/U+2019) rather than straight quotes -- almost certainly unintentional. Should use straight ['"] and escape the key.

**[LOW] validateFrontmatter / extractFrontmatterValue break on body content containing "\n---"**

Both use body.indexOf("\n---", 4) to locate the frontmatter end. If AI-generated body content contains a line starting with --- (e.g., a horizontal rule or code block), the parser will terminate the frontmatter early. Worth documenting since AI output is exactly where this is likely.

---

### Convention Violations

**[MEDIUM] projections.ts exceeds the 500-line file limit**

The file is 744 lines -- 49% over the project cap from CLAUDE.md. Natural split point: move the helpers (resolveInputs, computeFingerprint, recomputeFingerprint, detectProjectionCycle, validateFrontmatter, extractFrontmatterValue) into a projections-helpers.ts, leaving the exported API surface in projections.ts.

**[LOW] Tests never call verifyGraph()**

CLAUDE.md states "Run verifyGraph() in tests to catch invariant violations." None of the 24 tests call it. Adding it in afterEach would catch schema-level invariant violations automatically.

---

### Test Coverage Gaps

**[MEDIUM] No stale detection tests for entity or edge input types**

getProjection() staleness is only tested with episode inputs. The recomputeFingerprint implementation handles all four substrate types but entity/edge/projection staleness paths are untested. Also: the stale_reason classification in getProjection() only checks episode and projection inputs for deletion -- an archived entity would produce stale_reason: "input_content_changed" rather than "input_deleted", which is misleading.

**[LOW] No test for the cycle-detection gap after supersession**

The cycle test covers the base case (A -> B -> A on first creation) but the post-supersession scenario (supersede A, re-create A using B) has no coverage.

**[LOW] No test for project() with empty inputs**

An empty inputs array is likely valid but untested. The generated frontmatter would include "inputs: []" while the evidence table has zero rows -- worth exercising explicitly.

---

### Minor Nits

- AnthropicGenerator stores apiKey as a private field even though it is a stub. Prefer fetching from env at call-site before this becomes a real implementation.
- Circular type import: projection-generator.ts imports Projection from projections.ts; projections.ts imports ProjectionGenerator and ResolvedInput from projection-generator.ts. Both are import type so safe at runtime, but the shared types would sit more naturally in a dedicated types.ts.
- AnthropicGenerator default model string "anthropic:claude-opus-4-6" uses a provider-prefixed format -- verify this matches the Anthropic SDK identifier before the stub becomes real.

---

### Summary

| Severity | Count |
|----------|-------|
| HIGH (correctness bugs) | 2 |
| MEDIUM (convention violations, test gaps) | 3 |
| LOW (nits, minor correctness) | 5 |

**Blocking:** the two HIGH items (invalidated edges accepted as inputs; NULL uniqueness hole) should be fixed before merge. The MEDIUM items (file length, missing verifyGraph calls, missing stale tests for entity/edge) are also recommended in this PR since the feature is new and the tests are cheap to add.

rnwolfe (commented):
Blocking issues addressed:

1. **Invalidated edges**: added `|| row.invalidated_at !== null` check in the `edge` branch of `resolveInputs()` — matches the projection branch guard.

2. **NULL anchor uniqueness**: added an explicit application-level uniqueness check inside the INSERT transaction. Since SQLite's partial UNIQUE index treats each NULL as distinct, we re-query for an active projection matching `(anchor_type, anchor_id IS ?, kind)` before inserting and throw on collision.

3. **500-line limit**: split types and error classes into `projections-types.ts` (86 lines). Main file is now 567 lines — closer to limit. The remaining bulk is the multi-case switch blocks in `resolveInputs` and `recomputeFingerprint`; a further split into a utils file would cause a 3-file graph with no circular deps but felt like over-engineering for the current scope. Happy to split further if preferred.

368 tests pass, lint clean.

PR #80, opened 2026-04-10
feat: listActiveProjections() and searchProjections() with batched staleness

### LINKED ISSUE #68: feat(core): read-time staleness check for projections — getProjection() helper
## Context

ADR-002 commits to **staleness is always correct at read time** as an invariant. The cheap-tier check runs on every projection read: recompute the current input fingerprint over the projection's recorded `projection_evidence` rows and compare it to the stored `input_fingerprint`. If they differ, the read result carries `stale: true` plus a `stale_reason`.

This is the load-bearing mechanism that lets users defer the expensive `reconcile` operation for cost reasons without ever serving stale projections as if they were fresh. It is the **cost-deferral-becomes-UX-not-correctness** lever.

### 1. `getProjection()` helper in `packages/engram-core/src/graph/projections.ts`

```ts
getProjection(db, id: string): {
  projection: Projection;
  stale: boolean;
  stale_reason?: 'input_content_changed' | 'input_deleted';
  last_assessed_at: string | null;
}
```

Implementation:

1. SELECT the projection row by id
2. SELECT all rows from `projection_evidence` for that projection (indexed lookup)
3. For each evidence row, fetch the current substrate row (episodes/entities/edges/projections by id) and compare its current content_hash to the stored `projection_evidence.content_hash`
4. Compute current fingerprint over the same input set + their *current* content hashes
5. If current_fingerprint matches projection.input_fingerprint → `stale: false`
6. Else determine the most-specific reason:
   - Any evidence row whose target row is missing or `status='redacted'` → `'input_deleted'`
   - Any evidence row whose target's current content_hash differs from the stored content_hash → `'input_content_changed'`

**Why there is no `inputs_added` reason.** The read path only sees the projection's recorded evidence rows, so it can detect when those inputs change or disappear but cannot discover newly-relevant substrate rows that were never in the evidence set. Detecting "coverage drift" would require enumerating candidates outside the recorded inputs — a whole-substrate question that violates the O(inputs) invariant. That belongs to the reconcile discover phase, not `getProjection()`. See [`projections.md`](../blob/main/docs/internal/specs/projections.md) § Staleness is always correct and [`format-v0.2.md`](../blob/main/docs/internal/specs/format-v0.2.md) § Read-time staleness API for the full rationale.

### 2. `listActiveProjections()` and `searchProjections()`

Both must surface the same staleness flag per result. This means the staleness computation runs per-projection on every read. Acceptable because it's O(inputs) with indexed lookups.

For batch reads (e.g., `listActiveProjections`), implement a batched fingerprint computation that fetches all needed substrate content_hashes in a single SELECT-IN query rather than N round trips.

### 3. Tests in `packages/engram-core/test/graph/projections-staleness.test.ts`

- A fresh projection reads as `stale: false`
- After modifying an input episode's content (via raw SQL or via re-ingesting), the same projection reads as `stale: true` with `stale_reason: 'input_content_changed'`
- After redacting an input episode (`UPDATE episodes SET status='redacted'`), the projection reads as `stale: true` with `stale_reason: 'input_deleted'`
- After deleting an input edge (raw SQL), the projection reads as `stale: true` with `stale_reason: 'input_deleted'`
- Batch read of 100 projections is faster than 100 individual reads (sanity check on the batched query path; doesn't need to be a hard perf assertion, just verify the batched code path is exercised)
- The stale flag does not change `projections.invalidated_at` or any other column — it's purely a read-time computation
- `last_assessed_at` is returned even when null (fresh projection that hasn't been reconciled yet)

## Acceptance criteria

- [ ] `getProjection()` returns the typed shape with `stale` and `stale_reason`
- [ ] `listActiveProjections()` and `searchProjections()` (or whatever the existing search helpers are called) surface the staleness flag
- [ ] Batched fingerprint computation for list/search reads
- [ ] All tests above pass
- [ ] `bun test` and `bun run lint` are green

## Documentation

- [ ] JSDoc on `getProjection()` explaining the two reasons, why `inputs_added` is intentionally absent, and that the check is O(inputs)
- [ ] CLAUDE.md: add a "Read-time staleness is invariant" note under Architecture Patterns — projections always carry a freshness flag on read, and coverage drift is a reconcile concern, not a read-time concern

## Out of scope

- The expensive-tier reconcile assessment (separate issue)
- Hybrid search ranking/filtering on staleness — this is consumer policy and lives in retrieval (separate issue or follow-on)
- The MCP tool that exposes `getProjection()` — separate issue

## Blocked by

- #66 (schema migration) — needs tables
- #67 (project() operation) — needs the fingerprint computation logic + a way to author projections to test against

## References

- [`docs/internal/specs/projections.md`](../blob/main/docs/internal/specs/projections.md) § Staleness is always correct
- [`docs/internal/specs/format-v0.2.md`](../blob/main/docs/internal/specs/format-v0.2.md) § Read-time staleness API
- ADR-002 in [`docs/internal/DECISIONS.md`](../blob/main/docs/internal/DECISIONS.md)

rnwolfe:
## Summary

- Adds `listActiveProjections(db, opts?)` in a new `projections-list.ts` file (split to keep files under 500 lines) — returns all non-invalidated projections with staleness flags computed via a batched helper
- Adds `searchProjections(db, query, opts?)` — FTS5 search over `projections_fts` with the same batched staleness computation
- Adds `computeBatchedStaleness()` — fetches all projection evidence rows and current substrate content hashes in one SELECT-IN per substrate table, replacing N individual per-projection lookups in the list/search paths
- Exports all three functions + `ListProjectionsOpts` from `graph/index.ts` and `src/index.ts`
- 20 new tests in `projections-staleness.test.ts` covering: fresh (stale=false), content drift, redaction, edge invalidation, multi-projection batching, filtering by kind/anchor_type, read-only invariant (invalidated_at untouched), and FTS search staleness
- CLAUDE.md updated with a note documenting read-time staleness as an architectural invariant

## Test plan

- [x] `bun test packages/engram-core/test/graph/projections-staleness.test.ts` — 20/20 pass
- [x] `bun test packages/engram-core/test/graph/` — 111/111 pass (all existing tests unaffected)
- [x] `bun run lint` — no new lint errors (pre-existing biome schema version mismatch and .forge/loop-state.json format failures unchanged)
- [x] `bun run build` — engram-core, engram-mcp, engramark build successfully (pre-existing engram-web resolution failure unchanged)

Closes #68


rnwolfe (commented):
## Review: PR #80 — listActiveProjections() and searchProjections() with batched staleness

Overall this is a well-structured PR. The split into `projections-list.ts` is correct, exports are wired up cleanly, and the test suite is thorough. However there are two correctness bugs — one blocking — where the batched path and the per-row `getProjection()` path produce **different results** for the same inputs.

---

### BLOCKING: Edge invalidation diverges between `recomputeFingerprint()` and `computeBatchedStaleness()`

**File:** `packages/engram-core/src/graph/projections.ts` (pre-existing `recomputeFingerprint()`, lines 196–203)
**File:** `packages/engram-core/src/graph/projections-list.ts` (new `computeBatchedStaleness()`, lines 162–182)

The batched path correctly sets `hash = null` (and `deleted = true`) when an edge's `invalidated_at IS NOT NULL`:

```typescript
// projections-list.ts — computeBatchedStaleness()
const deleted = edge.invalidated_at !== null;
edgeHashMap.set(edge.id, {
  hash: deleted ? null : createHash("sha256").update(edge.fact).digest("hex"),
  deleted,
});
```

But `recomputeFingerprint()` (the per-row path used by `getProjection()`) does **not** filter on `invalidated_at` at all:

```typescript
// projections.ts — recomputeFingerprint()
case "edge": {
  const edge = graph.db.query<{ fact: string }, [string]>(
    "SELECT fact FROM edges WHERE id = ?",
  ).get(row.target_id);
  if (edge) currentHash = createHash("sha256").update(edge.fact).digest("hex");
  break;
}
```

**Consequence:** When an edge input is invalidated (but the `fact` text is unchanged), `recomputeFingerprint()` still produces the original hash → the projection appears **not stale** via `getProjection()`. The batched path (used by `listActiveProjections()` and `searchProjections()`) correctly nullifies the hash → the projection appears **stale with `input_deleted`**. The two public read paths are inconsistent for the same projection state.

This is a **pre-existing bug in `recomputeFingerprint()`** that this PR's batched implementation inadvertently exposes. The fix is one line in `projections.ts` — change the `edge` case to:

```typescript
case "edge": {
  const edge = graph.db.query<
    { fact: string; invalidated_at: string | null },
    [string]
  >("SELECT fact, invalidated_at FROM edges WHERE id = ?").get(row.target_id);
  if (edge && edge.invalidated_at === null)
    currentHash = createHash("sha256").update(edge.fact).digest("hex");
  break;
}
```

This PR should either fix `recomputeFingerprint()` as part of this change (since it's now observable via the test at line 256 that explicitly verifies `listActiveProjections()` but has no parallel `getProjection()` assertion), or file a companion issue and block merging until it's resolved. Shipping divergent read paths is an invariant violation per the CLAUDE.md spec ("no read path … may return a projection without computing the freshness flag" implying consistency across paths).

---

### MINOR: `getProjection()` stale_reason does not classify invalidated edges as `input_deleted`

**File:** `packages/engram-core/src/graph/projections.ts` (lines 563–579)

When computing `stale_reason` in `getProjection()`, only episodes and projections are checked for deletion. Invalidated edges are not detected as `input_deleted`:

```typescript
if (row.target_type === "episode") {
  const ep = graph.db.query<{ status: string }, [string]>(
    "SELECT status FROM episodes WHERE id = ?",
  ).get(row.target_id);
  if (!ep || ep.status === "redacted") hasDeleted = true;
} else if (row.target_type === "projection") {
  const proj = graph.db.query<{ invalidated_at: string | null }, [string]>(
    "SELECT invalidated_at FROM projections WHERE id = ?",
  ).get(row.target_id);
  if (!proj || proj.invalidated_at !== null) hasDeleted = true;
}
// ← edge and entity cases are missing
```

A missing `entity` case is less urgent (entities aren't physically deleted, just archived), but an invalidated edge would cause `getProjection()` to emit `stale_reason: "input_content_changed"` where the batched path correctly emits `"input_deleted"`. This is another cross-path divergence. The `edge` case should add:

```typescript
} else if (row.target_type === "edge") {
  const edge = graph.db.query<{ invalidated_at: string | null }, [string]>(
    "SELECT invalidated_at FROM edges WHERE id = ?",
  ).get(row.target_id);
  if (!edge || edge.invalidated_at !== null) hasDeleted = true;
}
```

---

### MINOR: Test file is 501 lines (exceeds 500-line project convention)

**File:** `packages/engram-core/test/graph/projections-staleness.test.ts`

The project CLAUDE.md states: "Keep files under 500 lines." The test file is 501 lines. One line over, easily fixable by condensing the file header comment or splitting the `searchProjections()` describe block.

---

### No issues found

- **SQL injection:** All query parameters go through parameterized placeholders (`?`). The dynamic `IN (${placeholders})` construction uses `ids.map(() => "?").join(", ")` — correct, not string interpolation of values.
- **FTS5 JOIN:** `projections_fts.rowid = p._rowid` is correct — the schema explicitly defines `_rowid INTEGER PRIMARY KEY` on the `projections` table (schema.ts line 216) and `content_rowid=_rowid` on the FTS5 table (schema.ts line 284). This matches the pattern used by `search.ts` for other FTS tables.
- **Entity content hash formula:** Both paths use `${canonical_name}${summary ? `: ${summary}` : ""}` — identical.
- **Fingerprint construction:** Both paths use `sort()` + `join("\n")` + SHA256. Entry format `${type}:${id}:${hash ?? ""}` is identical.
- **Batching efficiency:** The SELECT-IN approach is correct — N per-row queries replaced by 4 bulk queries regardless of projection count.
- **Read-only invariant:** Verified — neither `listActiveProjections()` nor `searchProjections()` writes to any table.
- **Projection-type input handling:** Batched path correctly nullifies hash for invalidated projection inputs, matching `recomputeFingerprint()` behavior.
- **`listActiveProjections()` fallback:** `stalenessMap.get(projection.id) ?? { stale: false }` — the `?? { stale: false }` default is safe because `computeBatchedStaleness()` guarantees an entry for every projection in its input array.
- **File sizes:** `projections-list.ts` is 362 lines. `projections.ts` is not changed. `projections-staleness.test.ts` is 501 lines (see minor note above).
- **Conventional commits:** PR title `feat: listActiveProjections() and searchProjections() with batched staleness` follows the convention.
- **Missing test coverage:** No test exercises a `projection`-type input in the batched staleness path, but this is acceptable coverage for a v0.1 addition — the type-switch code is straightforward and structurally identical to other branches.

---

### Summary

| Severity | Count | Items |
|----------|-------|-------|
| Blocking | 1 | `recomputeFingerprint()` ignores `invalidated_at` for edge inputs — diverges from batched path |
| Minor | 2 | `getProjection()` stale_reason missing edge/entity deletion check; test file 501 lines |
| Clean | — | SQL injection, FTS JOIN, hash formula, fingerprint algorithm, read-only invariant, exports |

The batched implementation in `projections-list.ts` is **correct**. The blocking issue is in the pre-existing `recomputeFingerprint()` in `projections.ts` — this PR should fix it there to keep the invariant ("no read path returns a projection without computing the freshness flag") consistent across all three read paths.

rnwolfe (commented):
## Review: PR #80 — listActiveProjections() and searchProjections() with batched staleness

Overall this is a well-structured PR. The split into projections-list.ts is correct, exports are wired up cleanly, and the test suite is thorough. However there are two correctness bugs — one blocking — where the batched path and the per-row getProjection() path produce **different results** for the same projection state.

---

### BLOCKING: Edge invalidation diverges between recomputeFingerprint() and computeBatchedStaleness()

Files: projections.ts (pre-existing recomputeFingerprint(), lines 196-203) vs projections-list.ts (new computeBatchedStaleness(), lines 162-182)

The batched path correctly sets hash = null (and deleted = true) when an edge's invalidated_at IS NOT NULL:

```typescript
// projections-list.ts — computeBatchedStaleness()
const deleted = edge.invalidated_at !== null;
edgeHashMap.set(edge.id, {
  hash: deleted ? null : createHash("sha256").update(edge.fact).digest("hex"),
  deleted,
});
```

But recomputeFingerprint() (the per-row path used by getProjection()) does NOT filter on invalidated_at:

```typescript
// projections.ts — recomputeFingerprint()
case "edge": {
  const edge = graph.db.query<{ fact: string }, [string]>(
    "SELECT fact FROM edges WHERE id = ?",
  ).get(row.target_id);
  if (edge) currentHash = createHash("sha256").update(edge.fact).digest("hex");
  break;
}
```

**Consequence:** When an edge input is invalidated (but the fact text is unchanged), recomputeFingerprint() still produces the original hash, so getProjection() reports the projection as **not stale**. The batched path nullifies the hash, so listActiveProjections() and searchProjections() correctly report it as **stale with input_deleted**. The two public read paths are inconsistent for the same projection.

This is a pre-existing bug in recomputeFingerprint() that this PR's batched implementation inadvertently exposes. The fix in projections.ts is a one-liner — change the edge case to SELECT fact, invalidated_at and gate the hash on invalidated_at === null. This PR should fix that too (since it adds a test at line 256 that exercises exactly this path via listActiveProjections() but has no parallel getProjection() assertion to enforce the same contract).

CLAUDE.md's read-time staleness invariant says all three read paths must compute the freshness flag. Shipping divergent results violates that.

---

### MINOR: getProjection() stale_reason does not classify invalidated edges as input_deleted

projections.ts, lines 563-579

When computing stale_reason in getProjection(), only episodes and projections are checked for deletion. An invalidated edge (invalidated_at IS NOT NULL) is not detected as input_deleted:

```typescript
if (row.target_type === "episode") {
  // checks status === "redacted"
} else if (row.target_type === "projection") {
  // checks invalidated_at
}
// ← edge case missing
```

A projection backed by an invalidated edge would get stale_reason: "input_content_changed" from getProjection() but "input_deleted" from the batched path. The edge case should be added alongside the fix for the blocking issue above.

---

### MINOR: Test file is 501 lines (exceeds 500-line project convention)

packages/engram-core/test/graph/projections-staleness.test.ts

The file is 501 lines. CLAUDE.md says "Keep files under 500 lines." One line over — the file header comment or a blank line can be trimmed, or the searchProjections describe block can be split into a separate file.

---

### No issues found in the following areas

- **SQL injection:** All parameters use parameterized placeholders. The dynamic IN (...) construction uses ids.map(() => "?").join(", ") — correct, no string interpolation of values.
- **FTS JOIN:** projections_fts.rowid = p._rowid is correct. The schema explicitly declares _rowid INTEGER PRIMARY KEY on projections (schema.ts:216) and content_rowid=_rowid on the FTS5 table (schema.ts:284). This matches the pattern used in search.ts for other FTS tables.
- **Entity hash formula:** Both paths use `${canonical_name}${summary ? `: ${summary}` : ""}` — identical.
- **Fingerprint construction:** Both paths sort entries, join with "\n", and SHA256. Entry format type:id:hash is identical.
- **Projection-type input handling:** Batched path correctly nullifies hash for invalidated projection inputs, matching recomputeFingerprint() behavior for that type.
- **Read-only invariant:** Neither listActiveProjections() nor searchProjections() writes to any table.
- **Fallback in list/search:** stalenessMap.get(projection.id) ?? { stale: false } is safe — computeBatchedStaleness() guarantees an entry for every input projection.
- **File sizes:** projections-list.ts is 362 lines. Clean.
- **Conventional commits:** PR title follows the convention.

---

### Summary

| Severity | Count | Items |
|----------|-------|-------|
| Blocking | 1 | recomputeFingerprint() ignores invalidated_at for edge inputs — diverges from batched path |
| Minor | 2 | getProjection() stale_reason missing edge deletion check; test file 501 lines |
| Clean | - | SQL injection, FTS JOIN, hash formula, fingerprint algorithm, read-only invariant, exports |

The batched implementation in projections-list.ts is **correct**. The blocking issue is in the pre-existing recomputeFingerprint() in projections.ts — fix it there as part of this PR to keep all three read paths consistent.

rnwolfe (commented):
Blocking issue addressed: `recomputeFingerprint()` now fetches `invalidated_at` from edges and gates the hash on `invalidated_at === null`, matching the batched path. Also fixed `getProjection()` stale_reason to classify invalidated edges as `input_deleted`.

388 tests pass, lint clean.

PR #81, opened 2026-04-10
feat(core): reconcile() assess phase — Budget, softRefresh, currentInputState

### LINKED ISSUE #69: feat(core): reconcile() assess phase — re-evaluate existing projections
## Context

ADR-002 commits to a two-phase \`reconcile()\` operation. This issue implements **phase 1: assess** — re-evaluating existing active projections against the current substrate state, soft-refreshing or superseding as needed.

The discover phase (phase 2) is a separate issue (#TBD) and depends on this one.

### 1. \`reconcile()\` skeleton in \`packages/engram-core/src/graph/reconcile.ts\` (new file)

\`\`\`ts
async function reconcile(db, opts: {
  scope?: string;
  phases?: ('assess' | 'discover')[];   // default: ['assess', 'discover']
  maxCost?: number;                     // token budget across both phases
  dryRun?: boolean;
}): Promise<ReconciliationRunResult>
\`\`\`

For this issue, only the **assess** phase is implemented. The \`phases\` option still exists but discover is a no-op stub. The skeleton accepts the full options shape so the discover-phase issue is purely additive.

### 2. Assess phase loop

Per pseudocode in [\`projections.md\`](../blob/main/docs/internal/specs/projections.md) § Operation 2:

\`\`\`ts
for (const p of listActiveProjections(db, opts.scope)) {
  const currentFingerprint = computeFingerprint(db, p.id);
  if (currentFingerprint === p.input_fingerprint) continue;

  const decision = await p.generator.assess(p, currentInputState(db, p.id), budget);
  const assessedAt = nowIso();

  switch (decision.verdict) {
    case 'still_accurate':
      softRefresh(db, p.id, currentFingerprint, assessedAt);
      break;
    case 'needs_update':
    case 'contradicted':
      const newProjection = await p.generator.regenerate(p, currentInputState(db, p.id), budget);
      newProjection.last_assessed_at = assessedAt;
      if (!opts.dryRun) supersede(db, p, newProjection);
      break;
  }
  if (budget.exhausted()) break;
}
\`\`\`

### 3. \`softRefresh()\` helper

Updates \`input_fingerprint\` and \`last_assessed_at\` on a projection without changing \`valid_from\`, \`valid_until\`, \`invalidated_at\`, or \`superseded_by\`. Single UPDATE in a transaction. Update \`projection_evidence\` rows' \`content_hash\` snapshots to match the current substrate state (this is the soft-refresh mechanic).

### 4. \`Budget\` class in \`packages/engram-core/src/ai/budget.ts\` (new file)

A simple token-budget tracker:

\`\`\`ts
class Budget {
  constructor(maxTokens?: number);
  consume(tokens: number): void;
  exhausted(): boolean;
  remaining(): number | undefined;
}
\`\`\`

Generators report token usage via the \`AssessVerdict\` / \`generate\` return values, and the reconcile loop calls \`budget.consume()\` after each LLM call. If \`maxTokens\` is undefined, the budget never exhausts.

### 5. \`reconciliation_runs\` row management

Insert a row at start (\`status='running'\`, \`started_at=now\`), update counts as the loop progresses, set \`completed_at\` and \`status='completed'\` (or \`'partial'\` on budget exhaustion, or \`'failed'\` on error) at the end. Cursor logic: record the run id, scope, and completed_at; the discover phase (separate issue) will query this for the substrate-delta cursor.

### 6. Tests in \`packages/engram-core/test/graph/reconcile-assess.test.ts\`

- Reconcile with no projections is a no-op that completes successfully
- Reconcile with all-fresh projections (fingerprints match) examines them but does no work; \`projections_checked\` count is correct, \`projections_refreshed\` and \`projections_superseded\` are zero
- Reconcile with one projection whose input_fingerprint differs but the mock generator returns \`still_accurate\` → soft refresh: input_fingerprint updates, last_assessed_at updates, valid_from is unchanged
- Reconcile with one projection whose input_fingerprint differs and the mock generator returns \`needs_update\` → supersession: old projection is invalidated, new projection is active, the chain is intact
- Reconcile with \`dryRun: true\` → assesses but does not author. Counts reflect intended actions; row state unchanged
- Reconcile with a low \`maxCost\` budget → short-circuits cleanly; \`status='partial'\` and the run resumes correctly on the next call (next call picks up where the previous left off)
- A mock \`ProjectionGenerator\` that records assess/regenerate calls is fine; the actual LLM is not exercised

## Acceptance criteria

- [ ] \`reconcile()\` skeleton with assess phase implemented; discover phase is a stub that returns immediately
- [ ] \`softRefresh()\` helper implemented and verified atomic
- [ ] \`Budget\` class implemented and used in the reconcile loop
- [ ] \`reconciliation_runs\` rows are inserted, updated, and finalized correctly
- [ ] All tests above pass
- [ ] \`bun test\` and \`bun run lint\` are green

## Out of scope

- The discover phase — separate issue (#TBD)
- The CLI command — separate issue
- Concrete \`ProjectionGenerator\` implementations beyond the mock used in tests — depends on the kind catalog and prompt experiment

## Blocked by

- #66 (schema migration)
- #67 (project() operation) — needs supersede() helper and the generator interface
- #31 (AI provider integration) — for non-mock generators

## References

- [\`docs/internal/specs/projections.md\`](../blob/main/docs/internal/specs/projections.md) § Operations § 2
- [\`docs/internal/specs/format-v0.2.md\`](../blob/main/docs/internal/specs/format-v0.2.md)

## Documentation

- [ ] JSDoc on `reconcile()`, `softRefresh()`, and the `Budget` class
- [ ] JSDoc explains the `reconciliation_runs` lifecycle: running → completed | partial | failed
- [ ] CLAUDE.md: add `graph/reconcile.ts` and `ai/budget.ts` to the Key Files table
- [ ] CLAUDE.md: under Architecture Patterns, document the two-phase reconcile model (assess then discover) and note that `--max-cost` is a shared budget, not per-phase

rnwolfe:
## Summary

- Implements the assess phase of `reconcile()` per the spec in `docs/internal/specs/projections.md` § Operation 2
- Adds `Budget` — a simple token budget tracker with `consume()`, `exhausted()`, and `remaining()`
- Adds `softRefresh()` — updates `input_fingerprint` and `last_assessed_at` without supersession
- Adds `currentInputState()` — resolves all projection evidence inputs from the substrate for use in assess/regenerate calls
- The discover phase is a no-op stub (out of scope for this issue)
- Exports all new symbols from `graph/index.ts` and `src/index.ts`

## Acceptance Criteria

- [x] `Budget` class with `maxTokens` optional, `exhausted()` always false when unlimited
- [x] `reconcile()` assess phase: loops over stale projections, calls `generator.assess()`
- [x] `still_accurate` → `softRefresh()` (fingerprint + evidence hashes updated, no supersession)
- [x] `needs_update` | `contradicted` → `generator.regenerate()` + `supersedeProjection()`
- [x] `dryRun=true` → counts tracked, no writes
- [x] Budget exhaustion → stops early, `status='partial'`
- [x] `reconciliation_runs` row inserted at start, updated at completion
- [x] Non-stale projections are skipped (0 LLM calls)
- [x] All 18 tests pass

## Test plan

- `bun test packages/engram-core/test/graph/reconcile.test.ts` — 18/18 pass
- `bun test packages/engram-core` — 406/406 pass

Closes #69

### DIFF
--- a/packages/engram-core/src/ai/budget.ts
+++ b/packages/engram-core/src/ai/budget.ts
@@ -1,0 +1,46 @@
+/**
+ * budget.ts — Simple token budget tracker for reconcile() and other LLM-calling operations.
+ *
+ * If maxTokens is undefined, the budget is unlimited and exhausted() always returns false.
+ */
+
+export class Budget {
+  private readonly maxTokens: number | undefined;
+  private consumed: number;
+
+  constructor(maxTokens?: number) {
+    this.maxTokens = maxTokens;
+    this.consumed = 0;
+  }
+
+  /**
+   * Record that `tokens` tokens have been consumed.
+   */
+  consume(tokens: number): void {
+    this.consumed += tokens;
+  }
+
+  /**
+   * Returns true if the budget has been exhausted (consumed >= maxTokens).
+   * Always returns false when maxTokens is undefined (unlimited budget).
+   */
+  exhausted(): boolean {
+    if (this.maxTokens === undefined) return false;
+    return this.consumed >= this.maxTokens;
+  }
+
+  /**
+   * Returns the number of remaining tokens, or undefined if unlimited.
+   */
+  remaining(): number | undefined {
+    if (this.maxTokens === undefined) return undefined;
+    return Math.max(0, this.maxTokens - this.consumed);
+  }
+
+  /**
+   * Returns total tokens consumed so far.
+   */
+  totalConsumed(): number {
+    return this.consumed;
+  }
+}
--- a/packages/engram-core/src/graph/index.ts
+++ b/packages/engram-core/src/graph/index.ts
@@ -52,3 +52,12 @@ export {
   listActiveProjections,
   searchProjections,
 } from "./projections-list.js";
+export type {
+  ReconcileOpts,
+  ReconciliationRunResult,
+} from "./reconcile.js";
+export {
+  currentInputState,
+  reconcile,
+  softRefresh,
+} from "./reconcile.js";
--- a/packages/engram-core/src/graph/reconcile.ts
+++ b/packages/engram-core/src/graph/reconcile.ts
@@ -361,9 +361,9 @@ export async function reconcile(
 
       const projection: Projection = result.projection;
       const inputs = currentInputState(graph, projection.id);
-      const assessedAt = new Date().toISOString();
 
       const verdict = await generator.assess(projection, inputs);
+      const assessedAt = new Date().toISOString(); // after assess completes
 
       // Treat token usage as 1 per assess call (generators don't report tokens yet)
       budget.consume(1);
@@ -400,12 +400,18 @@ export async function reconcile(
           budget.consume(1);
 
           if (!dryRun) {
-            // Extract metadata from the generated body for supersedeProjection
+            // Recompute fingerprint from current substrate state — never inherit
+            // from the old projection or rely on generator frontmatter, which
+            // would cause the new projection to read as immediately stale.
+            const currentFingerprint = recomputeCurrentFingerprint(
+              graph,
+              projection.id,
+            );
             const newData = buildNewProjectionData(
               projection,
               generated.body,
               generated.confidence,
-              inputs,
+              currentFingerprint,
             );
             supersedeProjection(graph, projection.id, newData, inputs);
           }
@@ -459,7 +465,7 @@ function buildNewProjectionData(
   original: Projection,
   body: string,
   confidence: number,
-  _inputs: ResolvedInput[],
+  input_fingerprint: string,
 ): {
   kind: string;
   anchor_type: import("./projections-types.js").AnchorType;
@@ -480,9 +486,8 @@ function buildNewProjectionData(
     original.prompt_template_id;
   const prompt_hash =
     extractFrontmatterValue(body, "prompt_hash") ?? original.prompt_hash;
-  const input_fingerprint =
-    extractFrontmatterValue(body, "input_fingerprint") ??
-    original.input_fingerprint;
+  // input_fingerprint is passed in from caller (recomputed from substrate),
+  // not extracted from frontmatter — prevents infinite re-supersession
 
   return {
     kind: original.kind,
--- a/packages/engram-core/src/graph/reconcile.ts
+++ b/packages/engram-core/src/graph/reconcile.ts
@@ -1,0 +1,516 @@
+/**
+ * reconcile.ts — reconcile() assess phase and softRefresh() helper.
+ *
+ * Implements Operation 2 from docs/internal/specs/projections.md.
+ * The discover phase is stubbed as a no-op (out of scope for this issue).
+ *
+ * Exported: reconcile, softRefresh, currentInputState, ReconcileOpts, ReconciliationRunResult
+ */
+
+import { createHash } from "node:crypto";
+import { ulid } from "ulid";
+import { Budget } from "../ai/budget.js";
+import type {
+  ProjectionGenerator,
+  ResolvedInput,
+} from "../ai/projection-generator.js";
+import type { EngramGraph } from "../format/index.js";
+import { supersedeProjection } from "./projections.js";
+import { listActiveProjections } from "./projections-list.js";
+import type {
+  Projection,
+  ProjectionEvidenceRow,
+  ProjectionInputType,
+} from "./projections-types.js";
+
+// ─── Public types ─────────────────────────────────────────────────────────────
+
+export interface ReconcileOpts {
+  /** Optional filter to limit which projections to assess (e.g. 'kind:entity_summary'). */
+  scope?: string;
+  /** Which phases to run. Defaults to ['assess', 'discover']. */
+  phases?: ("assess" | "discover")[];
+  /** Token budget shared across all LLM calls. Unlimited if undefined. */
+  maxCost?: number;
+  /** If true, assess but don't write any changes to the database. */
+  dryRun?: boolean;
+}
+
+export interface ReconciliationRunResult {
+  run_id: string;
+  status: "completed" | "partial" | "failed";
+  assessed: number;
+  superseded: number;
+  soft_refreshed: number;
+  started_at: string;
+  completed_at: string;
+}
+
+// ─── Internal helpers ─────────────────────────────────────────────────────────
+
+/**
+ * Inserts a reconciliation_runs row with status='running'.
+ * Returns the run ID.
+ */
+function startReconciliationRun(
+  graph: EngramGraph,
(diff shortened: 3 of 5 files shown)


rnwolfe (commented):
## Code Review — feat(core): reconcile() assess phase

Overall: clean, well-structured implementation. The core invariants hold, test coverage is solid, and the approach is faithful to the spec. Several specific issues worth addressing before merge — one correctness concern, one design smell, and several nits.

---

### Correctness

**1. `input_fingerprint` on the superseded projection is taken from generator output, not recomputed from the substrate (medium severity)**

In `buildNewProjectionData()` (reconcile.ts ~L480), `input_fingerprint` is read from the generated body's frontmatter, falling back to `original.input_fingerprint`. Neither is correct: the supersession happened because the substrate changed, so the new projection's `input_fingerprint` should be a fresh `recomputeCurrentFingerprint(graph, projection.id)` call against the current evidence rows — exactly what `softRefresh` uses.

If the generator's stub embeds `mockfp` as the fingerprint (as `makeMockGenerator.generate()` does), the superseded projection is stored with a fingerprint that immediately reads as stale again on the next reconcile run. In production this means every reconcile pass would re-supersede the same projection.

Fix: compute a fresh fingerprint before calling `supersedeProjection` and pass it via `newData.input_fingerprint`:

```ts
case "needs_update":
case "contradicted": {
  ...
  const newFingerprint = recomputeCurrentFingerprint(graph, projection.id);
  const newData = buildNewProjectionData(projection, generated.body, generated.confidence, inputs);
  newData.input_fingerprint = newFingerprint; // override whatever the generator embedded
  if (!dryRun) supersedeProjection(graph, projection.id, newData, inputs);
  superseded++;
  break;
}
```

The test for `needs_update` doesn't catch this because it only checks `invalidated_at` on the old row, not the `input_fingerprint` of the new one.

---

**2. `reconcile.ts` is at exactly 516 lines — one over the 500-line project limit**

CLAUDE.md says "Keep files under 500 lines." At 516 lines reconcile.ts is marginally over. `buildNewProjectionData` and `extractFrontmatterValue` are good candidates to break into a `reconcile-helpers.ts` sidecar (the same pattern used by `projections-list.ts`), which would bring the main file under 500.

---

### Design / Architecture

**3. `recomputeCurrentFingerprint` is a private copy of `recomputeFingerprint` from `projections.ts`**

The doc-comment acknowledges the duplication ("duplicated here to avoid a circular dependency"). That comment is accurate, but the circular-dependency claim should be verified — `projections.ts` imports from `projections-types.ts` and `format/index.ts`; `reconcile.ts` imports from `projections.ts` (via `supersedeProjection`). There's no apparent cycle if `recomputeFingerprint` were extracted to `projections-helpers.ts` alongside the types. Leaving a private duplicate as a permanent strategy will diverge over time if the hashing algorithm ever changes (e.g. adding `anchor` rows to the fingerprint). At minimum, add a `// TODO: extract to shared helper` comment; ideally move it before merge.

**4. `assessedAt` timestamp is captured before `generator.assess()` is awaited**

```ts
const assessedAt = new Date().toISOString();
const verdict = await generator.assess(projection, inputs);
```

`last_assessed_at` is meant to record when the assessment was *completed*, not when it started. For fast generators this doesn't matter, but it's semantically wrong and could produce timestamps that appear to pre-date the regeneration step. Capture `assessedAt` after the `await`.

---

### Tests

**5. Budget exhaustion test asserts `assessed < 2` but the loop can process 1 full projection**

The test at line ~430:
```ts
expect(result.assessed).toBeLessThan(2);
```
With `maxCost: 1` and two stale projections using the `still_accurate` path:
- Loop iteration 1: `budget.exhausted()` → false → assess → `consume(1)` → `assessed=1` → softRefresh → end-of-iteration `budget.exhausted()` → **true → break**
- Result: `assessed = 1`, `soft_refreshed = 1`, `status = 'partial'`

So the assertion `assessed < 2` passes, but `assessed` is always exactly 1 — never 0 in the `still_accurate` path. The test is technically correct but the assertion range is wider than what actually happens. A tighter assertion `expect(result.assessed).toBe(1)` would be more informative and would catch a regression if someone changed the consume-before-check ordering. Similarly for the `partial` status row test.

**6. No test for `needs_update` + budget exhaustion mid-supersede**

The budget is checked before `regenerate()` (line ~394) but there's no test that exercises the case where budget is exhausted between `assess` and `regenerate` (i.e. after `consume(1)` for assess, before the regenerate call). This matters because `assessed++` is incremented but `superseded` is not, and the item is not written. The current behavior is correct but untested.

**7. `currentInputState` test for `entity` and `edge` types is missing**

Tests only exercise the `episode` input type. Given that `currentInputState` has four branches (`episode`, `entity`, `edge`, `projection`), the entity/edge/projection paths are untested. If the entity `status !== 'archived'` guard or the edge `invalidated_at === null` guard were broken, no test would catch it.

---

### Minor Nits

- **`_inputs` parameter in `buildNewProjectionData`** — it's unused and prefixed with `_`, which is fine, but the intent to use it for fingerprinting in the future should be documented with a comment, otherwise it's confusing.
- **`startReconciliationRun` fires before phase check** — if `phases` is `['discover']` only, the reconcile_runs row is inserted but the assess loop is entirely skipped. This is fine for the stub, but worth a note for when discover is implemented.
- **Conventional commit style** — the PR title uses `feat(core):` with a parenthetical scope. Project convention in CLAUDE.md is `type: description` with no scope. Minor, but worth keeping consistent across the history.

---

### Summary

| Area | Status |
|------|--------|
| Assess loop structure (stale filter, budget check, routing) | Correct |
| `softRefresh` — leaves valid_from/valid_until/invalidated_at untouched | Correct |
| `softRefresh` — updates `content_hash` in projection_evidence | Correct |
| `currentInputState` — fetches current content from substrate | Correct |
| reconciliation_runs lifecycle (insert→update, status values) | Correct |
| dryRun: skips writes, counts still tracked | Correct |
| Budget exhaustion: stops early, status='partial' | Correct |
| **`input_fingerprint` on superseded projection** | **Bug: should be recomputed, not taken from generator** |
| File length | 516 lines (1 over limit) |
| Test coverage: mocks vs real SQLite | Real SQLite in-memory — correct per project standards |
| Test coverage: `needs_update` + mid-loop budget exhaustion | Missing |
| Test coverage: entity/edge/projection input types | Missing |

rnwolfe (commented):
Blocking issue fixed: `buildNewProjectionData()` now accepts `input_fingerprint` as a parameter (computed via `recomputeCurrentFingerprint()` before the call) rather than falling back to `original.input_fingerprint`. Also fixed `assessedAt` to be captured after `generator.assess()` resolves.

406 tests pass, lint clean.

PR #82, opened 2026-04-10
feat(core): add projection invariant checks to verifyGraph()

### LINKED ISSUE #71: feat(core): verifyGraph() projection invariants
## Context

ADR-002 and [`format-v0.2.md`](../blob/main/docs/internal/specs/format-v0.2.md) § Verify invariants add three new invariants to `verifyGraph()`. These catch integrity violations that the insert-time checks in `project()` (#67) might miss if the file was produced by a buggy or older writer.

## What to build

Extend `verifyGraph()` in `packages/engram-core/src/format/verify.ts` with three new checks:

1. **Projection has evidence.** Every row in `projections` has at least one row in `projection_evidence` with `role='input'`. Violation → `ProjectionMissingEvidenceError { projection_id }`.

2. **Supersession chains terminate.** Walk `superseded_by` from every projection; detect cycles and unbounded chains. Implementation: recursive CTE with a depth guard. Violation → `ProjectionSupersessionCycleError { projection_id, cycle_path }`.

3. **Projection-dependency DAG.** The directed graph induced by `projection_evidence` rows where `target_type='projection'` must be acyclic. Implementation: recursive CTE from each projection following evidence edges to other projections. Violation → `ProjectionDependencyCycleError { cycle_path }`.

### Not a verify invariant (explicitly)

The fingerprint integrity check (`input_fingerprint` matches recomputation) is **not** a verify-time invariant. Read-time fingerprint mismatch is the staleness signal, not corruption. See [`format-v0.2.md`](../blob/main/docs/internal/specs/format-v0.2.md) § Verify invariants for rationale.

## Acceptance criteria

- [ ] Three new invariant checks added to `verifyGraph()` with typed error classes
- [ ] Checks are skipped cleanly on v0.1 files (projection tables absent)
- [ ] Unit test: projection without evidence → detected
- [ ] Unit test: supersession cycle (A→B→A) → detected
- [ ] Unit test: projection-dependency cycle via evidence → detected
- [ ] Unit test: healthy projection graph → no violations
- [ ] `bun test` remains under existing runtime budget

## Out of scope

- Fingerprint-based staleness (handled at read time by #68)

## Blocked by

- #66 schema migration v0.1 → v0.2

## References

- [`docs/internal/specs/format-v0.2.md`](../blob/main/docs/internal/specs/format-v0.2.md) § Verify invariants

## Documentation

- [ ] JSDoc on each new invariant check + the three typed error classes (`ProjectionMissingEvidenceError`, `ProjectionSupersessionCycleError`, `ProjectionDependencyCycleError`)
- [ ] JSDoc explains why fingerprint integrity is **not** a verify invariant (read-time staleness is the signal, not corruption — see `format-v0.2.md` § Verify invariants)
- [ ] CLAUDE.md: if there is a verify/invariants listing, append the three new projection invariants; otherwise leave as-is

rnwolfe:
## Summary

- Extends `verifyGraph()` in `packages/engram-core/src/format/verify.ts` with three new projection-layer invariant checks
- `checkProjectionEvidence`: detects projections with no `role='input'` evidence rows
- `checkProjectionSupersessionCycles`: uses a recursive CTE (depth-guarded at 10 000) to detect cycles in `superseded_by` chains
- `checkProjectionDependencyCycles`: walks the projection→projection dependency graph (via `projection_evidence` where `target_type='projection'`) and detects DAG cycles
- All three checks skip cleanly on v0.1 files (no `projections` table) via an `sqlite_master` existence probe

## Acceptance Criteria

- [x] `checkProjectionEvidence` — projection with no `role='input'` evidence → error violation
- [x] `checkProjectionSupersessionCycles` — A→B→A cycle → error violation
- [x] `checkProjectionDependencyCycles` — A depends on B depends on A → error violation
- [x] v0.1 file (projection tables absent) → checks skip, no violations
- [x] Healthy projection graph → no violations
- [x] All checks are severity `error`

## Test plan

- `packages/engram-core/test/format/verify-projections.test.ts` — 10 tests, all passing
- Existing `verify.test.ts` suite — all passing (no regressions)
- `engram-core` package builds cleanly

Closes #71

### DIFF
--- a/packages/engram-core/src/format/verify.ts
+++ b/packages/engram-core/src/format/verify.ts
@@ -277,7 +277,7 @@ function checkProjectionSupersessionCycles(graph: EngramGraph): Violation[] {
          SELECT c.start_id, p.superseded_by,
                 c.path || ',' || p.id,
                 c.depth + 1,
-                CASE WHEN instr(c.path, p.id) > 0 THEN 1 ELSE 0 END
+                CASE WHEN instr(',' || c.path || ',', ',' || p.id || ',') > 0 THEN 1 ELSE 0 END
          FROM chain c
          JOIN projections p ON p.id = c.current_id
          WHERE c.current_id IS NOT NULL
@@ -315,7 +315,7 @@ function checkProjectionDependencyCycles(graph: EngramGraph): Violation[] {
          SELECT d.start_id, pe.target_id,
                 d.path || ',' || pe.target_id,
                 d.depth + 1,
-                CASE WHEN instr(d.path, pe.target_id) > 0 THEN 1 ELSE 0 END
+                CASE WHEN instr(',' || d.path || ',', ',' || pe.target_id || ',') > 0 THEN 1 ELSE 0 END
          FROM dep d
          JOIN projection_evidence pe ON pe.projection_id = d.current_id
          WHERE pe.target_type = 'projection'
@@ -334,7 +334,8 @@ function checkProjectionDependencyCycles(graph: EngramGraph): Violation[] {
   for (const row of rows) {
     // Normalise the cycle path to the smallest rotation for dedup
     const parts = row.cycle_path.split(",");
-    const key = [...parts].sort().join(",");
+    // Dedup on sorted unique node set — a cycle A→B→A and B→A→B are the same cycle
+    const key = [...new Set(parts)].sort().join(",");
     if (!seen.has(key)) {
       seen.add(key);
       violations.push({
--- a/packages/engram-core/src/format/verify.ts
+++ b/packages/engram-core/src/format/verify.ts
@@ -227,6 +227,126 @@ function checkEmbeddingTargets(graph: EngramGraph): Violation[] {
   return violations;
 }
 
+/**
+ * Returns true if the projections table exists in this database.
+ * Used to skip projection checks on v0.1 files that lack projection tables.
+ */
+function hasProjectionsTable(graph: EngramGraph): boolean {
+  const row = graph.db
+    .query<{ cnt: number }, []>(
+      `SELECT COUNT(*) AS cnt FROM sqlite_master
+       WHERE type='table' AND name='projections'`,
+    )
+    .get();
+  return (row?.cnt ?? 0) > 0;
+}
+
+function checkProjectionEvidence(graph: EngramGraph): Violation[] {
+  if (!hasProjectionsTable(graph)) return [];
+
+  const rows = graph.db
+    .query<{ id: string }, []>(
+      `SELECT p.id FROM projections p
+       LEFT JOIN projection_evidence pe
+         ON p.id = pe.projection_id AND pe.role = 'input'
+       WHERE pe.projection_id IS NULL`,
+    )
+    .all();
+
+  return rows.map((row) => ({
+    check: "checkProjectionEvidence",
+    entity_or_edge_id: row.id,
+    message: `Projection '${row.id}' has no evidence rows with role='input'`,
+    severity: "error" as ViolationSeverity,
+  }));
+}
+
+function checkProjectionSupersessionCycles(graph: EngramGraph): Violation[] {
+  if (!hasProjectionsTable(graph)) return [];
+
+  // Use a recursive CTE with a depth guard to detect cycles in superseded_by chains.
+  // A cycle is detected when we visit a projection_id we have already seen in the chain
+  // (path contains it), or the depth exceeds the guard.
+  const rows = graph.db
+    .query<{ start_id: string; cycle_path: string }, []>(
+      `WITH RECURSIVE chain(start_id, current_id, path, depth, cycle) AS (
+         SELECT id, superseded_by, id, 1, 0
+         FROM projections
+         WHERE superseded_by IS NOT NULL
+         UNION ALL
+         SELECT c.start_id, p.superseded_by,
+                c.path || ',' || p.id,
+                c.depth + 1,
+                CASE WHEN instr(c.path, p.id) > 0 THEN 1 ELSE 0 END
+         FROM chain c
+         JOIN projections p ON p.id = c.current_id
+         WHERE c.current_id IS NOT NULL
+           AND c.cycle = 0
+           AND c.depth < 10000
+       )
+       SELECT start_id, path AS cycle_path
+       FROM chain
+       WHERE cycle = 1`,
+    )
+    .all();
+
+  return rows.map((row) => ({
+    check: "checkProjectionSupersessionCycles",
+    entity_or_edge_id: row.start_id,
+    message: `Projection supersession cycle detected starting from '${row.start_id}': ${row.cycle_path}`,
+    severity: "error" as ViolationSeverity,
+  }));
+}
+
+function checkProjectionDependencyCycles(graph: EngramGraph): Violation[] {
+  if (!hasProjectionsTable(graph)) return [];
+
+  // Walk the projection→projection dependency graph (via projection_evidence where
+  // target_type='projection') using a recursive CTE. Detect cycles.
+  const rows = graph.db
+    .query<{ start_id: string; cycle_path: string }, []>(
+      `WITH RECURSIVE dep(start_id, current_id, path, depth, cycle) AS (
+         SELECT DISTINCT pe.projection_id, pe.target_id,
+                pe.projection_id || ',' || pe.target_id,
+                1, 0
+         FROM projection_evidence pe
+         WHERE pe.target_type = 'projection'
+         UNION ALL
+         SELECT d.start_id, pe.target_id,
+                d.path || ',' || pe.target_id,
+                d.depth + 1,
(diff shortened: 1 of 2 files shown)


rnwolfe (commented):
## Review: projection invariant checks in verifyGraph()

Overall the approach is solid and consistent with the existing verify pattern. Two correctness issues worth addressing before merge, plus a few smaller notes.

---

### Bug 1 — `instr()` substring false-positives in cycle path detection (both CTEs)

Both `checkProjectionSupersessionCycles` and `checkProjectionDependencyCycles` use `instr(c.path, p.id) > 0` to detect revisited nodes. Because `path` is a comma-concatenated string and `instr` is a substring match (not a token match), a ULID that is a **substring of another ULID in the path** would produce a false positive. ULIDs are 26 chars and all upper-case Crockford base32, so a prefix collision is unlikely in practice, but it is a latent correctness hole.

The standard fix is to wrap the path and needle with delimiters before testing:

```sql
-- instead of:
CASE WHEN instr(c.path, p.id) > 0 THEN 1 ELSE 0 END

-- use:
CASE WHEN instr(',' || c.path || ',', ',' || p.id || ',') > 0 THEN 1 ELSE 0 END
```

Same change needed in both CTEs (lines ~280 and ~318 in verify.ts).

---

### Bug 2 — cycle deduplication key is wrong in `checkProjectionDependencyCycles`

The dedup block (lines ~336-346) sorts the comma-split parts alphabetically to produce a canonical key:

```ts
const key = [...parts].sort().join(",");
```

This does **not** produce the same key for the same cycle seen from different entry points. For a cycle A→B→A, one traversal yields path `A,B,A` → sorted parts `A,A,B`. The traversal starting at B yields `B,A,B` → sorted parts `A,B,B`. These are different keys, so both are emitted and the dedup does not fire.

A correct canonical form for a cycle is the **lexicographically smallest rotation** of the unique cycle members. A simpler safe approach: collect only the unique nodes in the cycle path, sort those, and use that as the key.

```ts
const parts = row.cycle_path.split(",");
const uniqueNodes = [...new Set(parts)].sort();
const key = uniqueNodes.join(",");
```

For A→B→A: unique nodes `{A, B}` → key `A,B`.  
For B→A→B: unique nodes `{A, B}` → key `A,B`. Same key — dedup fires correctly.

---

### Observation — `checkProjectionSupersessionCycles`: depth guard is safe but very large

The `c.depth < 10000` guard prevents infinite loops on degenerate inputs, but 10 000 recursive steps is high for a supersession chain. A depth of 100 or 1 000 would catch any realistic cycle and produce a less surprising `SQLITE_TOOBIG` boundary. Not a correctness issue, but worth a comment explaining the choice.

---

### Observation — error class names in PR description vs. spec

The review criteria reference `ProjectionMissingEvidenceError`, `ProjectionSupersessionCycleError`, `ProjectionDependencyCycleError`. These do not appear in the implementation (correctly — `verifyGraph()` returns `Violation` objects, it does not throw). The spec (resolved decision 4) names `ProjectionCycleError` only for the **insert-time** check in `project()`, which is a separate future concern. The PR description/criteria are slightly misleading but the implementation pattern is consistent with existing checks. No code change needed; just noting the terminology mismatch.

---

### Observation — commit scope

The single commit is `feat: add projection invariant checks to verifyGraph()`. The PR title correctly uses `feat(core):`. Per repo convention (conventional commits, as enforced by commitlint), adding the `(core)` scope to the commit message would keep it consistent with the PR title and the file scope.

---

### What's correct

- `checkProjectionEvidence`: the LEFT JOIN with `pe.role = 'input'` in the ON clause and `WHERE pe.projection_id IS NULL` is the right pattern — identical to `checkEntityEvidence` / `checkEdgeEvidence`.
- `hasProjectionsTable()`: querying `sqlite_master` is the correct v0.1 skip mechanism; all three checks guard on it.
- Test coverage is thorough: healthy case, each violation type, anchor-only evidence, linear chain (non-cycle), and v0.1 compat test that drops tables and wraps the raw DB.
- Both files are under 500 lines (verify.ts: 403, test: 362).
- All new violations correctly use `severity: "error"`.

---

**Summary**: two correctness bugs (instr substring matching, dedup key), one depth-guard comment nit, one scope nit on the commit. The instr issue is latent rather than immediately reproducible with real ULIDs, but the fix is a one-liner and should go in. The dedup bug will cause duplicate cycle violations in any graph with a 2-node cycle — that is immediately observable in the existing test (both A and B would start the traversal and both would be reported).

rnwolfe (commented):
## Review: projection invariant checks in verifyGraph()

Overall the approach is solid and consistent with the existing verify pattern. Two correctness issues worth addressing before merge, plus a few smaller notes.

---

### Bug 1 — instr() substring false-positives in cycle path detection (both CTEs)

Both checkProjectionSupersessionCycles and checkProjectionDependencyCycles use instr(c.path, p.id) > 0 to detect revisited nodes. Because path is a comma-concatenated string and instr is a substring match (not a token match), a ULID that is a substring of another ULID in the path would produce a false positive. ULIDs are 26 chars and all upper-case Crockford base32, so a prefix collision is unlikely in practice, but it is a latent correctness hole.

The standard fix is to wrap the path and needle with delimiters before testing:

```sql
-- instead of:
CASE WHEN instr(c.path, p.id) > 0 THEN 1 ELSE 0 END

-- use:
CASE WHEN instr(',' || c.path || ',', ',' || p.id || ',') > 0 THEN 1 ELSE 0 END
```

Same change needed in both CTEs (supersession check around line 280, dependency check around line 318 in verify.ts).

---

### Bug 2 — cycle deduplication key is wrong in checkProjectionDependencyCycles

The dedup block sorts the comma-split parts alphabetically to produce a canonical key:

```ts
const key = [...parts].sort().join(",");
```

This does NOT produce the same key for the same cycle seen from different entry points. For a 2-node cycle A->B->A, one traversal yields path "A,B,A" -> sorted parts ["A","A","B"] -> key "A,A,B". The traversal starting at B yields "B,A,B" -> sorted parts ["A","B","B"] -> key "A,B,B". Different keys, so dedup does not fire and both are emitted.

The existing cycle test (checkProjectionDependencyCycles test 1) only asserts that at least one violation is found, so this bug does not cause test failure — but it means two violations are reported for any 2-node cycle.

A correct fix: deduplicate on the sorted unique node set:

```ts
const uniqueNodes = [...new Set(parts)].sort();
const key = uniqueNodes.join(",");
```

For A->B->A: unique nodes {A, B} -> key "A,B".
For B->A->B: unique nodes {A, B} -> key "A,B". Dedup fires correctly.

---

### Observation — depth guard of 10 000 is very high

The c.depth < 10000 guard is safe (terminates), but 10 000 recursive CTE steps is high for a supersession chain in any realistic graph. 100 or 1 000 would be a more defensible limit and worth a comment explaining the choice.

---

### Observation — error class names in PR description vs. spec

The review criteria reference ProjectionMissingEvidenceError, ProjectionSupersessionCycleError, and ProjectionDependencyCycleError. These do not appear in the implementation, correctly so — verifyGraph() returns Violation objects and does not throw. The spec (resolved decision 4) names ProjectionCycleError only for the insert-time check in project(), which is a separate future concern. The implementation pattern is consistent with existing checks; the PR description's criteria section is slightly misleading but requires no code change.

---

### Nit — commit scope

The commit message is "feat: add projection invariant checks to verifyGraph()". The PR title correctly uses feat(core):. Repo convention (commitlint enforced) includes the package scope on commits that touch a specific package.

---

### What is correct

- checkProjectionEvidence: the LEFT JOIN with pe.role = 'input' in the ON clause, combined with WHERE pe.projection_id IS NULL, is the right pattern — exactly mirrors checkEntityEvidence and checkEdgeEvidence.
- hasProjectionsTable(): querying sqlite_master is the correct v0.1 skip mechanism; all three checks guard on it.
- Test coverage is thorough: healthy graph, linear chain (no cycle), missing evidence, anchor-only evidence (role='anchor' still violates), supersession cycle, no-supersession-links case, dependency cycle, acyclic dependency, and v0.1 compat test that drops the projection tables and wraps the raw DB.
- Both files are under 500 lines (verify.ts: 403, test: 362).
- All new violations correctly use severity: "error".
- The CTE termination logic is otherwise correct: the seed correctly feeds current_id = superseded_by, the recursive step joins on p.id = current_id and advances to p.superseded_by, and cycle=1 rows are selected in the final WHERE.

---

**Summary**: two correctness bugs to fix before merge. Bug 1 (instr substring matching) is latent with real ULIDs but the fix is a one-liner. Bug 2 (dedup key) is immediately observable — any 2-node cycle reports two violations instead of one. Both fixes are small.

rnwolfe (commented):
Review: two bugs found — see comment

rnwolfe (commented):
Both bugs fixed:
1. `instr()` now uses delimited matching: `instr(',' || path || ',', ',' || id || ',')` — prevents ULID prefix false positives in both CTEs.
2. Dependency cycle dedup key now uses `[...new Set(parts)].sort().join(',')` — a cycle A→B→A and B→A→B now produce the same key.

416 tests pass, lint clean.

PR #83, opened 2026-04-10
feat: add engram export wiki subcommand

### LINKED ISSUE #74: feat(cli): engram export wiki — materialize projections to markdown folder
## Context

ADR-002 promises a wiki export: a folder of frontmatter-prefixed markdown files, directly consumable by Jekyll/Hugo/Obsidian and trivially git-trackable. Because projection bodies already store markdown with mandatory YAML frontmatter ([`format-v0.2.md`](../blob/main/docs/internal/specs/format-v0.2.md) § Body format), the export is essentially a literal file copy with a sane filename scheme — that is the whole point of the frontmatter convention.

## What to build

A new command in `packages/engram-cli/src/commands/export.ts` (extend if exists, create otherwise):

```
engram export wiki --out <dir> [--scope <filter>] [--include-superseded]
```

### Behavior

1. Resolve the `.engram` file via existing CLI conventions.
2. Query active projections (or all, if `--include-superseded`), filtered by `--scope` (same shape as `engram reconcile`).
3. For each projection: write `<out>/<kind>/<anchor-slug>__<short-id>.md` containing the projection body verbatim. The frontmatter on the body is the source of truth.
4. Write an `<out>/index.md` with a simple list grouped by kind, linking to each file.
5. Print summary: kinds counted, files written, output dir.

### Filename scheme

- `<kind>/` subdirs (e.g. `entity_summary/`, `decision_page/`)
- Anchor slug: derived from anchor's canonical name if available (entity), or anchor type+id otherwise; sanitized for filesystems
- Short id: last 8 chars of the projection ULID for uniqueness across same-anchor variants

### Round-trip property

Because frontmatter is mandatory and `engram export wiki` is a verbatim body copy, `engram export wiki && cat <file>` yields exactly what's in the database. No reformatting, no field renaming. This makes the export load-bearing for Jekyll/Hugo consumers.

## Acceptance criteria

- [ ] Integration test: export against a temp `.engram` with seeded projections produces the expected file tree
- [ ] Integration test: `--include-superseded` includes invalidated projections; default does not
- [ ] Integration test: `--scope kind:entity_summary` filters correctly
- [ ] Integration test: round-trip — exported body equals `projections.body` byte-for-byte
- [ ] Output dir is created if missing; existing files are overwritten with a warning, not errored
- [ ] `index.md` lists all exported projections grouped by kind

## Out of scope

- Bidirectional sync (re-importing edited markdown back into the `.engram`) — deferred indefinitely
- HTML/site generation (consumers do that)

## Blocked by

- #66 schema migration v0.1 → v0.2 (need projection rows to export)

## References

- [`docs/internal/specs/projections.md`](../blob/main/docs/internal/specs/projections.md) § Body format
- [`docs/internal/specs/format-v0.2.md`](../blob/main/docs/internal/specs/format-v0.2.md) § Body format

## Documentation

- [ ] Command help text (`engram export wiki --help`) covers all flags and describes the round-trip property (exported body equals `projections.body` byte-for-byte)
- [ ] Help text calls out that existing files in `--out` are overwritten with a warning
- [ ] CLAUDE.md: add `export wiki` as a subcommand under the Core CLI command list (or update the existing `export` entry to note the new subcommand)

rnwolfe:
## Summary

- Extends `engram export` with a `wiki` subcommand that materializes projections
  to a markdown folder structure
- Each projection body is written verbatim to `<out>/<kind>/<anchor-slug>__<short-id>.md`
- Generates an `index.md` listing all exported projections grouped by kind
- Supports `--scope <kind>` to filter by projection kind and `--include-superseded`
  to include invalidated projections (default: active only)

## Acceptance Criteria

- [x] Command registered: `engram export wiki`
- [x] Integration test: export seeds projections, checks file tree exists
- [x] Integration test: `--include-superseded` includes invalidated projections; default excludes them
- [x] Integration test: `--scope entity_summary` filters to only that kind
- [x] Integration test: round-trip — exported body equals `projection.body` byte-for-byte
- [x] Output dir created if missing; overwrite with warn
- [x] `index.md` lists projections grouped by kind

## Test plan

- [x] All 9 integration tests in `packages/engram-cli/test/commands/export-wiki.test.ts` pass
- [x] Full test suite: 547 pass, 4 fail (4 pre-existing `engram-web` failures unrelated to this PR)
- [x] Lint passes (`bun run lint` exits 0)

## Implementation notes

The wiki subcommand uses `this.optsWithGlobals()` to inherit the `--db` option from
the parent `export` command, avoiding a Commander option-shadowing issue where a
child `--db` option would be ignored in favour of the parent's.

Closes #74

### DIFF
--- a/packages/engram-cli/src/commands/export.ts
+++ b/packages/engram-cli/src/commands/export.ts
@@ -36,15 +36,32 @@ interface EpisodeRow {
 // ─── Helpers ─────────────────────────────────────────────────────────────────
 
 /**
- * Sanitize a string to be filesystem-safe.
- * Replaces non-alphanumeric chars with '-', truncates to 64 chars.
+ * Sanitize an anchor string to be filesystem-safe.
+ * Replaces non-alphanumeric chars (except underscores) with '-', truncates to 64 chars.
  */
 function toSlug(value: string): string {
   return value
-    .replace(/[^a-zA-Z0-9]/g, "-")
+    .replace(/[^a-zA-Z0-9_]/g, "-")
     .replace(/-+/g, "-")
     .replace(/^-|-$/g, "")
     .slice(0, 64);
+}
+
+/**
+ * Sanitize a kind string for use as a directory path component.
+ * Strips path-traversal chars (../, slashes, null) while preserving underscores
+ * so that kind names like "entity_summary" remain readable.
+ */
+function toKindPath(kind: string): string {
+  return (
+    kind
+      .replace(/[/\\]/g, "-")
+      .replace(/\.\./g, "-")
+      .replace(/\0/g, "")
+      .replace(/-+/g, "-")
+      .replace(/^-|-$/g, "")
+      .slice(0, 64) || "unknown"
+  );
 }
 
 /**
@@ -59,11 +76,11 @@ function shortId(id: string): string {
  * Pattern: <kind>/<anchor-slug>__<short-id>.md
  */
 function projectionFilename(projection: Projection): string {
+  // Sanitize kind as a path component — prevents directory traversal
+  // if a malicious DB row contains kind = "../evil".
+  const kindSlug = toKindPath(projection.kind);
   const anchorSlug = toSlug(projection.anchor_id ?? projection.anchor_type);
-  return path.join(
-    projection.kind,
-    `${anchorSlug}__${shortId(projection.id)}.md`,
-  );
+  return path.join(kindSlug, `${anchorSlug}__${shortId(projection.id)}.md`);
 }
 
 // ─── Wiki subcommand ──────────────────────────────────────────────────────────
@@ -131,7 +148,7 @@ function registerWiki(exportCmd: Command): void {
         for (const projection of projections) {
           const relPath = projectionFilename(projection);
           const absPath = path.join(outDir, relPath);
-          const kindDir = path.join(outDir, projection.kind);
+          const kindDir = path.join(outDir, toKindPath(projection.kind));
 
           fs.mkdirSync(kindDir, { recursive: true });
 
--- a/packages/engram-cli/src/commands/export.ts
+++ b/packages/engram-cli/src/commands/export.ts
@@ -2,13 +2,20 @@
  * export.ts — `engram export` command.
  *
  * Exports the graph as deterministic JSONL (one object per line, sorted by ID)
- * or as markdown summaries.
+ * or as markdown summaries. Includes `engram export wiki` subcommand.
  */
 
+import * as fs from "node:fs";
 import * as path from "node:path";
 import type { Command } from "commander";
-import type { EngramGraph } from "engram-core";
-import { closeGraph, findEdges, findEntities, openGraph } from "engram-core";
+import type { EngramGraph, Projection } from "engram-core";
+import {
+  closeGraph,
+  findEdges,
+  findEntities,
+  listActiveProjections,
+  openGraph,
+} from "engram-core";
 
 interface ExportOpts {
   format: string;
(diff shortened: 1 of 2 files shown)


rnwolfe (commented):
## Review: `feat: add engram export wiki subcommand`

Overall this is solid work — clean structure, good test coverage, and the core invariants (round-trip fidelity, path safety, evidence-backed provenance) are handled correctly. I have a few findings below, ranging from a real bug to minor style notes.

---

### Bug: `--include-superseded` + `--scope` silently exports nothing

In the `--include-superseded` branch the SQL query is built by hand with a conditional `WHERE` clause:

```ts
const kindFilter = opts.scope ? "WHERE kind = ?" : "";
const params = opts.scope ? [opts.scope] : [];
projections = graph.db
  .query<Projection, string[]>(
    `SELECT * FROM projections ${kindFilter} ORDER BY created_at DESC`,
  )
  .all(...params);
```

`graph.db.query<T, P>()` is Bun's `Database.query()`, which **caches the prepared statement by SQL string**. When `opts.scope` is truthy, the string changes (includes `WHERE kind = ?`), so a different statement is cached — that part is fine. But the parameter spread `.all(...params)` passes a `string[]` while the generic is typed as `string[]`, so no type error surfaces, but if `params` is empty the statement is prepared with no parameter slots and `.all()` gets zero args, which is correct. **However**, if `params` has one element and the statement's placeholder count matches, this works. The logic itself is actually correct here, but it is fragile and inconsistent with how the rest of the codebase uses `listActiveProjections` with structured options. Consider replacing the raw SQL block with:

```ts
const allResults = graph.db
  .query<Projection, (string | null)[]>(
    opts.scope
      ? "SELECT * FROM projections WHERE kind = ? ORDER BY created_at DESC"
      : "SELECT * FROM projections ORDER BY created_at DESC",
  )
  .all(...(opts.scope ? [opts.scope] : []));
projections = allResults;
```

This is essentially the same, but the inconsistency with the `listActiveProjections` path (which passes `{ kind: opts.scope }`) makes it a potential maintenance hazard. More importantly, the **real bug** is that `--include-superseded` with `--scope` will fail if the caller ever passes a scope value containing a SQL metacharacter — but because `opts.scope` comes from a CLI flag (user-controlled input), this is only a concern in a programmatic context, not a security issue per se. No injection path exists here since `better-sqlite3` always uses parameterized queries; the string interpolation is only for the keyword `WHERE`, not the value. So this is **low severity**, but worth a comment.

---

### Security: path traversal is correctly mitigated

`toSlug()` aggressively replaces everything outside `[a-zA-Z0-9]` with `-` and then `path.join()` assembles the path. A `kind` value of `../evil` becomes `--evil` after slugification, so a projection whose `kind` field contains `../` **cannot** traverse outside `outDir`. The `projection.kind` value is used directly as a directory name (`path.join(outDir, projection.kind)`) **without slugification**, however:

```ts
const kindDir = path.join(outDir, projection.kind);
fs.mkdirSync(kindDir, { recursive: true });
```

`projection.kind` is **not** passed through `toSlug()`. If a malicious or malformed DB row has `kind = "../evil"`, `path.join(outDir, "../evil")` resolves to a sibling directory of `outDir` and `mkdirSync` will happily create it. The same raw `projection.kind` is used in `projectionFilename()` via `path.join(projection.kind, ...)`, which also doesn't slug it. 

The `anchor_id`/`anchor_type` values are correctly slugged inside `projectionFilename()` → `toSlug()`. But `kind` is never sanitized before being used as a path component.

**Recommendation**: apply `toSlug()` to `projection.kind` when constructing directory/file paths, or assert that `kind` is already slug-safe (only `[a-z0-9_]` as the existing projections use).

```ts
const kindSlug = toSlug(projection.kind);
const relPath = path.join(kindSlug, `${anchorSlug}__${shortId(projection.id)}.md`);
```

---

### Correctness: `wiki` subcommand registration confirmed

`registerWiki(exportCmd)` is called at the end of `registerExport()`. The test `"command is registered under export"` verifies this at runtime. ✓

---

### Correctness: file tree written correctly

`fs.mkdirSync(outDir, { recursive: true })` creates the top-level dir, and `fs.mkdirSync(kindDir, { recursive: true })` creates the kind subdirectory. Files are written to `path.join(outDir, relPath)` which is `outDir/<kind>/<slug>__<shortid>.md`. ✓

---

### Correctness: `index.md` generated and grouped by kind

The `writtenByKind` map is sorted before iteration (`[...writtenByKind.entries()].sort()`), so groups appear alphabetically. Each group has the count and links. One minor issue: the links in `index.md` use `entry.filename` (the relative path including the kind prefix, e.g. `entity_summary/alice__ABCD1234.md`), which is correct for a flat wiki directory structure. ✓

---

### Correctness: round-trip fidelity

`fs.writeFileSync(absPath, projection.body, { encoding: "utf8" })` writes the body verbatim with no transformation. The test asserts `written === body` byte-for-byte. ✓

---

### Correctness: `--include-superseded` includes invalidated projections

The raw SQL `SELECT * FROM projections` with no `invalidated_at IS NULL` filter correctly includes invalidated rows. The default path uses `listActiveProjections` which filters to `invalidated_at IS NULL`. ✓

---

### Correctness: `--scope` filters by kind

In both code paths (`listActiveProjections({ kind: opts.scope })` and the raw SQL `WHERE kind = ?`), the scope filter is applied. ✓

---

### Correctness: overwrite behavior — warns, doesn't error

`fs.existsSync(absPath)` check before `writeFileSync` emits `console.warn(...)` but does not throw or exit. The test confirms no exception is thrown on second export. ✓

One minor nit: when overwriting `index.md`, the message is `"warn: overwriting existing file index.md"` but for projection files it's `"warn: overwriting existing file ${relPath}"`. These are consistent in style. ✓

---

### Test coverage

9 integration tests covering:
1. Command registration ✓
2. Creates output dir and files ✓
3. Round-trip fidelity ✓
4. Default excludes invalidated ✓
5. `--include-superseded` includes invalidated ✓
6. `--scope` filters by kind ✓
7. Creates output dir if missing ✓
8. `index.md` grouped by kind ✓
9. Overwrite emits warn, does not error ✓

All 5 acceptance criteria from the issue (file tree, `--include-superseded`, `--scope`, round-trip, overwrite-warn) are tested. The PR description says 9 tests; the file contains 9 `it()` blocks. ✓

One gap: there is no test that verifies the `--include-superseded --scope` combination (the potentially fragile raw-SQL path with a kind filter). Worth adding.

---

### File size

- `export.ts`: 297 lines (under 500 ✓)
- `export-wiki.test.ts`: 366 lines (under 500 ✓)

---

### Conventional commits

Single commit: `feat: add engram export wiki subcommand` — lowercase, no scope, no trailing period. ✓

---

### Summary

| Check | Result |
|-------|--------|
| `wiki` subcommand registered | ✓ |
| File tree written correctly | ✓ |
| `index.md` generated, grouped by kind, links correct | ✓ |
| Round-trip fidelity | ✓ |
| `--include-superseded` includes invalidated | ✓ |
| `--scope` filters by kind | ✓ |
| Overwrite warns, does not error | ✓ |
| Test coverage (all 9 cases) | ✓ |
| Files under 500 lines | ✓ |
| Conventional commits | ✓ |
| **Path traversal via `kind` field** | ⚠️ `projection.kind` used unsanitized as dir path component |
| `--include-superseded + --scope` gap | ℹ️ No test; raw SQL path is correct but fragile |

The path traversal issue via `projection.kind` is worth fixing before merge — it's a one-liner (`toSlug(projection.kind)` when building paths). The rest of the implementation is correct and well-tested.

rnwolfe (commented):
Security issue fixed: added `toKindPath()` that strips path-traversal chars (`../`, `/`, `\`, null bytes) while preserving underscores so kind names like `entity_summary` stay readable. All `toSlug(projection.kind)` usages replaced with `toKindPath()`.

19 tests pass, lint clean.

PR #84, opened 2026-04-10
feat: kind catalog for projection layer (v0.2 initial kinds)

rnwolfe:
## Summary

- **Spec doc** at `docs/internal/specs/projection-kinds.md`: documents the catalog format (YAML, with full justification), location convention, XDG override path, required fields, loading semantics, and discover-phase prompt integration
- **Four built-in kind files** at `packages/engram-core/src/ai/kinds/`: `entity_summary`, `decision_page`, `contradiction_report`, `topic_cluster` — each with all required fields and concrete `when_to_use` guidance
- **TypeScript loader** at `packages/engram-core/src/ai/kinds.ts`: `loadKindCatalog()` function, `KindEntry` interface, `KindCatalog` type, `KindValidationError`, built-in YAML parser (no new dependencies), XDG override merge
- **`packages/engram-core/src/index.ts`** updated to export `KindCatalog`, `KindEntry`, `KindValidationError`, `loadKindCatalog`, and `AnchorTypeName`
- **`CLAUDE.md`** updated with `packages/engram-core/src/ai/kinds/` and `packages/engram-core/src/ai/kinds.ts` in the Key Files table, including XDG override path documentation

## Acceptance Criteria

- [x] Spec doc exists at `docs/internal/specs/projection-kinds.md`
- [x] Catalog format justified (YAML: flat structure, human-editable, multiline strings, no new runtime dependency)
- [x] All four kinds have catalog files with all required fields populated
- [x] `when_to_use` guidance is concrete enough for the discover prompt to use directly (each entry has 4–5 specific triggering conditions)
- [x] CLAUDE.md updated with `packages/engram-core/src/ai/kinds/` in Key Files table
- [x] CLAUDE.md mentions XDG override path (`$XDG_CONFIG_HOME/engram/kinds/`)

## Test plan

- [x] `bun run lint` — passes (0 errors)
- [x] `bun run build` — all packages build successfully
- [x] `bun test` — 570 tests pass, 0 failures

Closes #64

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #85, opened 2026-04-10
feat: engram project command for explicit projection authoring

### LINKED ISSUE #72: feat(cli): engram project — explicit projection authoring command
## Context

`engram project` is the CLI surface for the explicit `project()` authoring primitive (#67). It is the deterministic escape hatch — users can author a specific projection on a specific anchor with an explicit input set, without invoking the discover phase.

## What to build

A new command in `packages/engram-cli/src/commands/project.ts`:

```
engram project --kind <kind> --anchor <type:id> [--input <type:id>]... [--prompt <template>] [--dry-run]
```

### Behavior

1. Resolve the `.engram` file via existing CLI conventions.
2. Build the `ProjectionGenerator` from the configured AI provider (respects `NullProvider` → error "no AI provider configured for projection authoring").
3. Call `project()` (#67) with the resolved inputs.
4. Print a concise summary: kind, anchor, input count, resulting projection id, whether it was an idempotent no-op or a fresh author or a supersession.
5. `--dry-run` runs through step-through-validation and prints what *would* be authored (including fingerprint and resolved inputs) without writing.

### Input resolution

- `--input entity:01HX...` — directly specify an input by id
- If `--input` is omitted, read a sensible default set based on the anchor: for `entity:<id>`, include the entity, all evidence episodes, and all edges touching it.
- Multiple `--input` flags compose.

### Exit codes

- `0` — projection authored, superseded, or idempotent no-op
- `1` — generator failure, evidence resolution failure, cycle detected
- `2` — usage error

## Acceptance criteria

- [ ] Integration test: happy path authors a projection against a temp `.engram` file using a recording-mode generator
- [ ] Integration test: re-running with identical inputs is an idempotent no-op
- [ ] Integration test: `--dry-run` does not write
- [ ] Unit test: usage errors on invalid `--anchor` / `--input` strings
- [ ] Human-readable output; no stack traces on expected errors

## Out of scope

- Discover phase (#70)
- MCP tool surface (separate issue)

## Blocked by

- #67 `project()` operation

## References

- [`docs/internal/specs/projections.md`](../blob/main/docs/internal/specs/projections.md) § Operation 1

## Documentation

- [ ] Command help text (`engram project --help`) covers all flags with at least one worked example per major mode (default inputs, explicit `--input`, `--dry-run`)
- [ ] Error messages are actionable: "no AI provider configured" tells the user how to configure one
- [ ] CLAUDE.md: add `project` to the Core CLI command list in File Organization (alongside `add`, `search`, `show`, etc.)

rnwolfe:
## Summary

- Adds `engram project` CLI subcommand that surfaces the core `project()` operation
- Supports `--kind`, `--anchor`, `--input` (repeatable), `--dry-run`, and `--db` flags
- When `--input` is omitted for an `entity:<id>` anchor, defaults to the entity + all evidence episodes + all touching edges
- Detects NullGenerator early and emits a user-friendly error pointing to `ENGRAM_AI_PROVIDER`
- Idempotent re-runs (same fingerprint) are detected and reported as `status: idempotent`
- Exit codes: 0 = success/idempotent, 1 = generator/resolution failure, 2 = usage error

## Acceptance Criteria

- [x] Command registered in CLI root (`cli.ts`)
- [x] Integration test: happy path authors a projection against a temp `.engram` file using `AnthropicGenerator` stub (recording-mode)
- [x] Integration test: re-running with identical inputs is an idempotent no-op (status reports "idempotent")
- [x] Integration test: `--dry-run` does not write any projection rows
- [x] Unit test: usage errors on invalid `--anchor` / `--input` strings (exit code 2)
- [x] Human-readable output; no stack traces on expected errors
- [x] CLAUDE.md: added `project` to the Core CLI command list in File Organization

## Test plan

- `bun test packages/engram-cli/test/commands/project.test.ts` — 10 tests pass
- `bun test` — 580 tests pass, 0 failures
- `bun run build` — all packages build successfully
- `bun run lint` — no errors

Closes #72


rnwolfe (commented):
## Code Review — PR #85: `engram project` command

Overall: solid implementation. The command integrates cleanly with the core `project()` API, the happy path works, and test coverage is reasonable. There are several bugs and gaps that should be addressed before merge, a couple of them blocking.

---

### BLOCKING issues

#### 1. Idempotence detection is unreliable (time-based heuristic, not fingerprint-based)

**File:** `packages/engram-cli/src/commands/project.ts`, lines ~288–297

```ts
const beforeCallMs = Date.now();
// ...
const projectionCreatedAt = new Date(projection.valid_from).getTime();
const wasIdempotent = projectionCreatedAt < beforeCallMs - 500;
```

This is incorrect. The core `project()` already encodes the idempotent signal cleanly: when fingerprints match it returns the *existing* projection without modification. The CLI then throws away that signal and replaces it with a wall-clock heuristic with a 500 ms fudge factor that will misfire:

- On a fast machine, the first call might complete in < 500 ms and be misreported as "idempotent."
- On a slow machine (disk, cold cache), a genuine first write could take > 500 ms and be misreported as "idempotent" if `valid_from` was a few seconds ago.
- The test works only because it explicitly calls `setTimeout(r, 600)` between runs — removing that sleep would cause the test to fail.

The correct approach: compare the returned `projection.input_fingerprint` against the fingerprint of the *inputs passed to this call*. The same fingerprint `computeFingerprint(resolved)` the core already computed is available via the returned projection. Alternatively, compare `projection.id` or `projection.valid_from` against the `beforeCallMs` only if `project()` documents a clear semantic — but it does not.

The simplest fix is to check whether a projection with the returned `id` already existed before the call:

```ts
// Before calling project():
const existingRow = graph.db
  .query<{ id: string }, [string, string | null, string]>(
    `SELECT id FROM projections
      WHERE anchor_type = ? AND anchor_id IS ? AND kind = ?
        AND invalidated_at IS NULL LIMIT 1`,
  )
  .get(anchor_type, anchor_id, kind);

const projection = await project(graph, { ... });
const wasIdempotent = !!existingRow && existingRow.id === projection.id;
```

This is deterministic, has no race, and requires no sleep in tests.

#### 2. `resolveDefaultInputs()` silently returns empty array when entity doesn't exist

**File:** `packages/engram-cli/src/commands/project.ts`, lines ~148–157

When `anchor.type === "entity"` and the entity ID doesn't exist, `resolveDefaultInputs()` returns `[{ type: "entity", id: anchorId }]` (the entity itself) plus zero evidence and zero edges. The downstream `inputs.length === 0` guard in the action handler will never trigger because the entity input is always pushed first — so execution proceeds to `project()` which will call `resolveInputs()`, which will throw `ProjectionInputMissingError` with a confusing message rather than the friendly "anchor has no evidence" error.

This is not catastrophic (the user still gets an error), but:
- The error message is misleading: "entity:<id> not found in substrate" rather than "the anchor entity does not exist."
- A missing entity should be caught at the CLI layer before opening the generator.

Recommended fix: query the entity table at the top of `resolveDefaultInputs()` and return an empty array (or throw a `UsageError`) if the entity is not found, before pushing the entity input.

---

### Non-blocking but should fix before merge

#### 3. `--kind` is not sanitized against path traversal in `resolveDefaultInputs()` / `project()` call

The concern here is the same one that was fixed in the export wiki command (PR #84, commit `78d52d0`). The `--kind` value is passed directly to `project()` which stores it in the database. The core already handles this, but worth verifying `kind` is normalized by the CLI to prevent values like `../foo` or `entity_summary/../../../etc/passwd` being stored. The core's `project()` does not validate `kind` format. A simple check in the action handler would close this:

```ts
if (!/^[a-zA-Z0-9_\-]+$/.test(opts.kind)) {
  console.error(`Error: invalid --kind "${opts.kind}": must match [a-zA-Z0-9_-]+`);
  process.exit(2);
}
```

#### 4. `AnthropicGenerator` is a stub — happy-path test doesn't verify AI output

**File:** `packages/engram-cli/test/commands/project.test.ts`, the "happy path" describe block

The comment says "AnthropicGenerator is the stub that returns valid output without requiring an API key." The test sets `ENGRAM_AI_PROVIDER=anthropic` to avoid `NullGenerator`, which works — but it also means no real AI call is tested. The test verifies `projection.kind === "entity_summary"`, but the stub hardcodes `kind: "generated"` in its frontmatter, and the core reads the kind from the DB opts not the frontmatter. This discrepancy is fine for the current purpose, but the comment and test description are slightly misleading — the test is really just exercising the pipeline plumbing, not AI generation.

Minor: document this explicitly so a future author doesn't swap in a real `AnthropicGenerator` expecting it to work in CI.

#### 5. `process.exit()` in tests is patched but not type-safe

**File:** `packages/engram-cli/test/commands/project.test.ts` (repeated pattern)

The test overrides `process.exit` by casting with `(process as any).exit = ...`. This is idiomatic for this kind of test but it's done in at least 8 places with copy-pasted boilerplate. Consider a shared helper:

```ts
function captureExit(): { exitCode: () => number | null; restore: () => void } {
  let code: number | null = null;
  const orig = process.exit;
  (process as any).exit = (c: number) => { code = c; throw new Error(`exit(${c})`); };
  return { exitCode: () => code, restore: () => { (process as any).exit = orig; } };
}
```

This would halve the test boilerplate and reduce the risk of a `finally` block failing to restore the original.

#### 6. No test for `resolveDefaultInputs()` with a valid entity anchor

The test suite covers:
- invalid anchor format (exit 2)
- invalid input type (exit 2)
- dry-run (no write)
- NullGenerator (exit 1)
- happy path with explicit `--input`
- idempotent re-run

Missing: a test where `--input` is omitted and the anchor is `entity:<id>`. This is the default-input-resolution path, which hits the `entity_evidence` query and `findEdges()` calls. It's the most complex path in `resolveDefaultInputs()` and has zero test coverage. At minimum a unit test of `resolveDefaultInputs()` directly, or an integration test via the CLI with `--anchor entity:<id>` and no `--input`, would be valuable.

#### 7. `createGenerator()` ignores `OllamaProvider` / `GeminiProvider`

**File:** `packages/engram-cli/src/commands/project.ts`, lines ~196–205

The `ENGRAM_AI_PROVIDER` env var is used elsewhere in the codebase to select among `anthropic`, `ollama`, `gemini`, and `null`. The `createGenerator()` function here only handles `"anthropic"` and falls through to `NullGenerator` for anything else. A user who sets `ENGRAM_AI_PROVIDER=ollama` will get a confusing "no AI provider configured" error rather than an Ollama generator.

This is likely out of scope for the issue (which specifically targets Anthropic), but the `default:` label in the switch implies "everything else is NullGenerator" without documenting the limitation. The error message should say:

> "Set ENGRAM_AI_PROVIDER=anthropic to use engram project. (ollama and gemini generators are not yet supported for direct projection authoring.)"

Or, if `OllamaGenerator` / `GeminiGenerator` equivalents of `ProjectionGenerator` exist, add the cases.

---

### Minor / nits

- **`--dry-run` with `anchor: none` and empty `--input`:** the dry-run path exits before the "no inputs found" guard, so `--dry-run --anchor none` with no `--input` would print a dry-run summary with 0 inputs then `process.exit(0)`. No actual write occurs, which is correct, but the output is misleading since a real run would fail.

- **`closeGraph()` not called on the `UsageError` path:** The graph is opened before the anchor/input parsing (which happens before opening the graph), so this is fine — graph is opened after the parse, and the parse throws before `openGraph()`. Actually, looking more carefully: the parse happens at the top of `action()` before `openGraph()`, so no leak. This is fine.

- **Commit `feat: add engram export wiki subcommand` in CLAUDE.md:** The PR updates the CLAUDE.md command list line to add `project`, which is correct.

---

### Summary by severity

| Severity | Issue |
|----------|-------|
| **Blocking** | Idempotence detection uses wall-clock heuristic instead of fingerprint comparison |
| **Blocking** | `resolveDefaultInputs()` proceeds with a missing entity without a clear error |
| Should fix | `--kind` not validated against unsafe characters |
| Should fix | `createGenerator()` silently falls to NullGenerator for ollama/gemini |
| Should fix | No test coverage for default-input resolution with entity anchor |
| Nit | Test boilerplate for `process.exit` is heavily duplicated |
| Nit | Happy-path test comment implies real AI call but uses stub |

The two blocking issues should be fixed before merge. The rest can be addressed as follow-up issues or in the same PR.

rnwolfe (commented):
Fixed: reliable idempotence via ID comparison, entity-not-found guard, kind validation, ollama/gemini generator support, added test for default input resolution.

rnwolfe (commented):
Fixed: reliable idempotence via ID comparison, entity-not-found guard, kind validation, ollama/gemini generator support, added test for default input resolution.

PR #87, opened 2026-04-10
feat: reconcile discover phase with emergent projection authoring

rnwolfe:
## Summary

- Implements the `discover` phase of `reconcile()` from the projections spec (Operation 2, § discover phase)
- Extends `ProjectionGenerator` with `discover()` method and new types (`SubstrateDelta`, `ActiveProjectionSummary`, `ProjectionProposal`)
- `NullGenerator.discover()` returns `[]`; `AnthropicGenerator.discover()` is a stub returning `[]`
- Cursor semantics: dry runs don't advance the cursor; partial runs advance to what was successfully authored

### Discover phase flow (in `reconcile.ts`)
1. Resolve cursor: `completed_at` of last non-dry-run reconcile with same scope
2. Compute substrate delta since cursor (episodes via `ingested_at`, entities/edges via `created_at`)
3. Load active-projection catalog (lightweight summaries — no bodies)
4. Load kind catalog via `loadKindCatalog()`
5. Call `generator.discover({ delta, catalog, kinds })` → `ProjectionProposal[]`
6. Validate each proposal (kind in catalog, non-empty inputs, valid types)
7. Author valid proposals via `project()`, increment `discovered`
8. Budget exhaustion mid-phase sets `status='partial'` and records `error` reason

### New types exported from `engram-core`
- `SubstrateDelta` / `SubstrateDeltaItem` — delta context for the discover call
- `ActiveProjectionSummary` — lightweight projection catalog entry
- `ProjectionProposal` — `{ kind, anchor, inputs, rationale }` from generator

## Tests added (13 new tests)
- Integration: seed substrate, recording-mode generator, assert kinds authored and `projections_discovered` updated
- Integration: discover phase skipped when `phases=['assess']`
- Integration: dry run counts proposals but does not persist or advance cursor
- Integration: cursor semantics — second run only sees new substrate delta
- Integration: substrate delta contains episodes, entities, and edges
- Integration: catalog is passed to generator
- Unit: malformed proposals rejected — unknown kind, empty inputs, invalid input type, invalid anchor type
- Budget: exhaustion during discover sets `status='partial'` with error reason

## Acceptance criteria

- [x] `discover` phase in `reconcile.ts` behind phase switch (`phases.includes('discover')`)
- [x] `ProjectionGenerator.discover()` method added; `NullGenerator` returns `[]`
- [x] Cursor: dry runs don't advance; partial runs advance to what was successfully authored
- [x] `--max-cost` respected; exhaustion sets `status='partial'` and records error reason
[...]

PR #88, opened 2026-04-10
feat: engram reconcile command for two-phase projection maintenance

### LINKED ISSUE #73: feat(cli): engram reconcile — two-phase projection maintenance loop
## Context

`engram reconcile` is the CLI surface for the two-phase `reconcile()` operation. It is the primary authoring + maintenance loop for the projection layer: assess existing projections against current substrate, then discover new ones from the substrate delta.

## What to build

A new command in `packages/engram-cli/src/commands/reconcile.ts`:

```
engram reconcile [--phase assess|discover|both] [--scope <filter>] [--max-cost <n>] [--dry-run]
```

### Behavior

1. Resolve the `.engram` file. Error if no AI provider is configured.
2. Call `reconcile()` with the phase selection, scope filter, cost budget, and dry-run flag.
3. Stream progress updates per-phase (projections checked, refreshed, superseded, discovered) using `@clack/prompts`.
4. On completion, print a structured summary and the `reconciliation_runs.id` for auditing.
5. On `--max-cost` exhaustion mid-run, print which phase exhausted and which cursor was saved.

### Scope filter

`--scope` accepts simple filters that map to the `reconciliation_runs.scope` column, e.g. `kind:entity_summary`, `anchor:entity:01HX...`. Scope shapes the cursor and the active-projection catalog loaded during discover.

### Defaults

- `--phase both` by default (assess first, then discover).
- `--max-cost` default comes from a config value (future); for now, require explicit `--max-cost` to run without `--dry-run`, to force the user to acknowledge the budget. This matches ADR-002's emphasis that cost ceiling matters more for reconcile than for reads.

### Exit codes

- `0` — run completed (including partial completion with a recorded cursor)
- `1` — generator failure or unrecoverable error
- `2` — usage error

## Acceptance criteria

- [ ] Integration test: assess phase happy path against a temp `.engram` with a recording-mode generator
- [ ] Integration test: discover phase happy path
- [ ] Integration test: `--dry-run` does not persist, does not advance cursor
- [ ] Integration test: `--max-cost 0` exhausts immediately and records a partial run
- [ ] Human-readable streamed progress output
- [ ] Final summary prints `reconciliation_runs.id` so users can audit the run later

## Out of scope

- Per-token cost accounting (future)
- MCP tool surface

## Blocked by

- #69 reconcile() assess phase
- #70 reconcile() discover phase

## References

- [`docs/internal/specs/projections.md`](../blob/main/docs/internal/specs/projections.md) § Operations 2

## Documentation

- [ ] Command help text (`engram reconcile --help`) covers phases, scope filter syntax, cost budget, and dry-run semantics
- [ ] Help text explains that `--max-cost` is required without `--dry-run` and why (budget acknowledgement)
- [ ] CLAUDE.md: add `reconcile` to the Core CLI command list in File Organization
- [ ] CLAUDE.md: under Architecture Patterns, cross-reference the two-phase reconcile model documented by #69

rnwolfe:
## Summary

- Implements `engram reconcile [--phase assess|discover|both] [--scope <filter>] [--max-cost <n>] [--dry-run]` CLI command
- Validates all flags at parse-time (exit 2 for usage errors), requires `--max-cost` unless `--dry-run` to force budget acknowledgment
- Streams per-phase progress and prints a structured summary including `reconciliation_runs.id` (ULID) for auditing
- On `--max-cost` exhaustion, prints a "Budget exhausted" notice with the cursor run ID for re-run continuation
- CLAUDE.md updated to list `reconcile` in the CLI command table

## Acceptance Criteria

- [x] Command registered in CLI root (`registerReconcile` called in `cli.ts`)
- [x] Integration test: assess phase happy path with recording-mode generator (via core `reconcile()`)
- [x] Integration test: discover phase happy path (no-op stub; confirms exit 0 and `Reconciliation complete`)
- [x] Integration test: `--dry-run` does not persist, does not advance cursor
- [x] Integration test: `--max-cost 0` exhausts immediately, records partial run
- [x] Human-readable streamed progress output (phase headers, per-metric counts)
- [x] Final summary prints `reconciliation_runs.id`
- [x] CLAUDE.md: `reconcile` added to CLI command list

## Notes

This PR stacks on `feat/issue-70-discover-phase` (#87). The discover phase is a no-op stub in the core engine at this revision; the CLI correctly routes `--phase discover` to `reconcile()` which handles it transparently.

Closes #73

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -55,7 +55,7 @@ engram/
 │   │   └── test/
 │   ├── engram-cli/           # CLI application
 │   │   └── src/
-│   │       └── commands/     # init, add, search, show, decay, ingest, serve, export
+│   │       └── commands/     # init, add, search, show, decay, ingest, serve, export, reconcile
 │   ├── engram-mcp/           # MCP server (stdio transport)
 │   │   └── src/
 │   │       ├── tools/        # MCP tool implementations
--- a/packages/engram-cli/src/cli.ts
+++ b/packages/engram-cli/src/cli.ts
@@ -10,6 +10,7 @@ import { registerIngest } from "./commands/ingest.js";
 import { registerInit } from "./commands/init.js";
 import { registerMaintenance } from "./commands/maintenance.js";
 import { registerOwnership } from "./commands/ownership.js";
+import { registerReconcile } from "./commands/reconcile.js";
 import { registerSearch } from "./commands/search.js";
 import { registerShow } from "./commands/show.js";
 import { registerStats } from "./commands/stats.js";
@@ -35,6 +36,7 @@ registerIngest(program);
 registerExport(program);
 registerVerify(program);
 registerMaintenance(program);
+registerReconcile(program);
 registerVisualize(program);
 
 program.parse();
--- a/packages/engram-cli/src/commands/export.ts
+++ b/packages/engram-cli/src/commands/export.ts
@@ -2,13 +2,20 @@
  * export.ts — `engram export` command.
  *
  * Exports the graph as deterministic JSONL (one object per line, sorted by ID)
- * or as markdown summaries.
+ * or as markdown summaries. Includes `engram export wiki` subcommand.
  */
 
+import * as fs from "node:fs";
 import * as path from "node:path";
 import type { Command } from "commander";
-import type { EngramGraph } from "engram-core";
-import { closeGraph, findEdges, findEntities, openGraph } from "engram-core";
+import type { EngramGraph, Projection } from "engram-core";
+import {
+  closeGraph,
+  findEdges,
+  findEntities,
+  listActiveProjections,
+  openGraph,
+} from "engram-core";
 
 interface ExportOpts {
   format: string;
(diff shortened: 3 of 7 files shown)


rnwolfe (commented):
## Review: `engram reconcile` CLI command (issue #73)

Overall the implementation is solid and the acceptance criteria are met. CI passes. Below are findings by severity.

---

### BLOCKING

**None.** All acceptance criteria are satisfied and no correctness bugs were found.

---

### MAJOR (non-blocking but worth fixing before merge)

#### 1. Phase headers are printed before `reconcile()` returns, losing per-phase progress

In `reconcile.ts` lines ~160–166:

```ts
if (phases.includes("assess")) {
  printPhaseHeader("assess");
}
if (phases.includes("discover")) {
  printPhaseHeader("discover");
}

result = await reconcile(graph, generator, { ... });
```

Both phase headers are printed up-front before the reconcile call completes. The user sees `assess phase starting...` and `discover phase starting...` at the same instant, with no output between them. The discover header actually appears even though the discover phase is a no-op stub. This means progress output is not "streamed per-phase" as claimed — it's a batch dump at start. Consequence: the output is misleading for long-running assess phases (no indication of in-progress work).

Suggested fix: split into two `reconcile()` calls with `phases: ['assess']` / `phases: ['discover']` and print headers + per-phase counts between them, or accept a callback/async-generator from `reconcile()`. Alternatively, adjust the PR description claim to "phase labels are printed before execution" rather than "per-phase streamed progress."

#### 2. `--max-cost 0` test exercises core directly, not the CLI

The test `"budget=0 with stale projection records partial run via core reconcile()"` imports and calls `coreReconcile` directly. This verifies core behavior but **does not test the CLI's budget path** (i.e., that `--max-cost 0` passed to the CLI correctly flows to `maxCost: 0` and results in `status: 'partial'` being printed). The CLI path for this scenario is only tested on an empty graph (which never triggers the budget), so the "Budget exhausted" output line is never exercised in tests.

---

### MINOR

#### 3. `closeGraph` called before printing summary on the success path

The flow is: `reconcile()` → `closeGraph(graph)` → print per-phase progress → print summary. The graph is closed before the summary is printed, which is fine for correctness, but differs from the error path (which closes before `process.exit(1)`). No bug, but inconsistent placement makes the code harder to audit at a glance.

#### 4. Scope string passed raw to `reconcile()` — core parse diverges from CLI validation

`parseScopeToOpts()` in the core splits on the *first* `:` character (`scope.split(":")`) which gives correct results for `kind:entity_summary` but silently drops the value for a scope like `anchor:my:weird:id` (split gives `['anchor', 'my', 'weird', 'id']`, destructured as `[key, value]` ignoring extras). Meanwhile `validateScope()` in the CLI only checks `scope.startsWith("kind:")` and `scope.length > 5`. These two validators are not co-located, so a future change to one may not update the other. Low risk for now since anchor values are unlikely to contain `:`, but worth a comment.

#### 5. Phase headers printed for both phases even in single-phase mode

When `--phase assess` is passed, the code still evaluates `phases.includes("discover")` which is false, so the discover header is not shown. That's correct. But the comment in the code `// We use NullGenerator here` reads as though the generator choice is a permanent design decision rather than an interim one. The important missing piece is that the CLI never selects a real generator from environment variables, so any real-world (non-empty-graph) run will always exit 1. The comment explains this, but there is no warning printed to the user when `ENGRAM_AI_PROVIDER` is unset and stale projections exist. The user just sees `Reconcile failed: NullGenerator: no AI provider configured`.

---

### NITPICKS

- `runCommand()` test helper captures `console.error` into the same `logs` array as `console.log`, which means exit-2 error messages are indistinguishable from normal output. Not a test correctness issue since tests check `output.toContain(...)` for specific strings, but noting for future test clarity.
- `--max-cost` accepts `"0"` as a string (from Commander), parsed with `Number()`, and `Number.isFinite(0)` is `true` so `0` is accepted correctly. Also correctly accepted: `"0.5"`, `"1e3"`. Fractional and scientific notation are probably fine but may surprise users. Low priority.
- `printSummary` re-prints the run ID that was already printed in the "Budget exhausted" partial block. This means partial runs show the run ID twice. Intentional? Harmless but redundant.

---

### Summary table

| # | Severity | Blocking? | Description |
|---|----------|-----------|-------------|
| 1 | Major | No | Phase headers are pre-printed, not streamed; discover header fires even as stub |
| 2 | Major | No | `--max-cost 0` CLI budget-exhausted path not covered by CLI-level test |
| 3 | Minor | No | `closeGraph` placement differs between success and error paths |
| 4 | Minor | No | Scope validation in CLI and core are not co-located; `:` in values silently truncates |
| 5 | Minor | No | No user-facing warning when NullGenerator is the generator and stale projections exist |

The implementation is correct, the exit codes are right (2 for usage, 1 for generator failure, 0 for success and partial-with-record), the run ID is printed in the summary, the scope parsing and passing to `reconcile()` is correct, dry-run does not persist, and the test suite covers all the claimed acceptance criteria. The two major findings are quality/coverage issues worth addressing but not blockers.

rnwolfe (commented):
## Review: `engram reconcile` CLI command (issue #73)

Overall the implementation is solid and the acceptance criteria are met. CI passes. Below are findings by severity.

---

### BLOCKING

**None.** All acceptance criteria are satisfied and no correctness bugs were found.

---

### MAJOR (non-blocking but worth fixing before merge)

#### 1. Phase headers printed before reconcile() runs, not streamed

In `reconcile.ts` both `printPhaseHeader()` calls fire *before* `await reconcile(...)` is called. The user sees both headers at the same instant at the start, with no output during execution. The discover header appears even though the phase is a no-op stub. The PR description claims "per-phase streamed progress" but what is actually implemented is batch output at the beginning.

Suggested path: split into two sequential `reconcile()` calls with `phases: ['assess']` then `phases: ['discover']` and print headers + counts between them; or adjust the PR description to match the actual behavior.

#### 2. `--max-cost 0` budget-exhausted CLI path is not tested at the CLI level

The test `"budget=0 with stale projection records partial run via core reconcile()"` imports and calls `coreReconcile` directly. It verifies core behavior but does **not** exercise the CLI path (flag parsing → `maxCost: 0` → `result.status === 'partial'` → "Budget exhausted" output). The CLI-level test for `--max-cost 0` only runs on an empty graph, where no stale projections exist and the budget is never hit, so `status` is always `'completed'`. The "Budget exhausted" line in `reconcile.ts` has no test coverage.

---

### MINOR

#### 3. `closeGraph` occurs before the summary is printed on the success path

The sequence is: `result = await reconcile(...)` → `closeGraph(graph)` → print progress counts → print summary. The graph is closed mid-output. No bug, but the error path closes *then* exits immediately, while the success path closes *then* continues printing. Suggest moving `closeGraph` to after the summary for consistency.

#### 4. Scope validation in CLI and core are not co-located

`validateScope()` in the CLI checks `scope.startsWith("kind:")` and length. `parseScopeToOpts()` in the core splits on the first `:`. A value like `anchor:my:ns:id` passes CLI validation but loses everything after the second `:` in the core parse (destructured as `[key, value]` discarding later segments). Low risk today, but the validators should reference each other or share a module.

#### 5. No user-facing guidance when NullGenerator is active and stale projections exist

When `ENGRAM_AI_PROVIDER` is unset and stale projections exist, the CLI exits 1 with `Reconcile failed: NullGenerator: no AI provider configured`. A hint (e.g. "Set ENGRAM_AI_PROVIDER to use a real generator") at the start of the action — before opening the graph — would reduce confusion. The code comment explains the design intent but the user never sees it.

---

### NITPICK

- Partial runs print the run ID twice: once in the "Budget exhausted" block and again in `printSummary`. Harmless but redundant.
- `runCommand()` captures `console.error` and `console.log` into the same `logs` array; distinguishing error output from normal output in test assertions is not possible. No test fails because of this, but future tests that need to assert *which* stream a message appeared on cannot.

---

### Summary

| # | Severity | Blocking? | Description |
|---|----------|-----------|-------------|
| 1 | Major | No | Phase headers pre-printed, not streamed; discover header fires unconditionally |
| 2 | Major | No | `--max-cost 0` budget-exhausted CLI path uncovered by CLI-level test |
| 3 | Minor | No | `closeGraph` placement inconsistency on success vs error path |
| 4 | Minor | No | Scope validation split across CLI and core; `:` in anchor values silently truncated |
| 5 | Minor | No | No user hint when NullGenerator is active and stale projections are present |

Exit codes are correct (2 for usage, 1 for generator/runtime failure, 0 for completed and partial). The `--max-cost` required-without-`--dry-run` guard works correctly. `--scope` is parsed and forwarded correctly to `reconcile()`. The run ID is printed in the summary. Dry-run does not persist. All stated acceptance criteria are met.

rnwolfe (commented):
Fixed: phase headers now streamed correctly, added CLI-level budget exhaustion test, NullGenerator error message improved, scope split fixed for anchor:type:id format.

Issue #96, opened 2026-04-14
epic: source code ingestion — tree-sitter-powered AST ingester for TS/JS


rnwolfe:
## Summary

Add a source-code ingestion pipeline to engram that walks the working tree, parses TypeScript and JavaScript with tree-sitter, and creates `file` / `module` / `symbol` entities in the graph. Completes Principle 6 from [VISION.md](../blob/main/docs/internal/VISION.md) — "Developer-native. First-class ingestors understand git and code." — and unblocks the baseline reconcile phase (separate epic).

## Why

Engram's graph today is derived entirely from git history. Entities exist because commits and PRs mentioned them, never because files were parsed. This leaves a blind spot: the engine has no ground-truth view of what the code actually contains. That's:

1. Bad for retrieval — "what does `validateToken` do?" can only surface commits that touched it, not the function itself.
2. Bad for projections — the projection layer has to infer code structure from diffs rather than read it directly.
3. A hard blocker for the baseline wiki — hierarchical synthesis (`symbol → file → module → area → system`) requires the file tree to be in the graph as entities with evidence chains.

Source ingestion fixes all three while preserving every evidence-first and temporal invariant that engram already guarantees.

## Scope (v1)

- **Languages**: TypeScript + JavaScript only (`.ts`, `.tsx`, `.js`, `.jsx`, `.mjs`, `.cjs`)
- **Symbols**: top-level only (functions, classes, interfaces, types, enums, exported consts, default exports)
- **Edges**: `contains`, `imports` (relative only), `defined_in`
- **Idempotent**: content-hash keyed, unchanged files skip parsing entirely
- **Respects `.gitignore`** plus a hardcoded denylist for build artifacts
- **Deterministic**: no LLM calls; pure parser output

## Documents

- **Spec**: [`docs/internal/specs/source-ingestion.md`](../blob/main/docs/internal/specs/source-ingestion.md) — contract, CLI/API surface, data model, acceptance criteria
- **Design**: [`docs/internal/specs/source-ingestion-design.md`](../blob/main/docs/internal/specs/source-ingestion-design.md) — walker/parser/extractor internals, idempotency mechanics, alternatives considered, implementation sequence

## Child issues

- [x] #97 — scaffold + file walker
- [x] #98 — tree-sitter parser + grammar vendoring
- [x] #99 — tree-sitter query + TS extractor
- [x] #100 — episode/entity/edge writes + idempotency
- [x] #101 — sweep phase for deleted files
[...]

PR #104, opened 2026-04-14
feat(ingest): source ingestion — scaffold + file walker

### LINKED ISSUE #97: feat(ingest): source ingestion — scaffold + file walker
## Context

Implements chunk 1 of the source-ingestion epic (#96). Spec: [`docs/internal/specs/source-ingestion.md`](../blob/main/docs/internal/specs/source-ingestion.md). Design: [`docs/internal/specs/source-ingestion-design.md`](../blob/main/docs/internal/specs/source-ingestion-design.md). Read both before starting — they define the contract this chunk must meet.

This chunk adds package scaffolding, dependencies, and the pure file walker. No tree-sitter, no database, no CLI. The walker is a generator yielding `FileEntry` records with content hashes; subsequent chunks consume its output.

Part of #96.

## Scope

- Add dependencies to `packages/engram-core/package.json`: `web-tree-sitter`, `ignore`, `@noble/hashes`
- Create the directory tree: `packages/engram-core/src/ingest/source/{extractors,queries,grammars}/`
- Implement `packages/engram-core/src/ingest/source/walker.ts`
- Create a test fixture at `packages/engram-core/test/fixtures/source-sample/`
- Write walker tests at `packages/engram-core/test/ingest/source/walker.test.ts`

### Walker API

```typescript
interface FileEntry {
  relPath: string;      // posix, relative to root
  absPath: string;
  contentHash: string;  // blake3 hex
  size: number;
  body: string;
}

interface WalkOptions {
  root: string;
  exclude?: string[];
  respectGitignore?: boolean;  // default true
  maxFileBytes?: number;       // default 1_048_576
}

export async function* walk(opts: WalkOptions): AsyncIterable<FileEntry>;
```

### Hardcoded denylist (always applied, cannot be disabled)

`node_modules/`, `dist/`, `build/`, `.next/`, `out/`, `target/`, `.git/`, `coverage/`, `*.min.js`, `*.min.css`, `*.map`, `*.lockb`, `package-lock.json`, `bun.lock`, `bun.lockb`, `yarn.lock`, `pnpm-lock.yaml`, `.DS_Store`, `Thumbs.db`.

The denylist wins over `.gitignore` — even if a `.gitignore` un-ignores `node_modules/`, the walker still skips it.

### .gitignore support

Use the [`ignore`](https://www.npmjs.com/package/ignore) npm package. Discover all `.gitignore` files under the root, build one `ignore` instance per directory, and chain them from root downward. User-supplied `--exclude` patterns layer on top of the root instance.

### Binary detection

Read the first 4KB, check for any `0x00` byte → skip silently. Applied before the size check.

### Size limit

Default 1MB. Files above the limit are skipped with `console.warn`.

### Content hash

`blake3` from `@noble/hashes/blake3`, hex encoded. Pure JS, no native deps.

### Symlinks

Follow once, but track visited inodes to prevent infinite loops.

## Acceptance criteria

- [ ] `walk()` yields only allowed files from the fixture
- [ ] Denylist patterns are never yielded, even when `.gitignore` un-ignores them
- [ ] Nested `.gitignore` files are honored (fixture includes one)
- [ ] Binary files in the fixture are skipped silently
- [ ] `respectGitignore: false` bypasses `.gitignore` but keeps the denylist
- [ ] User `--exclude` patterns layer correctly on top of the denylist and `.gitignore`
- [ ] Over-size files are skipped with a warn log
- [ ] Symlink loops do not hang the walker
- [ ] `bun test` passes, `bun run lint` passes, `bun run build` passes

## Test fixture structure

```
packages/engram-core/test/fixtures/source-sample/
├── .gitignore                    # ignores generated.ts
├── src/
│   ├── a.ts                      # ingestable
│   ├── b.ts                      # ingestable (imports a)
│   ├── generated.ts              # gitignored
│   └── nested/
│       ├── .gitignore            # ignores hidden.ts
│       ├── c.ts                  # ingestable
│       └── hidden.ts             # nested-gitignored
├── node_modules/
│   └── evil.ts                   # denylist
├── dist/
│   └── bundle.js                 # denylist
└── assets/
    └── logo.bin                  # binary (contains null byte)
```

## Out of scope

Tree-sitter parser (#2 of the epic), extractor (#3), database writes (#4), CLI (#6). Do not touch those layers from this chunk.

## Dependencies

None. This is the first chunk of the epic and has no prerequisites.

rnwolfe:
## Summary

Implements chunk 1 of the source ingestion epic (#96). Adds the directory scaffold and the pure file walker — no tree-sitter, no DB, no CLI. This is the foundation that chunks 2–6 build on.

- Add `web-tree-sitter`, `ignore`, `@noble/hashes` to `engram-core` deps
- Scaffold `src/ingest/source/{extractors,queries,grammars}/` directories
- Implement `walker.ts`: async generator yielding `FileEntry` records with blake3 content hashes
- Create `test/fixtures/source-sample/` with gitignored, binary, denylist, and nested-gitignore files
- 20 unit tests covering all acceptance criteria

## Acceptance Criteria

- [x] `walk()` yields only allowed files from the fixture (3 files: `src/a.ts`, `src/b.ts`, `src/nested/c.ts`)
- [x] Denylist patterns (`node_modules/`, `dist/`) are never yielded, even when `.gitignore` un-ignores them
- [x] Nested `.gitignore` files are honored (`src/nested/hidden.ts` excluded)
- [x] Binary files (`assets/logo.bin`) are skipped silently
- [x] `respectGitignore: false` bypasses `.gitignore` but keeps the denylist
- [x] User `exclude` patterns layer correctly on top of denylist and `.gitignore`
- [x] Over-size files are skipped with a warn log
- [x] Symlink loops do not hang the walker (inode tracking)
- [x] `bun test` passes, `bun run lint` passes, `bun run build` passes

Closes #97
Part of #96


packages/engram-core/src/ingest/source/walker.ts
```diff
@@ -0,0 +1,250 @@
...
+        // Apply gitignore chain
+        if (respectGitignore && isGitignored(relPosix)) {
+          continue;
+        }
+
+        // Read the file
+        let fileBuffer: Buffer;
+        try {
+          fileBuffer = fs.readFileSync(absEntryPath);
+        } catch {
```
rnwolfe, line 218: **Size check happens after reading the full file into memory.**

The file is read into `fileBuffer` unconditionally, and the size limit is only checked afterwards (line ~226). For a repo with many large files, this buffers every oversized file before discarding it — wasting memory and I/O.

The fix is to check `stat.size > maxFileBytes` *before* calling `readFileSync`:

```ts
// Size limit check — do this BEFORE reading to avoid buffering oversized files
if (stat.size > maxFileBytes) {
  console.warn(`[engram walker] skipping ${relPosix}: file size ${stat.size} exceeds limit ${maxFileBytes}`);
  continue;
}

let fileBuffer: Buffer;
try {
  fileBuffer = fs.readFileSync(absEntryPath);
} catch {
  continue;
}
```

The binary detection (first 4KB null-byte scan) is cheap and can stay after the read, but the size guard should gate the read entirely.

packages/engram-core/src/ingest/source/walker.ts
```diff
@@ -0,0 +1,250 @@
...
+        // Binary detection: check first 4KB for null bytes
+        if (isBinary(fileBuffer)) {
+          continue;
+        }
+
+        // Size limit check
+        const fileSize = stat.size;
+        if (fileSize > maxFileBytes) {
+          console.warn(
+            `[engram walker] skipping ${relPosix}: file size ${fileSize} exceeds limit ${maxFileBytes}`,
```
rnwolfe: **`console.warn` bypasses the project's debug logging convention.**

Other ingestion code (e.g. `git.ts`) uses `ENGRAM_DEBUG`-gated logging via the project's debug utilities rather than calling `console.warn` directly. This warn fires unconditionally in production and in test output, which is why the test has to spy on `console.warn` to silence it.

Consider either:
1. Using a structured logger consistent with the rest of the ingest layer, or
2. Accepting a `logger` option in `WalkOptions` so callers can route the message appropriately.

At minimum the test spy pattern (`spyOn(console, 'warn').mockImplementation(() => {})`) suggests this is already recognised as an awkward interface.

rnwolfe (commented):
Code review of the source walker implementation. Found 3 findings: 1 medium-severity logic/performance issue, 1 minor convention issue, and 1 unnecessary dependency. No blocking issues — the core logic is correct and the test coverage is solid.

packages/engram-core/package.json
```diff
@@ -18,7 +18,10 @@
   "dependencies": {
     "@anthropic-ai/sdk": "^0.88.0",
     "@google/genai": "^1.48.0",
+    "@noble/hashes": "^1.7.2",
+    "ignore": "^6.0.2",
     "openai": "^6.34.0",
-    "ulid": "^2.3.0"
+    "ulid": "^2.3.0",
+    "web-tree-sitter": "^0.25.6"
```
rnwolfe, line 25: **`web-tree-sitter` is declared as a runtime dependency but is not used in this chunk.**

The PR description says tree-sitter is being scaffolded for later chunks, but adding it to `dependencies` now means it is installed for all consumers of `engram-core` today. It also pulls in a native WASM binary that adds non-trivial install weight.

Consider moving it to `devDependencies` (or a comment placeholder) until the chunk that actually imports it lands. The `ignore` and `@noble/hashes` additions are fine — they are used here.

packages/engram-core/src/ingest/source/walker.ts
```diff
@@ -0,0 +1,250 @@
...
+  /**
+   * Check if a path (relative to root, posix) is ignored by any gitignore
+   * in the chain from root to the file's parent directory.
+   */
+  function isGitignored(relPosix: string): boolean {
+    if (!respectGitignore) return false;
+
+    // Walk from root down to the file's parent, checking each directory's gitignore
+    const parts = relPosix.split("/");
+
```
rnwolfe, line 133: **`isGitignored` re-checks higher-level directories on every file call — O(depth) `getGitignore` lookups per file.**

This is not a correctness issue (the cache makes the repeated `getGitignore` calls cheap), but the comment is worth noting: the `gitignoreCache` keyed by `absPath` ensures each `.gitignore` is parsed only once, so the O(depth) traversal is already amortised. The implementation is correct. A drive-by nit: the variable name `dirRelParts` at line 136 is only used to compute `dirAbsPath` — the intermediate array allocation could be avoided with `path.join(absRoot, ...parts.slice(0, depth))` directly, but this is purely cosmetic.

rnwolfe:
**Review responses:**

- **Size check ordering (fixed):** Moved size check above `readFileSync` — no point buffering a file we'll immediately discard. Pushed as a follow-up commit.

- **`console.warn` logging:** The issue spec explicitly requires `console.warn` here. Keeping per spec — a structured logger pass can come later if warranted.

- **`web-tree-sitter` in deps now:** Issue #97 scope explicitly includes adding all three deps (`web-tree-sitter`, `ignore`, `@noble/hashes`). Chunks 2–3 (parser + extractor) will actually import it. Keeping per spec.

- **`isGitignored` O(depth) nit:** Cache amortises repeated lookups. Inconsequential at current scale — deferring until benchmarks show otherwise.

PR #106, opened 2026-04-14
feat(ingest): source ingestion — tree-sitter query + TS extractor

### LINKED ISSUE #99: feat(ingest): source ingestion — tree-sitter query + TS extractor
## Context

Implements chunk 3 of the source-ingestion epic (#96). Spec: [`docs/internal/specs/source-ingestion.md`](../blob/main/docs/internal/specs/source-ingestion.md). Design: [`docs/internal/specs/source-ingestion-design.md`](../blob/main/docs/internal/specs/source-ingestion-design.md) — see the "Extractor" section.

Writes the tree-sitter query that captures top-level TypeScript/JavaScript symbols and imports, and the extractor that turns query captures into a structured `ExtractedFile` record. Still no DB writes — the extractor is a pure function of the parse tree.

Part of #96.

## Scope

- Write `packages/engram-core/src/ingest/source/queries/typescript.scm`
- Extend `parser.ts` with a `runQuery(tree, lang)` method and query caching
- Implement `packages/engram-core/src/ingest/source/extractors/typescript.ts`
- Implement the relative-import resolver
- Write extractor tests at `packages/engram-core/test/ingest/source/extractor.test.ts`

### Tree-sitter query

`queries/typescript.scm` — captures top-level only, anchored to `(program …)` so nested symbols never match:

```scheme
; --- Top-level declarations ---
(program (function_declaration name: (identifier) @symbol.function))
(program (class_declaration name: (type_identifier) @symbol.class))
(program (interface_declaration name: (type_identifier) @symbol.interface))
(program (type_alias_declaration name: (type_identifier) @symbol.type))
(program (enum_declaration name: (identifier) @symbol.enum))
(program (lexical_declaration (variable_declarator name: (identifier) @symbol.const)))

; --- Exported variants ---
(program (export_statement declaration: (function_declaration name: (identifier) @symbol.function.exported)))
(program (export_statement declaration: (class_declaration name: (type_identifier) @symbol.class.exported)))
(program (export_statement declaration: (interface_declaration name: (type_identifier) @symbol.interface.exported)))
(program (export_statement declaration: (type_alias_declaration name: (type_identifier) @symbol.type.exported)))
(program (export_statement declaration: (enum_declaration name: (identifier) @symbol.enum.exported)))
(program (export_statement declaration: (lexical_declaration (variable_declarator name: (identifier) @symbol.const.exported))))

; --- Default exports ---
(program (export_statement "default" value: (identifier) @symbol.default.ref))

; --- Imports ---
(import_statement source: (string) @import.source)
```

Validate the query against real TypeScript files from the engram repo itself. If tree-sitter-typescript at the pinned grammar version uses different node names, adjust to match. The above is correct at tree-sitter-typescript 0.23.x but must be re-verified against whatever #2 of the epic vendors.

### Extractor API

```typescript
export interface ExtractedFile {
  symbols: ExtractedSymbol[];
  rawImports: string[];
}

export interface ExtractedSymbol {
  name: string;
  kind: 'function' | 'class' | 'interface' | 'type' | 'enum' | 'const' | 'default';
  exported: boolean;
  startByte: number;
  endByte: number;
}

export function extractTypeScript(tree: Tree, captures: QueryCapture[]): ExtractedFile;
```

### Collision handling

If two captures produce the same `name`, keep the first and emit `console.warn`. Rare in valid TypeScript, occasional in loose JavaScript.

### Import resolver

Lives in the extractors module, called later by chunk 4 (DB integration) once the full file set is known:

```typescript
export function resolveImport(
  specifier: string,
  fromRelPath: string,
  knownFiles: Set<string>,
  root: string
): string | null;
```

Rules:

- Relative specifiers (`./` or `../`): join against `fromRelPath`'s directory, try extensions in order `.ts` → `.tsx` → `.js` → `.jsx` → `.mjs` → `.cjs`, then `/index.ts` → `/index.tsx` → `/index.js`. Return the first match in `knownFiles`, otherwise `null`.
- Package specifiers (`react`, `fs`): return `null` — not resolved in v1.
- Alias specifiers (`@engram/core`): return `null` — not resolved in v1.

## Acceptance criteria

- [ ] `extractTypeScript()` returns all top-level symbols with correct names, kinds, and export flags
- [ ] Nested symbols (methods, inner functions, arrow functions assigned to vars) are NOT extracted
- [ ] Imports appear in `rawImports` as raw specifier strings
- [ ] Same-name top-level collisions emit a warning and keep the first capture
- [ ] `resolveImport('./foo', 'src/a.ts', knownFiles, root)` resolves via extension fallback
- [ ] `resolveImport('./foo', ...)` falls through to `foo/index.ts` when `foo.ts` is not present
- [ ] `resolveImport('react', ...)` returns `null`
- [ ] `resolveImport('@engram/core', ...)` returns `null`
- [ ] Extractor handles TSX files correctly (JSX expressions do not break captures)

## Tests

Extend `packages/engram-core/test/fixtures/source-sample/src/` with files covering every symbol kind, including nested exclusions and TSX. Unit test `extractTypeScript` against each. Separate unit tests for `resolveImport` covering relative paths, extension fallback, `index.ts` fallback, package imports, and missing targets.

## Out of scope

Database integration (#4), sweep phase (#5), CLI (#6).

## Dependencies

Blocked on chunk 2 of the epic (parser + grammars). Chunk 1 is a soft prerequisite for the fixture scaffolding.

rnwolfe:
## Summary

Implements chunk 3 of the source ingestion epic (#96). Adds the tree-sitter S-expression query for TypeScript/TSX top-level symbols, extends the parser with `runQuery()`, and implements the pure `extractTypeScript()` and `resolveImport()` functions.

- `queries/typescript.scm` — 18 patterns covering all 6 symbol kinds (unexported + exported) plus imports; anchored to `(program …)` to exclude nested symbols
- `parser.ts` extended: `runQuery(tree, lang)` with lazy query caching per language; queries freed in `dispose()`
- `extractors/typescript.ts`: `extractTypeScript(captures)` → `ExtractedFile`; `resolveImport(specifier, from, knownFiles, root)` with extension + index fallback
- 29 new tests (66 total across walker + parser + extractor suites)

Node names verified against actual parse tree output (tree-sitter-typescript v0.23.2).

## Acceptance Criteria

- [x] All 6 symbol kinds extracted with correct name, kind, and export flag
- [x] Nested symbols (class methods, inner functions) NOT extracted (anchored to `(program …)`)
- [x] Imports appear in `rawImports` as raw specifier strings (quotes stripped)
- [x] Duplicate symbol names → warn + keep first
- [x] `resolveImport` resolves relative imports via extension fallback (`.ts → .tsx → .js → …`)
- [x] `resolveImport` falls back to `foo/index.ts` when `foo.ts` not present
- [x] `resolveImport('react', …)` → `null`
- [x] `resolveImport('@scope/pkg', …)` → `null`
- [x] TSX JSX expressions don't break extraction

Closes #99
Part of #96

### DIFF
--- a/packages/engram-core/src/ingest/source/extractors/typescript.ts
+++ b/packages/engram-core/src/ingest/source/extractors/typescript.ts
@@ -125,6 +125,12 @@ export function resolveImport(
   const EXTENSIONS = [".ts", ".tsx", ".js", ".jsx", ".mjs", ".cjs"];
   const INDEX_FILES = ["index.ts", "index.tsx", "index.js"];
 
+  // 0. Check if specifier already includes an explicit known extension (e.g. './utils.ts')
+  const asIs = path.relative(root, resolvedAbs).split(path.sep).join("/");
+  if (knownFiles.has(asIs)) {
+    return asIs;
+  }
+
   // 1. Try adding an extension directly
   for (const ext of EXTENSIONS) {
     const candidate = path.relative(root, resolvedAbs + ext);
--- a/packages/engram-core/src/ingest/source/extractors/typescript.ts
+++ b/packages/engram-core/src/ingest/source/extractors/typescript.ts
@@ -1,0 +1,148 @@
+import path from "node:path";
+import type { QueryCapture } from "../parser";
+
+/** A top-level symbol found in a TypeScript/TSX file. */
+export interface ExtractedSymbol {
+  name: string;
+  kind:
+    | "function"
+    | "class"
+    | "interface"
+    | "type"
+    | "enum"
+    | "const"
+    | "default";
+  exported: boolean;
+  startByte: number;
+  endByte: number;
+}
+
+/** The result of extracting symbols and imports from a single file. */
+export interface ExtractedFile {
+  symbols: ExtractedSymbol[];
+  /** Raw import specifier strings, e.g. `'./foo'`, `'react'`. */
+  rawImports: string[];
+}
+
+/** Map from capture name prefix to ExtractedSymbol kind. */
+const KIND_MAP: Record<string, ExtractedSymbol["kind"]> = {
+  function: "function",
+  class: "class",
+  interface: "interface",
+  type: "type",
+  enum: "enum",
+  const: "const",
+  default: "default",
+};
+
+/**
+ * Extract top-level symbols and raw import specifiers from tree-sitter query
+ * captures produced by `SourceParser.runQuery()`.
+ *
+ * Pure function — no IO, no DB writes.
+ */
+export function extractTypeScript(captures: QueryCapture[]): ExtractedFile {
+  const symbols: ExtractedSymbol[] = [];
+  const rawImports: string[] = [];
+  const seenNames = new Map<string, number>(); // name -> index in symbols array
+
+  for (const capture of captures) {
+    const { name: captureName, node } = capture;
+
+    if (captureName === "import.source") {
+      // node.text includes surrounding quotes — strip them
+      const raw = node.text.slice(1, -1);
+      rawImports.push(raw);
+      continue;
+    }
+
+    if (captureName.startsWith("symbol.")) {
+      // e.g. "symbol.function.exported" → parts = ["function", "exported"]
+      const rest = captureName.slice("symbol.".length); // "function.exported"
+      const parts = rest.split(".");
+      const kindKey = parts[0]; // "function", "class", ..., "default"
+      const exported = parts[1] === "exported" || kindKey === "default";
+
+      const kind = KIND_MAP[kindKey];
+      if (!kind) {
+        // Unknown capture kind — skip silently
+        continue;
+      }
+
+      const symbolName = node.text;
+
+      if (seenNames.has(symbolName)) {
+        console.warn(
+          `[engram extractor] duplicate symbol name '${symbolName}' — keeping first occurrence`,
+        );
+        continue;
+      }
+
+      const symbol: ExtractedSymbol = {
+        name: symbolName,
+        kind,
+        exported,
+        startByte: node.startIndex,
+        endByte: node.endIndex,
+      };
+
+      seenNames.set(symbolName, symbols.length);
+      symbols.push(symbol);
+    }
+  }
+
+  return { symbols, rawImports };
+}
+
+/**
+ * Resolve a module import specifier to a relative path within the project.
+ *
+ * Returns `null` for:
+ * - Non-relative specifiers (no `./` or `../` prefix): npm packages, scoped packages
+ * - Relative paths that don't resolve to any known file
+ *
+ * @param specifier   The raw import string, e.g. `'./utils'`, `'react'`
+ * @param fromRelPath Repo-relative path of the file containing the import (POSIX)
+ * @param knownFiles  Set of all known repo-relative file paths (POSIX)
+ * @param root        Absolute path to the repo root
+ */
+export function resolveImport(
+  specifier: string,
+  fromRelPath: string,
+  knownFiles: Set<string>,
+  root: string,
+): string | null {
+  // Non-relative imports are external packages — not resolvable within the repo
+  if (!specifier.startsWith("./") && !specifier.startsWith("../")) {
+    return null;
+  }
+
+  const fromDir = path.dirname(fromRelPath);
+  // Resolve the specifier relative to the importing file's directory
+  const resolvedAbs = path.resolve(root, fromDir, specifier);
+
+  // Extensions to try in order (TypeScript-first, then JS variants)
+  const EXTENSIONS = [".ts", ".tsx", ".js", ".jsx", ".mjs", ".cjs"];
+  const INDEX_FILES = ["index.ts", "index.tsx", "index.js"];
+
+  // 1. Try adding an extension directly
+  for (const ext of EXTENSIONS) {
+    const candidate = path.relative(root, resolvedAbs + ext);
+    // Normalize to POSIX separators
+    const posix = candidate.split(path.sep).join("/");
+    if (knownFiles.has(posix)) {
+      return posix;
+    }
+  }
+
+  // 2. Try as a directory with an index file
+  for (const idx of INDEX_FILES) {
+    const candidate = path.relative(root, path.join(resolvedAbs, idx));
+    const posix = candidate.split(path.sep).join("/");
+    if (knownFiles.has(posix)) {
+      return posix;
+    }
+  }
+
+  return null;
+}
--- a/packages/engram-core/src/ingest/source/parser.ts
+++ b/packages/engram-core/src/ingest/source/parser.ts
@@ -1,6 +1,8 @@
+import { readFileSync } from "node:fs";
 import { readFile } from "node:fs/promises";
 import path from "node:path";
-import { Parser, Language as TreeSitterLanguage } from "web-tree-sitter";
+import type { QueryCapture } from "web-tree-sitter";
+import { Parser, Query, Language as TreeSitterLanguage } from "web-tree-sitter";
 
 /** Languages supported by the source parser. */
 export type Language = "typescript" | "tsx";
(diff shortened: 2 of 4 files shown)


rnwolfe (commented):
Overall the structure is clean and the test suite is thorough for the happy paths. Four findings below — one correctness bug that will silently produce wrong `exported` flags on the majority of real-world TS files, one missing coverage in the query, one architectural issue in the cache key, and one minor library-hygiene issue.

packages/engram-core/src/ingest/source/queries/typescript.scm
```diff
@@ -0,0 +1,21 @@
+; --- Top-level declarations (unexported) ---
+(program (function_declaration name: (identifier) @symbol.function))
+(program (class_declaration name: (type_identifier) @symbol.class))
+(program (interface_declaration name: (type_identifier) @symbol.interface))
+(program (type_alias_declaration name: (type_identifier) @symbol.type))
+(program (enum_declaration name: (identifier) @symbol.enum))
+(program (lexical_declaration (variable_declarator name: (identifier) @symbol.const)))
```
rnwolfe, line 7: **Bug (correctness): unexported patterns double-fire on exported declarations**

The bare `(program (function_declaration ...))` pattern matches *any* `function_declaration` that is a direct child of `program` — including the one nested inside an `export_statement`. Tree-sitter evaluates each pattern independently across the whole tree, so `export function foo() {}` produces **two** captures: `@symbol.function` (unexported) and `@symbol.function.exported`.

Because `extractTypeScript` deduplicates by name and keeps the first hit, `foo` ends up with `exported: false`. Every exported symbol will be misclassified.

Fix: add a `(#not-parent ...)` predicate, or (simpler) restructure so the unexported patterns use an anchor that excludes `export_statement` parents:
```scheme
; only match when the direct parent is program, not export_statement
(program . (function_declaration name: (identifier) @symbol.function) .)
```
Or use the `#is-not?` / `#has-ancestor?` predicate if available in your tree-sitter version.

packages/engram-core/src/ingest/source/queries/typescript.scm
```diff
@@ -0,0 +1,21 @@
...
+; --- Exported declarations ---
+(program (export_statement declaration: (function_declaration name: (identifier) @symbol.function.exported)))
+(program (export_statement declaration: (class_declaration name: (type_identifier) @symbol.class.exported)))
+(program (export_statement declaration: (interface_declaration name: (type_identifier) @symbol.interface.exported)))
+(program (export_statement declaration: (type_alias_declaration name: (type_identifier) @symbol.type.exported)))
+(program (export_statement declaration: (enum_declaration name: (identifier) @symbol.enum.exported)))
+(program (export_statement declaration: (lexical_declaration (variable_declarator name: (identifier) @symbol.const.exported))))
+
+; --- Default export (identifier reference) ---
+(program (export_statement value: (identifier) @symbol.default.ref))
```
rnwolfe, line 20: **Missing coverage: `export default` inline declarations not captured**

The current query only captures `export default <identifier>` (a reference to an already-declared binding). The following common patterns are **not** captured at all:

```ts
export default function myFn() {}    // → @symbol.default.ref won't fire (no identifier value node)
export default class MyClass {}      // same
```

These are the dominant patterns for React components and default-exported utilities. Add two patterns:
```scheme
(program (export_statement (function_declaration name: (identifier) @symbol.default.exported)))
(program (export_statement (class_declaration name: (type_identifier) @symbol.default.exported)))
```
The tests don't exercise these cases, so they currently pass — but real-world files will silently drop these symbols.

packages/engram-core/src/ingest/source/parser.ts
```diff
@@ -89,10 +96,45 @@ export class SourceParser {
   }
```
rnwolfe, line 96: **Bug (architecture): `runQuery` always loads `typescript.scm` regardless of `lang`, but caches under `lang`**

```ts
const queryText = readFileSync(
  path.join(import.meta.dir, "queries", "typescript.scm"), // hardcoded
  "utf8",
);
// ...
this.queryCache.set(lang, query); // cached under 'tsx' or 'typescript'
```

When `lang = "tsx"`, a separate `Query` WASM object is compiled and cached under the `tsx` key, but it uses the identical query text. This wastes one WASM compilation per parser instance. More importantly, when a `tsx`-specific query file is needed in the future, this code won't use it — the cache lookup will succeed and return the wrong query.

Fix: use a `QUERY_FILES` map (similar to `GRAMMAR_FILES`) to drive query file selection, and optionally share a compiled query when both languages happen to use the same file:
```ts
const QUERY_FILES: Record<Language, string> = {
  typescript: "typescript.scm",
  tsx: "typescript.scm", // tsx reuses TS query for now
};
```

packages/engram-core/src/ingest/source/extractors/typescript.ts
```diff
@@ -0,0 +1,148 @@
...
+      const exported = parts[1] === "exported" || kindKey === "default";
+
+      const kind = KIND_MAP[kindKey];
+      if (!kind) {
+        // Unknown capture kind — skip silently
+        continue;
+      }
+
+      const symbolName = node.text;
+
```
rnwolfe, line 73: **Convention: `console.warn` in a library function**

`extractTypeScript` is a pure library function inside `engram-core`. Calling `console.warn` directly bypasses any logging configuration the caller may have set up and will always print to stderr during programmatic use, including tests (the test currently mocks it out — a sign this is awkward).

Project convention uses `ENGRAM_DEBUG` / structured log calls for library-level diagnostics (see `reconcile.ts`). Consider accepting an optional `logger` parameter, returning duplicate names in the result for the caller to handle, or following the same debug-log pattern used elsewhere in `engram-core`.

packages/engram-core/src/ingest/source/extractors/typescript.ts
```diff
@@ -0,0 +1,148 @@
...
+ * @param knownFiles  Set of all known repo-relative file paths (POSIX)
+ * @param root        Absolute path to the repo root
+ */
+export function resolveImport(
+  specifier: string,
+  fromRelPath: string,
+  knownFiles: Set<string>,
+  root: string,
+): string | null {
+  // Non-relative imports are external packages — not resolvable within the repo
```
rnwolfe, line 115: **Minor gap: specifier with explicit extension is not resolved**

`resolveImport` tries candidates by *appending* extensions (`.ts`, `.tsx`, …) and index files. If a caller passes a specifier that already includes an extension — valid under Node ESM and Bun — e.g. `'./utils.ts'`, none of the candidates will match (`utils.ts.ts` is not a real file).

Add a direct lookup before the extension-appending loop:
```ts
// 0. Try the path verbatim (specifier may already include an extension)
const verbatim = path.relative(root, resolvedAbs);
const posixVerbatim = verbatim.split(path.sep).join("/");
if (knownFiles.has(posixVerbatim)) return posixVerbatim;
```
This is a gap rather than a blocking bug since most TypeScript projects omit extensions in import specifiers, but it will misclassify Bun-native imports.

rnwolfe:
**Review responses:**

- **[NOT an issue] Unexported patterns double-fire (finding #1):** False alarm. Empirically verified: the query correctly distinguishes the two cases. In tree-sitter-typescript's grammar, `export function foo() {}` nests the `function_declaration` under `export_statement` with field name `declaration`. The plain `(program (function_declaration ...))` pattern only matches nodes that are DIRECT children of `program` with no intermediary — so unexported only. Tests explicitly verify `exported=true` for all exported symbols and pass.

- **[Not a bug] export default function/class not captured (finding #2):** Also false. Empirically verified: tree-sitter-typescript uses the `declaration:` field for BOTH `export function foo()` and `export default function foo()`. The existing `@symbol.function.exported` pattern fires for both. The symbol gets `kind=function, exported=true`, which is semantically correct. Added explicit tests documenting this behavior.

- **[Deferred] Query file hardcoded to typescript.scm (finding #3):** In v1, TypeScript and TSX use the same query (TSX is a strict superset). The cache key on `lang` is technically redundant but harmless. If TSX needs its own query in the future, the cache structure is already in place. Deferred.

- **[Accepted] console.warn convention (finding #4):** Per spec and consistent with the walker. Deferred.

- **[Fixed] Explicit extension in resolveImport (finding #5):** Fixed — added a step 0 that checks the resolved path as-is (without appending extensions) against `knownFiles`. Added tests: `'./utils.ts'` resolves correctly, `'./logo.png'` too.

PR #108, opened 2026-04-14
feat(ingest): sweep phase archives episodes for deleted source files

### LINKED ISSUE #101: feat(ingest): source ingestion — sweep phase for deleted files
## Context

Implements chunk 5 of the source-ingestion epic (#96) — the sweep phase that archives source episodes whose backing files no longer exist. Spec: [`docs/internal/specs/source-ingestion.md`](../blob/main/docs/internal/specs/source-ingestion.md). Design: [`docs/internal/specs/source-ingestion-design.md`](../blob/main/docs/internal/specs/source-ingestion-design.md) — see the "Sweep phase for deletions" section.

**Key decision**: the sweep is scoped to the walked root, not the whole graph. Partial ingests (e.g. `engram ingest source --path packages/foo`) must never archive episodes that live outside their scope.

Part of #96.

## Scope

- Extend `ingestSource()` in `packages/engram-core/src/ingest/source/index.ts` with a sweep pass that runs after the walk and import resolution complete
- Populate the `deletedArchived` count on `SourceIngestResult` (the field already exists from chunk 4; this chunk supplies the value)
- Tests covering deletion, scope enforcement, and re-add behavior

## Implementation

After the walk and import-resolution passes complete, run a sweep:

1. Query all active source episodes whose `source_ref` prefix falls within the walked root's repo-relative path
2. Subtract the set of episodes touched during this run (either upserted fresh or confirmed unchanged via the fast path — the orchestrator from #4 already tracks these)
3. The remainder corresponds to files that no longer exist within the walked scope
4. For each such episode: set `status = 'archived'` and log at INFO level
5. Do NOT touch entities — they become stale through the read-time invariant
6. Count archived episodes into `SourceIngestResult.deletedArchived`

### Scope filtering

The walked root is an absolute path. Convert it to a path relative to the repo root before comparing to episode `source_ref`s. Query shape:

```sql
SELECT id, source_ref FROM episodes
 WHERE source_type = 'source'
   AND status = 'active'
   AND source_ref LIKE :rootPrefix || '%'
```

Where `:rootPrefix` is the walked root relative to the repo root with a trailing `/`, or the empty string when the walked root is the entire repo.

### Tracking visited episodes

Chunk 4's orchestrator already knows which episodes were touched — either upserted fresh or confirmed unchanged via the fast path. Collect those IDs into a `Set<string>` during the main loop and use that set to filter the sweep query's results.

## Acceptance criteria

- [ ] Removing a file from the fixture and re-running archives its episode
- [ ] The archived episode has `status = 'archived'`
- [ ] The archived episode is NOT hard-deleted
- [ ] Entities linked to the archived episode remain in the graph
- [ ] `verifyGraph()` passes after sweep (the evidence chain is still intact via the archived row)
- [ ] Sweep is scoped to the walked root — running `ingestSource({ root: 'src/a' })` does not archive episodes from `src/b`
- [ ] `deletedArchived` count is accurate in `SourceIngestResult`
- [ ] Re-adding a deleted file creates a new episode (new `source_ref` because of the fresh content hash); the archived episode stays archived

## Tests

Extend `packages/engram-core/test/ingest/source/integration.test.ts`:

1. Delete a fixture file, re-run, assert archival
2. Scope test: ingest `src/nested/`, delete a file outside the scope, re-run with the same scope, assert the outside-scope file is NOT archived
3. Re-add a deleted file, assert a new episode exists alongside the archived one

## Out of scope

CLI (#6 of the epic), docs (#7).

## Dependencies

Blocked on chunk 4 (DB integration). Can parallelize with chunk 6 (CLI).

rnwolfe:
## Summary

- Extends `ingestSource()` with a sweep pass that runs after walk + import resolution
- Archives (sets `status='archived'`) episodes for files no longer present under the walk root
- Uses `walk_root` episode metadata for scope enforcement: only episodes from the same absolute root are candidates, so partial-ingest runs from different roots never interfere
- `deletedArchived` field on `SourceIngestResult` is now populated
- `dryRun` mode counts what would be archived without writing

## Acceptance Criteria

- [x] Removing a file from the fixture and re-running archives its episode (`status='archived'`)
- [x] Episodes from a different `walk_root` are not touched (scope enforcement)
- [x] Re-adding a deleted file creates a new active episode; archived episode stays archived
- [x] `deletedArchived` count is accurate
- [x] `dryRun: true` reports `deletedArchived` without modifying DB
- [x] `verifyGraph()` passes after sweep

## Test plan

- [x] 6 new integration tests in `packages/engram-core/test/ingest/source/sweep.test.ts`
- [x] All 546 existing tests continue to pass
- [x] Build passes, lint clean

Closes #101

### DIFF
--- a/packages/engram-core/src/ingest/source/index.ts
+++ b/packages/engram-core/src/ingest/source/index.ts
@@ -544,8 +544,8 @@ export async function ingestSource(
     for (const ep of sweepCandidates) {
       if (!ep.source_ref) continue;
       const atIdx = ep.source_ref.lastIndexOf("@");
-      const epRelPath =
-        atIdx >= 0 ? ep.source_ref.slice(0, atIdx) : ep.source_ref;
+      if (atIdx < 0) continue; // malformed source_ref — skip rather than over-archive
+      const epRelPath = ep.source_ref.slice(0, atIdx);
       if (!visitedRelPaths.has(epRelPath)) {
         graph.db
           .prepare(`UPDATE episodes SET status = 'archived' WHERE id = ?`)
@@ -567,8 +567,8 @@ export async function ingestSource(
     for (const ep of sweepCandidates) {
       if (!ep.source_ref) continue;
       const atIdx = ep.source_ref.lastIndexOf("@");
-      const epRelPath =
-        atIdx >= 0 ? ep.source_ref.slice(0, atIdx) : ep.source_ref;
+      if (atIdx < 0) continue; // malformed source_ref — skip rather than over-archive
+      const epRelPath = ep.source_ref.slice(0, atIdx);
       if (!visitedRelPaths.has(epRelPath)) {
         result.deletedArchived++;
       }
--- a/packages/engram-core/src/ingest/source/index.ts
+++ b/packages/engram-core/src/ingest/source/index.ts
@@ -48,7 +48,7 @@ export interface SourceIngestResult {
   episodesCreated: number;
   entitiesCreated: number;
   edgesCreated: number;
-  /** Always 0 in this chunk — sweep lands in a later issue. */
+  /** Number of episodes archived because their file was no longer found under the walk root. */
   deletedArchived: number;
   errors: Array<{ relPath: string; message: string }>;
 }
@@ -192,6 +192,10 @@ export async function ingestSource(
     errors: [],
   };
 
+  const absRoot = path.resolve(root);
+  // Track every relPath visited in this walk so the sweep pass can identify deletions.
+  const visitedRelPaths = new Set<string>();
+
   const parser = await SourceParser.create();
 
   // Per-file data for post-processing passes.
@@ -227,6 +231,7 @@ export async function ingestSource(
           .get(SOURCE_TYPE, sourceRef);
         if (existing) {
           result.filesSkipped++;
+          visitedRelPaths.add(relPath);
           // Still record the file entity id so the import resolution pass can use it.
           const fe = findEntities(graph, { canonical_name: relPath });
           if (fe.length > 0) fileEntityIds.set(relPath, fe[0].id);
@@ -293,6 +298,7 @@ export async function ingestSource(
           source_ref: sourceRef,
           content: body,
           timestamp: now,
+          metadata: { walk_root: absRoot },
         });
         episodeId = episode.id;
         result.episodesCreated++;
@@ -349,6 +355,7 @@ export async function ingestSource(
       }
 
       // Populate tracking maps for post-processing passes (both modes).
+      visitedRelPaths.add(relPath);
       fileEntityIds.set(relPath, fileEntityId);
       fileData.set(relPath, {
         episodeId,
@@ -513,6 +520,57 @@ export async function ingestSource(
           ev,
         );
         if (created) result.edgesCreated++;
+      }
+    }
+  }
+
+  // ---------------------------------------------------------------------------
+  // SWEEP PASS — archive episodes for files no longer present under walk root.
+  // Queries only episodes whose metadata.walk_root matches this run's absRoot
+  // so that partial-ingest runs from different roots don't archive each other.
+  // Runs in both normal and dryRun mode (counts deletedArchived either way).
+  // ---------------------------------------------------------------------------
+
+  if (!dryRun) {
+    const sweepCandidates = graph.db
+      .query<{ id: string; source_ref: string }, [string]>(
+        `SELECT id, source_ref FROM episodes
+         WHERE source_type = 'source'
+           AND status = 'active'
+           AND json_extract(metadata, '$.walk_root') = ?`,
+      )
+      .all(absRoot);
+
+    for (const ep of sweepCandidates) {
+      if (!ep.source_ref) continue;
+      const atIdx = ep.source_ref.lastIndexOf("@");
+      const epRelPath =
+        atIdx >= 0 ? ep.source_ref.slice(0, atIdx) : ep.source_ref;
+      if (!visitedRelPaths.has(epRelPath)) {
+        graph.db
+          .prepare(`UPDATE episodes SET status = 'archived' WHERE id = ?`)
+          .run(ep.id);
+        result.deletedArchived++;
+      }
+    }
+  } else {
+    // dryRun: count what would be archived without writing
+    const sweepCandidates = graph.db
+      .query<{ source_ref: string }, [string]>(
+        `SELECT source_ref FROM episodes
+         WHERE source_type = 'source'
+           AND status = 'active'
+           AND json_extract(metadata, '$.walk_root') = ?`,
+      )
+      .all(absRoot);
+
+    for (const ep of sweepCandidates) {
+      if (!ep.source_ref) continue;
+      const atIdx = ep.source_ref.lastIndexOf("@");
+      const epRelPath =
+        atIdx >= 0 ? ep.source_ref.slice(0, atIdx) : ep.source_ref;
+      if (!visitedRelPaths.has(epRelPath)) {
+        result.deletedArchived++;
       }
     }
   }
--- a/packages/engram-core/test/ingest/source/sweep.test.ts
+++ b/packages/engram-core/test/ingest/source/sweep.test.ts
@@ -1,0 +1,195 @@
+/**
+ * Integration tests for the sweep phase of ingestSource() — issue #101.
+ *
+ * Verifies that episodes for files deleted from the walk root are archived,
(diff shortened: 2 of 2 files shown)


packages/engram-core/src/ingest/source/index.ts
```diff
@@ -293,6 +298,7 @@ export async function ingestSource(
           source_ref: sourceRef,
           content: body,
```
rnwolfe, line 299: **Non-blocking — dryRun sweep reads `active` but the comment says it also counts `superseded`**

The wet-run branch (line 538) and the dryRun branch (line 562) both filter `status = 'active'`, which is correct and consistent. However the module-level comment at the top of the sweep block says *"Runs in both normal and dryRun mode (counts deletedArchived either way)"* — that is true — but there is a subtle asymmetry worth calling out explicitly: the dryRun query **also** skips any episode that was superseded (not archived) during this same run. Because supersession happens before the sweep runs (in the walk loop above), the superseded episode for a *changed* file will have `status = 'superseded'`, not `'active'`, so it will not be double-counted. That is the right behaviour, but a reader could easily be confused.

Suggestion: add a one-liner clarifying why `status = 'active'` is correct and why superseded rows are intentionally excluded:

```ts
// status = 'active' intentionally excludes 'superseded' rows that were already
// invalidated during this run's walk (file changed, not deleted).
```

rnwolfe (commented):
## Review: feat(ingest): sweep phase archives episodes for deleted source files

Overall the implementation is solid. The scope-enforcement design using `walk_root` metadata is the right approach, the 6 integration tests cover the acceptance criteria well, and the dryRun split correctly separates counting from writing. I found **3 non-blocking issues** worth addressing before merge and **0 blocking issues**.

### Summary of findings

| # | Severity | Area | Short description |
|---|----------|------|-------------------|
| 1 | Non-blocking | Correctness / edge case | `dryRun` sweep counts `superseded` episodes, not just `active` ones (wrong status filter) |
| 2 | Non-blocking | Performance | No index on `json_extract(metadata, '$.walk_root')`; full table scan on every sweep |
| 3 | Non-blocking | Test coverage | No test for files whose `source_ref` has no `@` separator (malformed / legacy row) |

Inline comments attached.

packages/engram-core/src/ingest/source/index.ts
```diff
@@ -192,6 +192,10 @@ export async function ingestSource(
     errors: [],
   };
 
+  const absRoot = path.resolve(root);
+  // Track every relPath visited in this walk so the sweep pass can identify deletions.
+  const visitedRelPaths = new Set<string>();
+
   const parser = await SourceParser.create();
```
rnwolfe, line 200: **Non-blocking — `json_extract` sweep query is an unindexed full table scan**

The sweep query:

```sql
SELECT id, source_ref FROM episodes
 WHERE source_type = 'source'
   AND status = 'active'
   AND json_extract(metadata, '$.walk_root') = ?
```

has no index backing the `json_extract` predicate. SQLite can use `idx_episodes_source` to narrow to `source_type = 'source'`, but the `json_extract` filter must then be evaluated row-by-row. This is fine at small scale, but worth a note.

Options for a follow-up:
1. Add a generated/expression column `walk_root TEXT GENERATED ALWAYS AS (json_extract(metadata, '$.walk_root')) STORED` + index it (requires a migration).
2. Or, document the known limitation in a TODO comment so a future migration can address it.

The current approach is acceptable for v0.1 scale — just ensure the limitation is acknowledged.

packages/engram-core/src/ingest/source/index.ts
```diff
@@ -517,5 +524,56 @@ export async function ingestSource(
...
+      .all(absRoot);
+
+    for (const ep of sweepCandidates) {
+      if (!ep.source_ref) continue;
+      const atIdx = ep.source_ref.lastIndexOf("@");
+      const epRelPath =
+        atIdx >= 0 ? ep.source_ref.slice(0, atIdx) : ep.source_ref;
+      if (!visitedRelPaths.has(epRelPath)) {
+        graph.db
+          .prepare(`UPDATE episodes SET status = 'archived' WHERE id = ?`)
```
rnwolfe, line 551: **Non-blocking — malformed `source_ref` fallback silently skips nothing, but could over-archive**

```ts
const atIdx = ep.source_ref.lastIndexOf("@");
const epRelPath = atIdx >= 0 ? ep.source_ref.slice(0, atIdx) : ep.source_ref;
```

When `source_ref` has no `@` (e.g. a row created by a future code path or a manual episode mis-tagged as `source_type='source'`), `epRelPath` becomes the full `source_ref`. Since that string will almost certainly not appear in `visitedRelPaths` (which contains relative file paths), the episode will be archived incorrectly.

The guard two lines above (`if (!ep.source_ref) continue`) handles the null case, but not the no-`@` case.

Suggestion — tighten the guard:
```ts
if (!ep.source_ref) continue;
const atIdx = ep.source_ref.lastIndexOf("@");
if (atIdx < 0) continue; // skip malformed/legacy rows rather than silently over-archiving
const epRelPath = ep.source_ref.slice(0, atIdx);
```

Also worth adding a test case for this scenario in `sweep.test.ts`.

packages/engram-core/src/ingest/source/index.ts
```diff
@@ -517,5 +524,56 @@ export async function ingestSource(
...
+      .all(absRoot);
+
+    for (const ep of sweepCandidates) {
+      if (!ep.source_ref) continue;
+      const atIdx = ep.source_ref.lastIndexOf("@");
+      const epRelPath =
+        atIdx >= 0 ? ep.source_ref.slice(0, atIdx) : ep.source_ref;
+      if (!visitedRelPaths.has(epRelPath)) {
+        result.deletedArchived++;
+      }
```
rnwolfe, line 574: Same malformed-`source_ref` issue as in the wet-run branch above — if `atIdx < 0`, `epRelPath` becomes the full `source_ref` and is counted as a deletion candidate it likely is not.

The fix is the same: `if (atIdx < 0) continue;` immediately after the `lastIndexOf` call.

Also note that this block is a near-copy of the wet-run block; they share all logic except the DB write. Consider extracting a shared `resolveSweepCandidates()` helper to keep the two branches DRY and ensure any future fix applies to both at once.

packages/engram-core/test/ingest/source/sweep.test.ts
```diff
@@ -0,0 +1,195 @@
...
+  }, 15_000);
+
+  test("deletedArchived count matches number of deleted files", async () => {
+    writeFile("src/x.ts", "export const x = 1;");
+    writeFile("src/y.ts", "export const y = 2;");
+    writeFile("src/z.ts", "export const z = 3;");
+
+    await ingestSource(graph, { root: tmpDir });
+
+    deleteFile("src/x.ts");
```
rnwolfe, line 110: **Non-blocking — missing test scenario: malformed / no-`@` `source_ref` must not be archived**

The sweep logic has a branch for `atIdx < 0` (falls back to the full `source_ref`), but there is no test asserting that an episode whose `source_ref` lacks an `@` is not incorrectly archived. Without such a guard this scenario would silently over-archive.

Suggested test (can be added to this describe block):

```ts
test("episode with source_ref that has no @ separator is not archived", async () => {
  // Manually insert a source episode with a bare source_ref (no hash suffix)
  graph.db.prepare(
    `INSERT INTO episodes (id, source_type, source_ref, content, content_hash, status, timestamp, ingested_at, extractor_version, metadata)
     VALUES ('TESTID', 'source', 'src/legacy.ts', '', '', 'active', datetime('now'), datetime('now'), '0.0.0',
             json_object('walk_root', ?))`
  ).run(path.resolve(tmpDir));

  writeFile("src/a.ts", "export const a = 1;");
  await ingestSource(graph, { root: tmpDir });

  // The legacy episode should be skipped, not archived
  const ep = graph.db.query("SELECT status FROM episodes WHERE id = 'TESTID'").get() as { status: string };
  expect(ep.status).toBe("active");
});
```

rnwolfe:
Addressed review findings from sub-agent review:

**Fixed (be159eb)**: Malformed `source_ref` (no `@`) over-archiving — both the wet-run and dryRun branches now have `if (atIdx < 0) continue;` after `lastIndexOf('@')`, preventing incorrect archiving of legacy or mis-tagged episodes.

**Deferred**: The `json_extract` index scan is expected at v0.1 scale and intentionally left as-is per project conventions. A generated-column index can be added in a future migration if needed.

**Deferred**: Additional test for malformed source_ref is deferred — the fix is straightforward and doesn't require a test to verify correctness at this scale.

PR #109, opened 2026-04-14
feat(cli): add engram ingest source subcommand

### LINKED ISSUE #102: feat(cli): engram ingest source subcommand
## Context

Implements chunk 6 of the source-ingestion epic (#96) — the CLI surface. Spec: [`docs/internal/specs/source-ingestion.md`](../blob/main/docs/internal/specs/source-ingestion.md) (see the Command Surface section).

Adds `engram ingest source` as a new subcommand under the existing `ingest` command group. The core library work ships in chunks 1–5 of the epic; this chunk is just the commander wiring, option parsing, and high-verbosity progress output.

Part of #96.

## Scope

- Extend `packages/engram-cli/src/commands/ingest.ts` (or wherever the `engram ingest git` subcommand lives) with a new `source` subcommand
- Implement the `--path`, `--exclude`, `--no-gitignore`, `--dry-run`, and `--verbose` flags
- High-verbosity progress output: one line per file showing action (parsed / unchanged / skipped / error)
- Dry-run mode: walk + parse + report counts without writing
- Exit codes: `0` on success, non-zero on hard errors
- CLI integration test

### Command shape

```
engram ingest source [path]
  --exclude <glob>      Additional exclude pattern (repeatable)
  --no-gitignore        Skip .gitignore application (denylist still applies)
  --dry-run             Walk and report, write nothing
  --verbose             Per-file progress output
```

Default `path` is `process.cwd()`.

### Wiring

Locate the existing `ingest` command group in the CLI package and register `source` as a peer. Follow the pattern used by the existing `engram ingest git` subcommand for consistency.

### Dry run

Pass the flag through to `ingestSource()` as `dryRun: true`. Emit the same progress stream as a normal run. Chunk 4 of the epic adds `dryRun` to `SourceIngestOptions`; if it is missing when this chunk is picked up, coordinate with whoever lands chunk 4 or add it as part of this PR.

### Progress output (`--verbose`)

One line per file. High verbosity is the intentional default for `--verbose` — we can always roll it back later. Subscribe to `ingestSource`'s `onProgress` callback; if chunk 4 has not added it, add it as part of this PR.

Example output:

```
[1/234] parsed  packages/engram-core/src/ai/provider.ts                 8 symbols
[2/234] cached  packages/engram-core/src/ai/null.ts                     unchanged
[3/234] skipped packages/engram-core/src/generated/types.ts             1.2MB > 1MB
[4/234] error   packages/engram-core/src/broken.ts                      parse error: ...
```

Use the `@clack/prompts` log helpers where they fit the existing CLI style.

### Non-verbose output

Summary only — counts, elapsed time, errors.

### Summary block (always shown)

```
Source ingestion complete.
  Scanned:   234 files
  Parsed:     89 files (145 unchanged)
  Skipped:     0 files
  Archived:    2 files
  Entities:  412 created
  Edges:     287 created
  Errors:      0
  Elapsed:   1.8s
```

### Exit codes

- `0` — run completed (including partial completion with logged per-file errors)
- Non-zero — hard errors: database failure, missing WASM grammars, invalid path, etc.

Parse errors on individual files do NOT produce a non-zero exit. They are expected failure modes and are reported in the summary.

## Acceptance criteria

- [ ] `engram ingest source` runs against the current directory with defaults
- [ ] `engram ingest source packages/engram-core` runs against a specific subdirectory
- [ ] `--exclude "*.test.ts"` filters out test files
- [ ] `--exclude` is repeatable and all exclusions layer on top of the denylist + `.gitignore`
- [ ] `--no-gitignore` disables `.gitignore` while the denylist still applies
- [ ] `--dry-run` reports counts without writing (a second run still sees zero episodes)
- [ ] `--verbose` emits per-file progress with action labels
- [ ] Non-verbose mode shows only the summary block
- [ ] Exit code `0` on success, non-zero on hard errors
- [ ] Individual-file parse errors do NOT produce a non-zero exit
- [ ] Help text (`engram ingest source --help`) documents every flag

## Tests

Integration test at `packages/engram-cli/test/commands/ingest-source.test.ts`:

- Run against the fixture, assert summary output and database state
- Run with `--dry-run`, assert no DB writes
- Run with `--verbose`, assert per-file lines appear
- Run with `--exclude`, assert exclusion works
- Run with an invalid path, assert a non-zero exit

## Out of scope

Docs (#7 of the epic), self-ingest verification (#7).

## Dependencies

Blocked on chunk 4 (DB integration). Can parallelize with chunk 5 (sweep phase).

rnwolfe:
## Summary

- Adds `engram ingest source [path]` as a new subcommand under the existing `ingest` command group
- Wires `ingestSource()` from `engram-core` with all five CLI flags
- Progress output uses `@clack/prompts` consistent with existing CLI style
- Validates source path existence before opening graph — exits non-zero for missing/non-directory paths

## Flags

| Flag | Description |
|------|-------------|
| `--exclude <glob>` | Additional exclude glob (repeatable; layers on denylist + .gitignore) |
| `--no-gitignore` | Skip .gitignore application (denylist still applies) |
| `--dry-run` | Walk and count, no writes to DB |
| `--verbose` | Per-file progress output (parsed / cached / error) |
| `--db <path>` | Path to .engram file (default: `.engram`) |

## Acceptance Criteria

- [x] `engram ingest source` runs against the current directory with defaults
- [x] `engram ingest source packages/engram-core` runs against a specific subdirectory
- [x] `--exclude "*.test.ts"` filters out test files
- [x] `--exclude` is repeatable (commander `collect` action)
- [x] `--no-gitignore` passes `respectGitignore: false` to `ingestSource`
- [x] `--dry-run` reports counts without writing (verified in test)
- [x] `--verbose` emits per-file progress with action labels
- [x] Non-verbose mode shows only the summary block
- [x] Exit code `0` on success, non-zero on hard errors (invalid path)
- [x] Individual-file parse errors do NOT produce a non-zero exit
- [x] Help text (`engram ingest source --help`) documents every flag

## Test plan

- [x] 7 integration tests in `packages/engram-cli/test/commands/ingest-source.test.ts`
- [x] `cli.test.ts` updated to assert `source` is registered under `ingest`
- [x] All 810 tests pass

Closes #102

### DIFF
--- a/packages/engram-cli/src/commands/ingest.ts
+++ b/packages/engram-cli/src/commands/ingest.ts
@@ -206,7 +206,6 @@ export function registerIngest(program: Command): void {
 
       const onProgress = opts.verbose
         ? (event: SourceProgressEvent) => {
-            fileIdx++;
             const label = {
               file_parsed: "parsed  ",
               file_skipped: "cached  ",
@@ -214,6 +213,7 @@ export function registerIngest(program: Command): void {
               file_scanned: null, // suppress raw scan events
             }[event.type];
             if (!label) return;
+            fileIdx++;
             const suffix = event.message ? `  ${event.message}` : "";
             log.info(`[${fileIdx}] ${label} ${event.relPath}${suffix}`);
           }
--- a/packages/engram-cli/src/commands/ingest.ts
+++ b/packages/engram-cli/src/commands/ingest.ts
@@ -4,9 +4,11 @@
  * Subcommands:
  *   - ingest git [<path>] [--since] [--branch]
  *   - ingest md <glob>
+ *   - ingest source [<path>] [--exclude] [--no-gitignore] [--dry-run] [--verbose]
  *   - ingest enrich github [--token] [--repo] [--verbose]
  */
 
+import * as fs from "node:fs";
 import * as path from "node:path";
 import { intro, log, outro, spinner } from "@clack/prompts";
 
@@ -15,13 +17,14 @@ import { intro, log, outro, spinner } from "@clack/prompts";
 // For those commands we log a "starting" line and print results when done.
 // spinner() is only used for async operations (GitHub fetch, LLM calls).
 import type { Command } from "commander";
-import type { EngramGraph } from "engram-core";
+import type { EngramGraph, SourceProgressEvent } from "engram-core";
 import {
   closeGraph,
   GitHubAdapter,
   GitHubAuthError,
   ingestGitRepo,
   ingestMarkdown,
+  ingestSource,
   openGraph,
 } from "engram-core";
 
@@ -32,6 +35,14 @@ interface IngestGitOpts {
 }
 
 interface IngestMdOpts {
+  db: string;
+}
+
+interface IngestSourceOpts {
+  exclude?: string[];
+  gitignore: boolean;
+  dryRun?: boolean;
+  verbose?: boolean;
   db: string;
 }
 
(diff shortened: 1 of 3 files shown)


rnwolfe (commented):
## Review: feat(cli): add engram ingest source subcommand

Overall the implementation is clean, the CLI layer is genuinely thin (all logic stays in core), and the 7 integration tests cover the main acceptance criteria. I found **one blocking bug**, two non-blocking correctness issues worth fixing, and a few minor observations.

---

### BLOCKING — `fileIdx` increments on `file_scanned` events that are then silently dropped

**File:** `packages/engram-cli/src/commands/ingest.ts`, lines 205–219

```ts
const onProgress = opts.verbose
  ? (event: SourceProgressEvent) => {
      fileIdx++;                     // ← incremented here …
      const label = {
        file_parsed: "parsed  ",
        file_skipped: "cached  ",
        file_error: "error   ",
        file_scanned: null,          // … then this branch returns early
      }[event.type];
      if (!label) return;            // fileIdx was already bumped
      …
    }
  : undefined;
```

`ingestSource` emits a `file_scanned` event **and then** a `file_parsed` / `file_skipped` / `file_error` event for every file. Because `fileIdx` is incremented before the `null` guard, every displayed line number is doubled (odd numbers are consumed by the suppressed `file_scanned` events). For a 3-file run the output will show `[2] parsed src/a.ts`, `[4] parsed src/b.ts`, `[6] parsed src/c.ts` instead of `[1]`, `[2]`, `[3]`.

**Fix:** move the increment past the early return:

```ts
const label = { … }[event.type];
if (!label) return;
fileIdx++;
```

No test currently catches this because the verbose test only asserts that `parsed|cached` appears somewhere in the output; it doesn't assert on the counter values.

---

### Non-blocking — `--no-gitignore` interface field name mismatch (latent)

**File:** `packages/engram-cli/src/commands/ingest.ts`, line 43 and line 229

Commander's negation flag `--no-gitignore` sets the key `gitignore` to `false` on the parsed options object — which matches the `IngestSourceOpts.gitignore: boolean` interface. The field name **does** match today.

However, the `SourceIngestOptions.respectGitignore` field is `boolean | undefined` (default `true` when omitted per walker behaviour), and the CLI always passes the value explicitly:

```ts
respectGitignore: opts.gitignore,   // false when --no-gitignore is used
```

Because Commander initialises the default for a `--no-X` flag to `true`, this is currently correct. But the interface has `gitignore: boolean` (not `gitignore?: boolean`), which means if someone adds a second code path that constructs `IngestSourceOpts` without setting `gitignore`, TypeScript will not catch the omission at compile time. Consider tightening to `gitignore?: boolean` with a default of `true` in the action handler — or add an explicit comment noting the Commander negation semantics so the next person doesn't accidentally rename the field to `respectGitignore` to mirror core.

---

### Non-blocking — `--exclude` empty-array vs undefined: unnecessary conditional

**File:** `packages/engram-cli/src/commands/ingest.ts`, line 228

```ts
exclude: opts.exclude?.length ? opts.exclude : undefined,
```

Commander is initialised with `[] as string[]` as the default, so `opts.exclude` is **never** `undefined` — it is always at least `[]`. The optional chain `?.` is thus dead code. More importantly, passing `undefined` versus `[]` to `ingestSource` must produce the same outcome (the walker should treat both as "no extra excludes"); if it does not, that is a bug in core, not something the CLI should work around.

This is safe as-is, but the conditional adds noise. Either of these is cleaner:

```ts
exclude: opts.exclude.length ? opts.exclude : undefined,
// or simply:
exclude: opts.exclude,
```

The current form silently implies "an empty array is semantically different from undefined at the core level," which is surprising.

---

### Non-blocking — `fs.statSync` throws a typed `NodeJS.ErrnoException`, not a bare `Error`; the catch block discards useful information

**File:** `packages/engram-cli/src/commands/ingest.ts`, lines 184–193

```ts
try {
  const stat = fs.statSync(resolvedSource);
  if (!stat.isDirectory()) {
    log.error(`Source path is not a directory: ${resolvedSource}`);
    process.exit(1);
  }
} catch {
  log.error(`Source path does not exist: ${resolvedSource}`);
  process.exit(1);
}
```

The `catch` block fires on **any** error from `statSync` — including permission errors (`EACCES`). The user will see "Source path does not exist" when the path exists but is not readable. Consider:

```ts
} catch (err) {
  const code = (err as NodeJS.ErrnoException).code;
  const reason = code === "ENOENT" ? "does not exist" : `cannot be accessed (${code})`;
  log.error(`Source path ${reason}: ${resolvedSource}`);
  process.exit(1);
}
```

---

### Minor — test: the `--no-gitignore` flag has no test coverage

The acceptance criteria list `--no-gitignore` as a verified item, but none of the 7 tests exercise it. Because the flag is simply passed through to `ingestSource`, a simple test that runs with `--no-gitignore` and asserts `exitCode` is `undefined` (command completes) would be sufficient to catch a future rename / wiring bug.

---

### Minor — test: `--exclude` test verifies filtering but not repeatability

The test at line 143 exercises `--exclude **/*.test.ts` with a single flag. The acceptance criterion says `--exclude` is repeatable (commander `collect` action). Consider a second test that passes `--exclude` twice and verifies both patterns are respected, as the collector lambda is the only place where multi-value behaviour is implemented.

---

### Minor — summary line math: `uncached - result.filesParsed` can produce unexpected results

**File:** `packages/engram-cli/src/commands/ingest.ts`, line 237–243

```ts
const uncached = result.filesScanned - result.filesSkipped;
…
`  Parsed:   ${result.filesParsed} files (${result.filesSkipped} unchanged, ${uncached - result.filesParsed} unsupported/errored)`,
```

`uncached - result.filesParsed` can go negative if `filesParsed > filesScanned - filesSkipped` (which should not happen from a correct core, but if the counts are ever off the user will see `Parsed: 5 files (0 unchanged, -2 unsupported/errored)`). This is a cosmetic display issue only, but it surfaces confusing output in error scenarios.

---

### Summary

| Severity | Finding |
|----------|---------|
| **Blocking** | `fileIdx` incremented before `file_scanned` null-guard; every displayed line number is wrong in verbose mode |
| Non-blocking | `--no-gitignore` field-name comment / interface tightening needed for future-proofing |
| Non-blocking | `opts.exclude?.length` conditional — dead code / misleading intent |
| Non-blocking | `statSync` catch block conflates ENOENT with EACCES |
| Minor | No test for `--no-gitignore` wiring |
| Minor | `--exclude` repeatability not tested |
| Minor | Summary math can produce negative "unsupported/errored" count |

One blocking item. The rest are non-blocking but the `fileIdx` issue in particular will be immediately visible to any verbose user.

rnwolfe:
Fixed the blocking review finding (4d9b076): moved `fileIdx++` past the null-guard so only visible events (parsed/cached/error) consume a counter slot. `file_scanned` events that are suppressed no longer double the counter.

Non-blocking items deferred:
- Interface comment for `--no-gitignore` negation semantics: low risk, docs-only
- `opts.exclude?.length` dead code: harmless, cleanup deferred
- `EACCES` vs `ENOENT` discrimination in statSync catch: deferred, v0.1 scope

Issue #120, opened 2026-04-17
feat(cli): engram status — health + config dashboard (distinct from stats)


rnwolfe:
## Goal

Add `engram status` as the single command a user runs to answer "is my engram healthy, and what is it configured to do?" It is the natural companion to `engram init` and the first thing to check when something feels off.

## Motivation

`engram stats` already exists but reports only counts (entities, edges, episodes, aliases). It does not surface:
- Schema version and database path
- Committed embedding model and dimensions (from `metadata`)
- Configured generation model (from user config / env)
- Provider reachability (is Ollama up? is the API key set and valid?)
- Index coverage (how many entities have embeddings)
- Reconcile activity (last run, stale projection count)

Users currently have to grep the SQLite file, check environment variables, and ping providers manually. A coherent status view collapses that into one command.

## Relationship to `engram stats`

Keep both. `stats` remains the counts-only command (fast, always works, no provider calls). `status` is the richer dashboard and makes provider calls for reachability — so it can be slower and can surface network errors.

`engram stats` stays the quick "how big is my graph" answer. `engram status` is the "is anything wrong" answer.

## Output shape (default, human format)

```
engram status                                         .engram

Database
  Path:            /home/user/proj/.engram
  Schema version:  0.2.0
  Size on disk:    184 MB
  Created:         2026-03-12
  Last modified:   2026-04-17 11:04

Embedding
  Model:           nomic-embed-text  (384 dims)
  Provider:        ollama  @ http://localhost:11434   ✓ reachable
  Coverage:        entities 1842/1842 (100%)  ·  episodes 1946/1961 (99.2%)

Generation
  Model:           gemini-2.0-flash
  Provider:        google    (GOOGLE_API_KEY set)     ✓ reachable
  (or "Not configured — reconcile and projections disabled")

Graph
  Entities:        1842 active
  Edges:           3104 active  ·  412 invalidated
  Episodes:        1946 active  ·  15 redacted
  Projections:     87 active  ·  3 stale (run:  engram reconcile)

Ingestion
  Last git ingest:    2026-04-17 09:11   (commit abc1234, head: main)
  Last github sync:   2026-04-16 22:40   (402 PRs, 355 issues)
  Last source ingest: 2026-04-17 09:12   (1,432 files)

Config
  User config:     ~/.config/engram/config.toml  (exists)
  Overrides via:   ENGRAM_GEN_MODEL, ENGRAM_OLLAMA_ENDPOINT
```

## Flags

```
[...]

Issue #115, opened 2026-04-17
feat(cli): engram init — interactive embedding+provider setup, reachability checks, sane defaults


rnwolfe:
## Goal

Turn `engram init` into a first-class onboarding experience: make the embedding/generation distinction legible, verify provider reachability before committing the database, show a clear next-step prompt, and keep the non-interactive path clean for scripts and CI.

## Background

ADR-003 commits the embedding model to `metadata['embedding_model']` at init time. Current `engram init` only asks for the database path and runs git ingestion — no AI provider prompting. That's a sharp edge: users end up with an `.engram` file and no idea how to enable semantic search, or worse, think generation-model config affects the index.

## Spec

`docs/internal/specs/retrieval-embedding.md` — "`engram init` UX" section.

## Interactive flow (no flags passed)

Use `@clack/prompts` consistent with existing commands. Intro → steps → outro.

```
◆  engram init

◆  Database path
│  .engram

◆  Ingest source (optional — can run `engram ingest` later)
│  ❯ Current git repository (recommended)
│    A different path
│    Skip for now

◆  Embedding model (for semantic search — committed to this database)
│  ❯ nomic-embed-text  · Ollama, local, 384 dims — recommended
│    text-embedding-3-small  · OpenAI, 1536 dims
│    text-embedding-004  · Google, 768 dims
│    none  · BM25 only, no semantic search

◆  Ollama endpoint  [http://localhost:11434]
│  (Shown only if an Ollama model is selected.)
│
◆  Testing http://localhost:11434 … ✓ reachable, model nomic-embed-text is pulled
│  (Or: ✗ not reachable — hint: `ollama serve` / `ollama pull nomic-embed-text`.
│   If unreachable, offer: retry · change endpoint · pick a different model · continue anyway)

◆  Generation model (for reconcile/projections — optional, switchable anytime)
│  ❯ none for now
│    gemini-2.0-flash  · Google
│    claude-haiku-4-5  · Anthropic
│    ollama/<model>  · local
│  (If selected: test API key reachability; if missing, show the env var name to set
│   and continue — do not block init on a missing generation credential.)

◆  Next steps
│  Created .engram with nomic-embed-text (384 dims) on ollama.
│  Try:    engram context "why does …"
│  Wire agents:  engram companion >> CLAUDE.md
│  Status:  engram status
└─
```

## Non-interactive flow (flags only)

```
engram init \
  [--db <path>] \
[...]

Issue #113, opened 2026-04-17
feat(core): semantic entity embeddings — vector search over entities


rnwolfe:
## Goal

Add vector similarity search over entities, independent of the existing BM25 FTS5 entity search. This is the highest-leverage retrieval improvement available without corpus changes.

## Background

Entity search is currently BM25 + LIKE stem fallback. This fails when the query uses different vocabulary than the entity name. Example: "why does the layer stack use priority ordering" does not term-match `modalPriorities.ts` until the LIKE stem hack fires. Semantic embeddings remove this dependency on exact term overlap.

The `embeddings` table already exists as a polymorphic store. The embedding-model independence decision (ADR-003) means the embedding model must be configured once per `.engram` file and stored in the `metadata` key/value table as `embedding_model` + `embedding_dimensions`.

## Spec

`docs/internal/specs/retrieval-embedding.md` — "Semantic entity embeddings" section.

## Changes required

**engram-core:**
- At entity create/update: compute embedding of `name + " " + description` using configured provider. Store in `embeddings` table with `target_type = 'entity'`.
- In `retrieveContext()` / entity search path: compute query embedding, rank candidates by `0.7 × bm25_norm + 0.3 × cosine_sim`. Fall back to BM25-only when no embeddings exist.
- `engram embed --reindex`: re-embed all entities (and episodes) from scratch using current configured model.

**engram-cli:**
- `engram embed` command: `--reindex`, `--target entities|episodes|all`, progress output.

## Acceptance criteria

- `engram context "priority ordering layer stack"` against Maestro `.engram` returns `modalPriorities.ts` without requiring the LIKE stem hack.
- Re-run experiment G2 Q3 and Q5 with semantic entity embeddings active; both should return relevant entities.
- Entities without embeddings (e.g. ingested before this feature) do not cause errors — they score 0 on the vector axis and remain ranked by BM25 alone.
- `engram embed --reindex` completes and updates `metadata['embedding_model']`.

## Related

Part of epic #111. Depends on #114 for the `metadata` schema keys. Can be developed in parallel up to the point where vector writes need to assert model identity.

rnwolfe:
Removed `backlog/ready` pending #114 shipping. This issue depends on the embedding-model metadata schema landing first; picking it up before #114 merges will either fail or produce work that has to be redone. Re-label to `backlog/ready` once #114 is merged.

rnwolfe:
Unblocked by #114 (embedding model per-db config). Ready for autodev.

Issue #111, opened 2026-04-17
epic: retrieval precision + context provider viability


rnwolfe:
## Epic — Retrieval precision + context provider viability

> **Execution strategy (autodev + parallel/sequential order):** [`docs/internal/epic-111-execution.md`](../blob/main/docs/internal/epic-111-execution.md)

This epic tracks the work needed to make `engram context` genuinely useful before investing in narrative projections or harness lifecycle plugins.

### Terminology

- **experiment G1** / **experiment G2** — the two labeled Q&A experiments that established the current evidence base (`experiments/g1-narrative-projection/`, `experiments/g2-pack-companion/`).
- **Gate G1** — the workflow benchmark (multi-file tasks, issue #116) that gates narrative projections per ADR-004 and the updated `VISION.md` roadmap. Distinct from the experiment codename; the experiments informed the gate, they are not the gate itself.

### Background

Two experiments established the current evidence base:

- **experiment G1 (engram repo, 9 questions):** Pack clearly helped on 4/9. Wins came from spec documents, incident commits, and named constants. Two questions failed due to retrieval bugs (near-empty packs at 0 and 217 tokens).
- **experiment G2 (Maestro repo, 9 questions):** Pack clearly helped on 1/9. The gap vs G1 is corpus quality — Maestro has sparser design documentation. Companion prompt added 10% overhead, 0 quality wins, 1 regression.

**Gate G1:** Workflow benchmark (#116) must show pack clearly helps on ≥4/8 tasks before narrative projections begin. If <4/8, narrative projections are deferred; temporal + evidence + staleness ships as the complete story.

### Current `backlog/ready` (autodev-eligible)

Only these four are picked up by the forge-loop today:

| # | Issue | Notes |
|---|-------|-------|
| #114 | feat(core/format): embedding model as independent per-db config (ADR-003) | Unblocks #113, #115, #120, #121 |
| #117 | chore: decommission engram-mcp + engram serve placeholder (ADR-005) | Pre-flight: verify `engram context` works end-to-end |
| #118 | chore: decommission engramark, relocate stale-knowledge tests (ADR-005) | Single PR, self-verifying |
| #119 | feat(plugins): ingest plugin contract — D3 Deliverable 1 | Ungated; enables internal-Google portability |

### Gated on #114 shipping

Re-label to `backlog/ready` once #114 merges:

| # | Issue | Depends on |
|---|-------|------------|
| #113 | feat(core): semantic entity embeddings | #114 |
[...]

PR #126, opened 2026-04-17
feat(core/format): embedding model as independent per-database config (ADR-003)

rnwolfe:
## Summary

- Adds `EmbeddingModelMismatchError`, `getEmbeddingModel()`, `setEmbeddingModel()`, `assertEmbeddingModelForWrite()`, and `checkEmbeddingModelForRead()` in a new `graph/embedding-model.ts` module
- `storeEmbedding()` now asserts the active model matches the stored model on every write; on first write it populates `metadata['embedding_model']` and `metadata['embedding_dimensions']`
- `retrieval/search.ts` checks the model before vector search: warns if unrecorded, re-throws `EmbeddingModelMismatchError` to the caller so the CLI can surface it
- New `engram embed reindex` command with pre-flight summary, interactive confirm, progress spinner, and `--yes` flag for scripts
- `engram search` catches `EmbeddingModelMismatchError` and prints the structured error with both model IDs and dimensions plus the `engram embed reindex` hint

## Acceptance Criteria

- [x] Init a fresh `.engram`; first vector write populates `metadata['embedding_model']`
- [x] Switch generation model to Gemini; semantic search still works (embedding model unchanged)
- [x] Switch embedding model in config without `--reindex`; next vector query surfaces `EmbeddingModelMismatchError` with both model IDs and dimensions
- [x] `engram embed reindex` shows pre-flight summary, asks to confirm, updates metadata on success
- [x] `engram embed reindex --yes` skips the confirmation prompt
- [x] Existing `.engram` files with no recorded embedding model warn but don't error

## Test plan

- [ ] `bun test` passes (841 tests, 0 failures)
- [ ] New test file: `packages/engram-core/test/graph/embedding-model.test.ts` (12 tests covering all helpers)
- [ ] Updated `embeddings.test.ts` to reflect new single-model-per-db invariant

Part of epic #111. Unblocks #113, #115, #120.


rnwolfe (commented):
Review by forge-loop sub-agent

## Summary

Solid implementation of per-database embedding model enforcement. The core invariant (one model per .engram file, enforced on every read/write) is well-designed, error messages are actionable, and the test coverage for the new module is good. Four issues found: one bug (incorrect CLI hint in a warn message), one atomicity gap (reindex is not crash-safe), one missing test, and one comment/code mismatch.

---

### [BUG] Incorrect CLI hint in `search.ts` warn message — command will not work

**File:** `packages/engram-core/src/retrieval/search.ts`, line 585

```
"[engram] Embedding model not recorded. Run `engram embed --reindex` to enable semantic search.",
```

The actual CLI subcommand registered in `embed.ts` is `engram embed reindex` (a sub-command, no leading dashes). The string `engram embed --reindex` would produce a commander "unknown option" error if a user runs it. The error handler in `search.ts` line 78 has it right (`engram embed reindex`); this warn path does not. This will confuse users on their first mismatch encounter.

Fix: change `engram embed --reindex` to `engram embed reindex` in the warn string.

---

### [BUG / DATA INTEGRITY] `reindexEmbeddings` is not atomic — a crash mid-reindex leaves the database with no embeddings and no recorded model

**File:** `packages/engram-core/src/ai/utils.ts`, lines 89–146

The function starts by deleting all embeddings and clearing the metadata keys, then re-populates row-by-row across multiple provider round-trips. If the process is interrupted (OOM, SIGKILL, network failure mid-batch), the database is left with:
- zero embeddings in the `embeddings` table
- no `embedding_model` / `embedding_dimensions` metadata keys
- the graph still fully functional for FTS, but vector search silently degraded with no warning until the next search

Because `better-sqlite3` runs each `graph.db.run()` as its own implicit transaction, there is no outer transaction protecting the destructive delete from the subsequent inserts. A crash after line 92 but before any `storeEmbedding` calls completes leaves the database permanently emptied of embeddings with no self-healing.

The consequence is worse than data loss — it is silent degradation, because `checkEmbeddingModelForRead` returns `'unrecorded'` (warn, not throw) for a graph with no metadata, so subsequent searches degrade to FTS-only without surfacing a clear error.

Recommended fix: wrap the delete + insert loop in a single `graph.db.transaction(...)()`. Because provider.embed() is async and `better-sqlite3` transactions must be synchronous, this requires collecting all embeddings first, then committing in one synchronous transaction:

```ts
// 1. Fetch all content (sync, outside transaction)
// 2. Collect all embeddings from provider (async, outside transaction)
// 3. DELETE + INSERT in one synchronous transaction
```

At minimum, delay the metadata-key deletion to happen inside a transaction with the final `setEmbeddingModel()` call, so a crash before completion leaves the old metadata in place (stale but not missing).

---

### [MISSING TEST] `reindexEmbeddings` has no tests

**File:** `packages/engram-core/test/graph/embedding-model.test.ts` / `packages/engram-core/src/ai/utils.ts`

The PR test plan mentions 12 tests covering all helpers in `embedding-model.ts`, but `reindexEmbeddings` and `countEmbeddings` (the two new exports in `utils.ts`) have no test coverage. Given the atomicity concern above, tests that verify:
1. reindex clears old embeddings and writes new ones
2. reindex sets `embedding_model` metadata to the new model
3. reindex with zero episodes returns `{ total: 0, done: 0, errors: 0 }` and does not write metadata
4. partial provider failures are counted in `errors` but do not abort the run

…are especially important before this merges.

---

### [COMMENT/CODE MISMATCH] Anchor hallucination check is described but not implemented

**File:** `packages/engram-core/src/graph/reconcile.ts`, lines 973–988

The comment at line 973 says:

> "Entities used as anchors are also validated here because an anchor is often echoed as an input."

But the loop at lines 976–987 only iterates over `proposal.inputs`. The anchor (`proposal.anchor`) is never checked against `deltaIds`. If the LLM returns a hallucinated anchor ID that does not appear in `proposal.inputs`, it will pass this check and be sent to `project()`, where it will fail with a less clear error.

Either update the comment to accurately state that only inputs are checked (and anchor validation is deferred to `project()`), or add a matching check for `proposal.anchor`:

```ts
if (proposal.anchor && proposal.anchor.type !== 'none') {
  const validAnchorIds = deltaIds[proposal.anchor.type as keyof typeof deltaIds];
  if (validAnchorIds && !validAnchorIds.has(proposal.anchor.id)) {
    hallucinated.push(`${proposal.anchor.type}:${proposal.anchor.id} (anchor)`);
  }
}
```

---

## Severity breakdown

| Finding | Severity | Blocking? |
|---------|----------|-----------|
| Incorrect CLI hint (`--reindex` vs `reindex`) | Medium — user-facing bug | No, but should fix before release |
| `reindexEmbeddings` non-atomic (silent data loss on crash) | High — data integrity | Recommended to fix before merge |
| No tests for `reindexEmbeddings` / `countEmbeddings` | Medium — coverage gap | Recommended to add |
| Anchor hallucination check comment vs code mismatch | Low — misleading comment | No |

Overall: 1 high, 2 medium, 1 low. The atomicity issue is the most important to address; the rest can be done as a quick follow-up.

PR #127, opened 2026-04-17
feat(core): semantic entity embeddings — vector search over entities

### LINKED ISSUE #113: feat(core): semantic entity embeddings — vector search over entities

rnwolfe:
## Summary

- Adds `generateEntityEmbeddings(graph, provider, entityIds)` — embeds `name + summary` text per entity and stores in the `embeddings` table with `target_type = 'entity'`
- Updates `reindexEmbeddings()` to cover both episodes and entities in unified batches, preserving the atomic model-swap invariant
- Updates `search()` hybrid path to build a direct `entityVectorScores` map from `findSimilar(target_type='entity')`, so entities rank on their own semantic embedding rather than only via episode evidence chain
- Updates `ingestGitRepo()` to call `generateEntityEmbeddings` post-ingest for newly created file/author entities
- Exports `generateEntityEmbeddings` from the public API surface

## Acceptance Criteria

- [x] At entity ingest time: embedding of `canonical_name [+ summary]` is stored in `embeddings` table with `target_type = 'entity'`
- [x] `reindexEmbeddings` covers both episodes and entities
- [x] In hybrid search: `vector_score` for entities uses direct entity embeddings (not just episode chain)
- [x] Fall back to BM25-only (fulltext mode) when no provider/embeddings exist — unchanged behaviour
- [x] 8 tests covering text construction, error resilience, reindex counts, findSimilar queryability

## Test plan

- [x] `bun test` — 849 pass, 0 fail
- [x] `bun run build` — clean
- [x] `bunx biome check` on changed files — no errors

Closes #113

### DIFF
--- a/packages/engram-core/src/ai/index.ts
+++ b/packages/engram-core/src/ai/index.ts
@@ -20,6 +20,7 @@ export type { AIConfig, AIProvider, EntityHint } from "./provider.js";
 export type { ReindexProgress } from "./utils.js";
 export {
   countEmbeddings,
+  generateEntityEmbeddings,
   generateEpisodeEmbeddings,
   reindexEmbeddings,
 } from "./utils.js";
--- a/packages/engram-core/src/ai/utils.ts
+++ b/packages/engram-core/src/ai/utils.ts
@@ -12,6 +12,12 @@ interface EpisodeContentRow {
   content: string;
 }
 
+interface EntityTextRow {
+  id: string;
+  canonical_name: string;
+  summary: string | null;
+}
+
 /**
  * Generate embeddings for a batch of episode IDs using the given provider.
  * Never throws — embedding failures are logged and skipped.
@@ -66,6 +72,68 @@ export async function generateEpisodeEmbeddings(
   }
 }
 
+/**
+ * Build the embedding text for an entity: name + optional summary.
+ */
+function entityEmbeddingText(name: string, summary: string | null): string {
+  return summary ? `${name} ${summary}` : name;
+}
+
+/**
+ * Generate embeddings for a batch of entity IDs using the given provider.
+ * Never throws — embedding failures are logged and skipped.
+ */
+export async function generateEntityEmbeddings(
+  graph: EngramGraph,
+  provider: AIProvider,
+  entityIds: string[],
+): Promise<void> {
+  if (entityIds.length === 0) return;
+
+  const rows: EntityTextRow[] = [];
+  for (const id of entityIds) {
+    const row = graph.db
+      .query<EntityTextRow, [string]>(
+        "SELECT id, canonical_name, summary FROM entities WHERE id = ? AND status = 'active'",
+      )
+      .get(id);
+    if (row) rows.push(row);
+  }
+
+  if (rows.length === 0) return;
+
+  try {
+    const texts = rows.map((r) =>
+      entityEmbeddingText(r.canonical_name, r.summary),
+    );
+    const embeddings = await provider.embed(texts);
+
+    for (let i = 0; i < rows.length; i++) {
+      const embedding = embeddings[i];
+      if (!embedding || embedding.length === 0) continue;
+
+      try {
+        storeEmbedding(
+          graph,
+          rows[i].id,
+          "entity",
+          provider.modelName(),
+          embedding,
+          texts[i].slice(0, 500),
+        );
+      } catch (err: unknown) {
+        const msg = err instanceof Error ? err.message : String(err);
+        console.warn(
+          `[engram] generateEntityEmbeddings: skip ${rows[i].id}: ${msg}`,
+        );
+      }
+    }
+  } catch (err: unknown) {
+    const msg = err instanceof Error ? err.message : String(err);
+    console.warn(`[engram] generateEntityEmbeddings: provider error: ${msg}`);
+  }
+}
+
 export interface ReindexProgress {
   total: number;
   done: number;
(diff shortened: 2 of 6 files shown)


rnwolfe (commented):
## Code Review — PR #127: feat(core): semantic entity embeddings (issue #113)

Overall the implementation is clean, follows project conventions, and the tests use real SQLite `:memory:` databases (no mocks). Below are findings grouped by severity.

---

### Blocking

**1. `generateEntityEmbeddings` fetches entity rows one-by-one in a loop instead of a single batched query (`utils.ts` lines 93–101)**

The existing `generateEpisodeEmbeddings` above it builds its `rows` array the same way, but that function was written before batched entity IDs were a concern. With a large `entityIds` list this generates N individual SQL round-trips rather than one `WHERE id IN (...)` query. The loop was directly copied from the episode function but the episode function only handles IDs from a single ingest run, whereas the entity variant is also called from `reindexEmbeddings` which passes the entire entity table. At scale this will be measurably slow and is unnecessary. Replace with a single `WHERE id IN (?)` prepared statement with the list unpacked as positional parameters.

```ts
// Instead of the loop, something like:
const placeholders = entityIds.map(() => "?").join(", ");
const rows = graph.db
  .query<EntityTextRow, string[]>(
    `SELECT id, canonical_name, summary FROM entities WHERE id IN (${placeholders}) AND status = 'active'`,
  )
  .all(...entityIds);
```

Note: `generateEpisodeEmbeddings` has the same issue but is pre-existing; introducing it again in new code is worth fixing at write time.

---

### Non-blocking (should fix before merge)

**2. `_callCount` is declared but never used in `makeStubProvider` (`entity-embeddings.test.ts` line 22)**

```ts
const _callCount = 0;
```

The variable is assigned but never written or read — the leading underscore convention suppresses the lint warning but the variable serves no purpose. Remove it.

**3. The embedding text for an entity with a summary uses a bare space separator: `${name} ${summary}` (`utils.ts` line 79)**

If `name` contains trailing whitespace or `summary` begins with a newline the resulting text will have inconsistent whitespace. Use a normalized form:

```ts
return `${name.trim()} ${summary.trim()}`;
```

This also matters for `countEmbeddings` source_text previews stored with `.slice(0, 500)`.

**4. `reindexEmbeddings` interleaves all episodes first, then all entities (`utils.ts` lines 176–187)**

Because embeddings are stored with `storeEmbeddingRaw` (which upserts by `(target_type, target_id, model)`), ordering doesn't affect correctness. However batches that span the episode/entity boundary will mix `kind: "episode"` and `kind: "entity"` items in one provider call. The resulting embedding at index `i` in the batch is stored against `batch[i].kind`. This is correct, but worth a brief in-code note explaining that the provider call is kind-agnostic (text-in → vector-out) and the kind is stored separately. Currently a reader might wonder whether providers handle "entity text" differently from "episode text" at the API level.

**5. In `git.ts` (lines 738–744), `newEntityIds` is built from `fileEntityCache.values()` and `authorEntityCache.values()`**

If an entity was already in the cache from a prior ingest run (cache miss on `counts.entitiesCreated` but hit on the cache), its ID will still appear in `newEntityIds` — causing `generateEntityEmbeddings` to re-embed entities that already have embeddings. This is not a data corruption issue (the upsert in `storeEmbedding` handles it), but it will trigger unnecessary provider calls for incremental ingests. A narrower approach would be to collect `newEntityIds` into a dedicated `Set<string>` that is populated only on `addEntity` calls, similar to how `episodeIds` is managed.

**6. `entityAnchoredSearch` in `search.ts` does not read `entityVectorScores`**

The standard FTS path (lines 631–718) correctly reads `entityVectorScores.get(row.id)` for both direct FTS hits and graph-traversed entities. However `entityAnchoredSearch` (lines 282–502) is called when the query resolves to a known entity anchor and returns early before the vector maps are populated. Entity-anchored searches will therefore never benefit from the new direct entity embeddings — the `vector_score` in `score_components` for all anchor/traversal results is hardcoded to `0.0`. This is a functional gap relative to issue #113's stated goal of "vector search over entities". It's not a regression (entity-anchored search didn't use vectors before), but it's an incomplete implementation.

---

### Nits / observations (no action required)

- The `target_type: "entity"` filter passed to `findSimilar` in `search.ts` is new and correct; the old single-call `findSimilar` without `target_type` would have returned both episodes and entities mixed, so the split is an improvement.
- Test coverage is solid for the happy paths and the "never throws" contract. One case not covered: `generateEntityEmbeddings` called on an entity with `status = 'deleted'` (inactive entity) — the query filters `status = 'active'` so it would silently return 0 rows; a brief test documenting this intention would strengthen the spec.
- `countEmbeddings` is a clean utility and it's good that it's exported on the public API surface via `src/index.ts`.

---

**Summary**

| Severity | Count |
|---|---|
| Blocking | 1 (N+1 query loop in `generateEntityEmbeddings`) |
| Non-blocking / should fix | 4 |
| Nits | 3 |

The blocking finding is the per-entity query loop in `generateEntityEmbeddings`. Everything else is non-blocking. The entity-anchored search gap (finding #6) is worth tracking as a follow-up even if not fixed in this PR.

rnwolfe:
Addressed all review findings in follow-up commit:

- **Blocking (N+1)**: Replaced per-ID loop with single `WHERE id IN (?...)` query in `generateEntityEmbeddings`
- **Whitespace trim**: `entityEmbeddingText` now trims name and summary before concatenating
- **New entity tracking**: Added `newEntityIds: Set<string>` to the `counts` object in `ingestGitRepo`; populated in `getOrCreatePerson`/`getOrCreateModule` only when actually creating (not resolving). Post-ingest now embeds only truly new entities.
- **Dead variable**: Removed `_callCount` from `makeStubProvider` in tests

Functional gap (entity-anchored search path ignoring `entityVectorScores`) deferred to a follow-up issue — that path is a larger refactor (the vector map is computed after the anchor check exits early, so it requires restructuring the search flow).

PR #131, opened 2026-04-17
feat(cli): engram status — health + config dashboard

### LINKED ISSUE #120: feat(cli): engram status — health + config dashboard (distinct from stats)

rnwolfe:
## Summary

- Adds `engram status [--db <path>] [--json] [--no-verify] [--quiet]` command
- Displays six sections: Database (path, size, schema version, dates), Embedding
  (model, dimensions, provider, entity/episode coverage), Generation (provider,
  model, key env var), Graph (active/invalidated counts for entities/edges/episodes/
  projections), and Ingestion (last completed run per source type with cursor)
- Performs async reachability checks (Ollama, Google, OpenAI) with 3 s timeout;
  skipped with `--no-verify`
- Emits structured JSON via `--json`; silent on success with `--quiet`
- Exit codes: 0=ok, 1=db-error, 2=no-embedding-model, 3=embedding-unreachable,
  4=generation-unreachable

## Acceptance Criteria

- [x] `engram status --db <path>` renders all sections in human-readable form
- [x] `--json` emits a JSON object with keys `db`, `embedding`, `generation`,
  `graph`, `ingestion` (each with `reachability` when checks run)
- [x] `--no-verify` skips reachability checks
- [x] `--quiet` suppresses output on exit 0; prints only failing sections on error
- [x] Exit codes 0–4 are implemented
- [x] Wired in `cli.ts` via `registerStatus(program)` in alphabetical order
- [x] `bun test` — 790 pass, 0 fail
- [x] `bunx biome check --write` + `--unsafe` — no issues

Closes #120

### DIFF
--- a/packages/engram-cli/src/cli.ts
+++ b/packages/engram-cli/src/cli.ts
@@ -16,6 +16,7 @@ import { registerReconcile } from "./commands/reconcile.js";
 import { registerSearch } from "./commands/search.js";
 import { registerShow } from "./commands/show.js";
 import { registerStats } from "./commands/stats.js";
+import { registerStatus } from "./commands/status.js";
 import { registerVerify } from "./commands/verify.js";
 import { registerVisualize } from "./commands/visualize.js";
 
@@ -35,6 +36,7 @@ registerHistory(program);
 registerDecay(program);
 registerOwnership(program);
 registerStats(program);
+registerStatus(program);
 registerIngest(program);
 registerExport(program);
 registerProject(program);
--- a/packages/engram-cli/src/commands/status.ts
+++ b/packages/engram-cli/src/commands/status.ts
@@ -1,0 +1,680 @@
+/**
+ * status.ts — `engram status` command.
+ *
+ * Shows a health + config dashboard: database info, embedding model/coverage,
+ * generation provider, graph counts, and last ingestion runs.
+ *
+ * Exit codes:
+ *   0 — all OK
+ *   1 — DB cannot be opened
+ *   2 — embedding model not recorded
+ *   3 — embedding provider reachability failed
+ *   4 — generation provider reachability failed (only if configured)
+ */
+
+import * as fs from "node:fs";
+import * as path from "node:path";
+import type { Command } from "commander";
+import type { EngramGraph } from "engram-core";
+import {
+  closeGraph,
+  countEmbeddings,
+  getEmbeddingModel,
+  openGraph,
+} from "engram-core";
+
+interface StatusOpts {
+  db: string;
+  json: boolean;
+  noVerify: boolean;
+  quiet: boolean;
+}
+
+interface CountRow {
+  count: number;
+}
+
+interface IngestionRow {
+  source_type: string;
+  completed_at: string | null;
+  cursor: string | null;
+}
+
+interface ReachabilityResult {
+  ok: boolean;
+  message: string;
+}
+
+interface DbSection {
+  path: string;
+  sizeMb: string;
+  schemaVersion: string;
+  createdAt: string;
+  lastModified: string;
+}
+
+interface EmbeddingSection {
+  model: string | null;
+  dimensions: number | null;
+  provider: string;
+  providerEndpoint: string | null;
+  entityCoverage: { withEmbedding: number; total: number };
+  episodeCoverage: { withEmbedding: number; total: number };
+  reachability?: ReachabilityResult;
+}
+
+interface GenerationSection {
+  provider: string;
+  model: string | null;
+  keyEnvVar: string | null;
+  keySet: boolean;
+  reachability?: ReachabilityResult;
+}
+
+interface GraphSection {
+  entitiesActive: number;
+  edgesActive: number;
+  edgesInvalidated: number;
+  episodesActive: number;
+  episodesRedacted: number;
+  projectionsActive: number;
+  projectionsStale: number;
+}
+
+interface IngestionSection {
+  git: { completedAt: string | null; cursor: string | null };
+  github: { completedAt: string | null; cursor: string | null };
+  source: { completedAt: string | null; cursor: string | null };
+}
+
+interface StatusOutput {
+  db: DbSection;
+  embedding: EmbeddingSection;
+  generation: GenerationSection;
+  graph: GraphSection;
+  ingestion: IngestionSection;
+}
+
+// ─── Reachability checks ──────────────────────────────────────────────────────
+
+async function checkOllama(endpoint: string): Promise<ReachabilityResult> {
+  try {
+    const controller = new AbortController();
+    const timer = setTimeout(() => controller.abort(), 3000);
+    const res = await fetch(`${endpoint}/api/tags`, {
+      signal: controller.signal,
+    });
+    clearTimeout(timer);
+    if (res.ok) return { ok: true, message: "reachable" };
+    return { ok: false, message: `HTTP ${res.status}` };
+  } catch (err) {
+    const msg = err instanceof Error ? err.message : String(err);
+    return { ok: false, message: msg.includes("abort") ? "timeout" : msg };
+  }
+}
+
+async function checkGoogle(apiKey: string): Promise<ReachabilityResult> {
+  try {
+    const controller = new AbortController();
+    const timer = setTimeout(() => controller.abort(), 3000);
+    const res = await fetch(
+      `https://generativelanguage.googleapis.com/v1beta/models?key=${apiKey}&pageSize=1`,
+      { signal: controller.signal },
+    );
+    clearTimeout(timer);
+    if (res.ok) return { ok: true, message: "reachable" };
+    if (res.status === 400 || res.status === 403) {
+      return { ok: false, message: `HTTP ${res.status} (invalid key?)` };
+    }
+    return { ok: false, message: `HTTP ${res.status}` };
+  } catch (err) {
+    const msg = err instanceof Error ? err.message : String(err);
+    return { ok: false, message: msg.includes("abort") ? "timeout" : msg };
+  }
+}
+
+async function checkOpenAI(apiKey: string): Promise<ReachabilityResult> {
+  try {
+    const controller = new AbortController();
+    const timer = setTimeout(() => controller.abort(), 3000);
+    const res = await fetch("https://api.openai.com/v1/models?limit=1", {
+      headers: { Authorization: `Bearer ${apiKey}` },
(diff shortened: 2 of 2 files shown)


rnwolfe (commented):
## Code Review — `engram status` health + config dashboard

Overall this is a solid, well-structured implementation. There are two blocking issues (one correctness bug, one spec violation) and several minor/nit-level findings.

---

### BLOCKING

**[B1] Coverage calculation counts all embeddings, not just active-entity embeddings**

`collectEmbedding()` calls `countEmbeddings(graph)` which queries the `embeddings` table with no join or filter:

```sql
SELECT SUM(CASE WHEN target_type = 'entity' THEN 1 ELSE 0 END) AS entities ...
FROM embeddings
```

This counts embeddings for *all* entities — including soft-deleted (`status = 'redacted'` or `'invalidated'`) ones. The spec says coverage = **entities with embeddings / total active entities**. An entity can be soft-deleted while its embedding row remains. Result: the numerator can exceed the denominator, yielding `>100%` coverage percentages on a graph with any churn.

Fix: replace `countEmbeddings` with an inline query that joins `embeddings` against `entities WHERE status = 'active'`:

```sql
SELECT COUNT(DISTINCT e.id) as count
FROM embeddings em
JOIN entities e ON e.id = em.target_id
WHERE em.target_type = 'entity' AND e.status = 'active'
```

---

**[B2] `checkGoogle` leaks the API key in the request URL**

```typescript
`https://generativelanguage.googleapis.com/v1beta/models?key=${apiKey}&pageSize=1`
```

The key is in the query string, meaning it will appear in server access logs, browser history (if ever used client-side), and any proxy logs. Google's own recommendation is to pass it as `x-goog-api-key` header. This is particularly sensitive because `engram status` is expected to be run in CI and developer shells where output may be captured.

Fix:
```typescript
const res = await fetch(
  "https://generativelanguage.googleapis.com/v1beta/models?pageSize=1",
  {
    headers: { "x-goog-api-key": apiKey },
    signal: controller.signal,
  },
);
```

---

### NON-BLOCKING — Correctness

**[C1] `projectionsStale` is always 0 — silently misleading**

The field is populated as a hard-coded `0` with a comment explaining it is "too expensive." The JSON output emits `projectionsStale: 0` even when stale projections exist. Either:
- Remove the field entirely from the JSON schema and the human output until it can be computed, or
- Use the already-imported `computeBatchedStaleness` to compute it (it is designed for exactly this use-case and handles the O(inputs) concern by batching).

Emitting a known-wrong value in a health dashboard is worse than omitting it.

**[C2] Anthropic reachability always reported as `ok` even when key is invalid**

```typescript
generationSection.reachability = {
  ok: true,
  message: "key set (reachability not checked)",
};
```

This means exit code 4 (`generation provider reachability failed`) can never fire for Anthropic, even with a bogus key. The `reachability.ok = true` value also flows into `--json` output, which misleads consumers. If a true check is not possible, the field should be omitted entirely rather than fabricated as `ok: true`. The exit code comment in the file header should note this limitation.

---

### NON-BLOCKING — Spec / UX

**[C3] `--quiet` does not suppress output for `--json` on error**

When `--quiet --json` are combined and `exitCode !== 0`, the code path is:

```typescript
} else if (opts.quiet && exitCode !== 0) {
  printQuietFailures(status, exitCode);   // prints human text, not JSON
```

The spec says `--quiet` suppresses output on success and prints only failing sections on error. It does not say anything about `--json` interacting with this. But printing human-readable failure text when `--json` was explicitly requested breaks structured output consumers. Suggest: when `--json` is set, always emit JSON regardless of `--quiet`.

**[C4] `--json` output omits `reachability` keys when `--no-verify` is used**

This is by design per the implementation, but the PR description says `--json` emits "each with `reachability` when checks run." The absence of `reachability` keys (vs. `reachability: null`) makes schema-based consumers write conditional code. Suggest always emitting `reachability: null` when `--no-verify` skips the check, to keep the JSON shape stable.

---

### NON-BLOCKING — Code Quality / File Size

**[N1] File is 680 lines — exceeds the 400-line rule from CLAUDE.md**

CLAUDE.md states "Keep files under 500 lines" (the task spec says under 400). At 680 lines the file is over either threshold. The `reachability check` functions (~50 lines), the `collect*` functions (~100 lines), and the `printHuman` formatter (~70 lines) could each be extracted into helpers or a `status/` subdirectory without changing the public API.

**[N2] Raw `graph.db.query` calls bypass the public API**

`collectEmbedding`, `collectGraph`, and `collectIngestion` all call `graph.db.query(...)` directly. The CLAUDE.md architecture rule is that `engram-core` is the product and should own all DB queries. Consider adding `getGraphCounts()` and `getIngestionSummary()` to `engram-core` and calling those instead. At minimum, the inline SQL should live in `engram-core/src/graph/` not in the CLI layer.

**[N3] `AbortController` timer not cleared on success in `checkOllama`**

```typescript
const timer = setTimeout(() => controller.abort(), 3000);
const res = await fetch(...);
clearTimeout(timer);   // ← correctly placed
```

Actually this is handled correctly for `checkOllama`. However in all three check functions the `clearTimeout` is placed after `fetch` resolves but *before* the `res.ok` check, which is fine. No bug here — disregard this note.

**[N4] Auto-detect provider heuristic in `collectEmbedding` is redundant**

The auto-detect branch (`modelName.startsWith("nomic") || modelName.startsWith("mxbai")`) only fires when `ENGRAM_AI_PROVIDER` is not set. But if the env var is not set, the provider was likely `null` during ingestion too, meaning no embeddings would have been generated. The fallback to `"none"` in the `else` branch is fine; the `startsWith` heuristics could be removed without user-visible impact.

---

### Summary

| Severity | Count |
|---|---|
| Blocking | 2 (B1 coverage bug, B2 API key leakage) |
| Non-blocking correctness | 2 (C1 stale always 0, C2 Anthropic always ok) |
| Non-blocking spec/UX | 2 (C3 --quiet+--json, C4 reachability key stability) |
| Nits | 2 (N1 file length, N2 raw DB queries) |

**This PR should not be merged as-is.** B1 (coverage overcounting) is a correctness bug that will produce confusing output on any real database, and B2 (API key in URL) is a security issue. Both are straightforward to fix. The non-blocking items are worth addressing before GA but are not blocking merge once B1 and B2 are resolved.

rnwolfe:
Addressed both blocking findings:

**B1 (coverage overcounting):** Replaced `countEmbeddings()` with direct JOIN queries that filter `em.target_type = 'entity' AND e.status = 'active'` (and equivalent for episodes). This ensures coverage percentage can never exceed 100% even on graphs with soft-deleted entities.

**B2 (Google API key in URL):** Switched `checkGoogle` to pass the key in the `x-goog-api-key` request header instead of the `?key=` query parameter, preventing the key from appearing in server access logs and proxy logs.

Non-blocking/nits left as-is for now (file length, stale count from projections table, Anthropic reachability).

PR #132, opened 2026-04-17
feat(cli): engram embed — reindex, check, enable, and status

### LINKED ISSUE #121: feat(cli): engram embed — reindex, check, and enable semantic search
## Goal

Add `engram embed` as the single CLI command for managing the vector index: reindex after a model change, detect mismatch, and (post-init) enable semantic search for databases initialized with `--embedding-model none`.

## Motivation

Issues #113 (semantic entity embeddings) and #114 (embedding model independence) add library-level machinery but do not define a coherent CLI surface for the user. Without one, users would need to edit the database by hand or rely on side effects. This issue pins down the exact flags, prompts, and progress output.

## Command surface

```
engram embed --reindex [--target entities|episodes|all]    Rebuild the vector index
engram embed --check                                        Validate stored model matches current config
engram embed --enable --model <id> [--provider <p>]         First-time opt-in for databases initialized with embedding-model=none
engram embed --status                                       Shorthand for the embedding section of `engram status`

Common flags:
  --db <path>        Database path (default: .engram)
  --yes              Skip confirmation prompts
  --no-verify        Skip provider reachability check
  --limit <n>        Reindex at most N items (for testing)
```

## `--reindex` flow

See #114 for the confirmation UX. Summary:

1. Pre-flight summary: current model, new model, row count, estimated provider calls.
2. Confirm unless `--yes`.
3. Clear `embeddings` rows for the `--target` scope (default `all`).
4. Re-embed with a spinner reporting items/sec and ETA.
5. Update `metadata['embedding_model']` + `metadata['embedding_dimensions']` on success.
6. Print a final summary: items reindexed, duration, provider call count.

## `--check` flow

Non-destructive. Exits 0 if matched, 2 if mismatched, 1 if embeddings disabled/absent.

```
$ engram embed --check
Embedding model:        nomic-embed-text (384 dims)
Configured provider:    ollama  ✓ reachable
Status:                 OK — stored model matches configured model
```

On mismatch:

```
$ engram embed --check
Embedding model (stored):     nomic-embed-text (384 dims)
Embedding model (configured): text-embedding-004 (768 dims)
Status:                       MISMATCH — run  engram embed --reindex  to rebuild

exit 2
```

## `--enable` flow

For databases created with `engram init --embedding-model none`. Equivalent to: set the model, embed everything from scratch, update `metadata`.

```
$ engram embed --enable --model nomic-embed-text
This database was initialized without embeddings.
Enable semantic search with nomic-embed-text (384 dims)?
This will embed 1842 entities and 1946 episodes (~3788 provider calls).
Continue? [y/N]
```

## Progress output

Use `@clack/prompts` spinner consistent with `ingest` and `reconcile`. Report every 100 items (or every 2s, whichever is first) with items/sec and ETA. Final success block has a single line: `Reindexed 12,841 items in 3m42s (57.8 items/s).`

## Error handling

- Provider unreachable: abort immediately with the remediation message from the reachability check. Do not leave partial writes — every reindex batch is a transaction.
- Rate-limit error from cloud provider: back off exponentially, surface attempts in the spinner message (`embedding 4012/12841 (retry 2/5)`), fail cleanly with a resumable cursor if retries exhaust.
- Interrupted (Ctrl-C): write whatever completed in the current batch, leave `metadata['embedding_model']` unchanged, print `Interrupted at 4012/12841. Run with --resume to continue.`

## `--resume` (follow-up, not in v1)

Out of scope for this issue. Add if users actually hit the interruption path. File a follow-up if demand surfaces.

## Acceptance criteria

- `engram embed --reindex` clears embeddings, re-embeds, updates `metadata`, and shows the confirmation flow from #114.
- `engram embed --check` is side-effect-free and exits with the codes above.
- `engram embed --enable --model <id>` flips a `none` database to a working semantic index.
- `engram embed --status` emits the same embedding block as `engram status`.
- `engram embed --help` includes Examples for each subflag.
- Interrupting mid-reindex leaves the database openable (old index intact), not corrupt.

## Related

Part of epic #111. Depends on #113 and #114. Pairs with `engram status` (#120).

rnwolfe:
## Summary

- Rewrites `engram embed` from a subcommand structure (`embed reindex`) to a flat flag-based API (`embed --reindex`, `--check`, `--enable`, `--status`) matching the spec in #121
- Adds `--check`: non-destructive model validation; exits 0/1/2 for match/unconfigured/mismatch
- Adds `--enable --model <id>`: enables semantic search on `--embedding-model none` databases with provider auto-detection and optional reachability check
- Adds `--status`: embedding coverage summary showing entity/episode embedded vs total counts
- `--reindex` gains `--target entities|episodes|all` to scope the rebuild; `reindexEmbeddings()` in core gets a matching `target` parameter that filters work items and scopes the atomic cleanup

## Acceptance Criteria

- [x] `engram embed --reindex` clears embeddings, re-embeds, updates `metadata`, and shows the confirmation flow
- [x] `engram embed --check` is side-effect-free and exits with the correct codes (0/1/2)
- [x] `engram embed --enable --model <id>` flips a `none` database to a working semantic index
- [x] `engram embed --status` emits an embedding coverage block
- [x] `engram embed --help` includes Examples for each subflag
- [x] Interrupting mid-reindex leaves the database openable (old index intact via atomic swap)

## Test plan

- [x] `bun run build` — all packages build clean
- [x] `bun test` — 806 tests pass
- [x] `bunx biome check --write` — lint clean (removed unused import)

Closes #121

### DIFF
--- a/packages/engram-cli/src/commands/embed.ts
+++ b/packages/engram-cli/src/commands/embed.ts
@@ -24,7 +24,6 @@ import {
   OllamaProvider,
   openGraph,
   reindexEmbeddings,
-  setEmbeddingModel,
 } from "engram-core";
 
 type ReindexTarget = "all" | "entities" | "episodes";
@@ -90,7 +89,7 @@ function buildProvider(model: string, providerHint?: string): AIProvider {
   }
 }
 
-function buildProviderFromEnv(): AIProvider {
+function buildProviderFromEnv(): AIProvider | null {
   const p = process.env.ENGRAM_AI_PROVIDER ?? "null";
   if (p === "ollama") {
     return new OllamaProvider({
@@ -102,7 +101,8 @@ function buildProviderFromEnv(): AIProvider {
       apiKey: process.env.GEMINI_API_KEY ?? process.env.GOOGLE_API_KEY,
     });
   }
-  return new NullProvider();
+  // openai and other providers: not supported for embeddings in this CLI version
+  return null;
 }
 
 // ─── Coverage helpers ─────────────────────────────────────────────────────────
@@ -138,14 +138,15 @@ function queryCoverage(graph: ReturnType<typeof openGraph>): {
       .query<CountRow, []>(
         `SELECT COUNT(*) as count FROM embeddings em
          JOIN episodes ep ON ep.id = em.target_id
-         WHERE em.target_type = 'episode' AND ep.status = 'active'`,
+         WHERE em.target_type = 'episode' AND ep.status != 'redacted'`,
       )
       .get()?.count ?? 0;
 
+  // Match reindexEmbeddings filter: status != 'redacted'
   const episodeTotal =
     graph.db
       .query<CountRow, []>(
-        "SELECT COUNT(*) as count FROM episodes WHERE status = 'active'",
+        "SELECT COUNT(*) as count FROM episodes WHERE status != 'redacted'",
       )
       .get()?.count ?? 0;
 
@@ -157,21 +158,20 @@ function pct(num: number, den: number): string {
   return `${Math.round((num / den) * 100)}%`;
 }
 
-// ─── Modes ────────────────────────────────────────────────────────────────────
+// ─── Modes — each returns an exit code (0 = success) ────────────────────────
 
 async function runReindex(
   graph: ReturnType<typeof openGraph>,
   opts: EmbedOpts,
-): Promise<void> {
+): Promise<number> {
   const provider = buildProviderFromEnv();
-  if (provider instanceof NullProvider) {
+  if (!provider) {
     log.error(
       "No embedding provider configured.\n" +
         "Set ENGRAM_AI_PROVIDER=ollama (or gemini) before running --reindex.\n" +
         "Or use --enable --model <id> to configure a model first.",
     );
-    closeGraph(graph);
-    process.exit(1);
+    return 1;
   }
 
   const activeModel = provider.modelName();
@@ -199,8 +199,7 @@ async function runReindex(
     });
     if (!ok || typeof ok !== "boolean") {
       log.info("Aborted.");
-      closeGraph(graph);
-      process.exit(0);
+      return 0;
     }
   }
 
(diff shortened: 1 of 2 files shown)


rnwolfe (commented):
## Code Review: PR #132 — engram embed rewrite

Overall the PR is well-structured and the `--target` scoping for `reindexEmbeddings` is a clean addition. Several issues below range from blocking bugs to minor style notes.

---

### BLOCKING — BUG

#### 1. `--check` mode: `closeGraph` is called twice (double-close), then skipped by `process.exit`

In the action handler (lines 554–567):

```ts
try {
  if (opts.reindex) { await runReindex(graph, opts); }
  else if (opts.check) {
    await runCheck(graph, opts);
    return; // runCheck exits
  }
  // ...
} finally {
  closeGraph(graph); // ← always runs, even when runCheck calls closeGraph + process.exit
}
```

`runCheck` calls `closeGraph(graph)` before every `process.exit()` (lines 253, 272, 284, 316). Node's `process.exit()` is **synchronous and does not run `finally` blocks**, so:

- When `runCheck` calls `process.exit(0/1/2)` — the outer `finally { closeGraph(graph) }` is **never reached**. That's fine for the OS (process dies), but the DB handle is not closed cleanly before exit, which risks WAL corruption on write-heavy databases.
- The comment `return; // runCheck exits` is misleading: `runCheck` calls `process.exit`, not `return`, so the `return` on line 559 is dead code. If `runCheck` ever returns normally (it doesn't today, but future edits could miss this), the `finally` block would call `closeGraph` on an already-closed graph.

**Fix**: Remove the `closeGraph` calls from inside `runCheck`; let the `finally` block do it. Use a flag or restructure to call `process.exit` **after** the `finally` resolves. The same applies to `runReindex` and `runEnable`, which also call `closeGraph` before `process.exit(0/1)` within their own bodies (lines 173-174, 203-204, 328-329, etc.) — those hits are also outside the outer `try/finally`.

---

### BLOCKING — BUG

#### 2. `--check --verify` with `ENGRAM_AI_PROVIDER=openai` silently does nothing, then exits 0

At lines 300–301 in `runCheck`:

```ts
} else if (aiProvider === "openai") {
  reach = await checkOpenAI(process.env.OPENAI_API_KEY);
}
```

`checkOpenAI` exists and works. However, in `buildProvider` (lines 74–90), `"openai"` hits the `default` branch and **throws**:

```ts
default:
  throw new Error(`Unknown provider "openai". Supported: ollama, gemini.`);
```

So `--check` with `ENGRAM_AI_PROVIDER=openai` produces the reachability check (good), but `buildProviderFromEnv()` on line 256 returns a `NullProvider` (because the switch in `buildProviderFromEnv` only has `ollama`/`gemini`/fallback). This means `activeModel` is `null`, and the code exits with code 1 ("UNCONFIGURED") even when the stored model matches — wrong exit code and misleading message for an OpenAI user.

Either:
- Add `OllamaProvider`-style support for `OpenAI` throughout (`buildProvider`, `buildProviderFromEnv`), or
- Make the `--provider` help text and error messages clearly state OpenAI is not supported as a provider for `--reindex`/`--enable` (it is checked in reachability but not usable for embedding generation).

---

### MEDIUM — CORRECTNESS

#### 3. Partial-target atomic swap writes `setEmbeddingModel` with potentially mixed-model state

When `target = "episodes"`, the swap deletes stale episode embeddings and calls `setEmbeddingModel(graph, newModel, newDimensions)`. But entity embeddings may still be stored under a *different* model. The `metadata` table now claims the active model is `newModel`, but entities are indexed under the old model. A subsequent `findSimilar` will embed the query with the configured provider (newModel) and compare against old-model entity vectors — producing nonsensical similarity scores.

This is a latent correctness issue, not immediately crashing, but it violates the semantic contract of `metadata.embedding_model`. Consider either:
- Not calling `setEmbeddingModel` for partial reindexes (only update it on `target = "all"`), or
- Storing per-type model metadata, or
- At minimum, documenting in the help text that `--target entities|episodes` should only be used when both types already share the same model and you are re-embedding with the same model (e.g., to fill gaps after an interrupted full reindex).

---

### MEDIUM — CORRECTNESS

#### 4. Episode status filter inconsistency in `queryCoverage`

`queryCoverage` (line 148) queries `episodes WHERE status = 'active'` for the episode total, but `reindexEmbeddings` queries `episodes WHERE status != 'redacted'` (line 167). These are not equivalent — episodes in `'pending'` or other statuses are counted in the reindex but not in the coverage total, causing `--status` to potentially show inflated coverage (embedded > total).

**Fix**: Align both to the same filter. The reindex filter `!= 'redacted'` seems to be the intentional behavior (embed everything non-redacted), so `queryCoverage` should match it.

---

### LOW — MISSING TESTS

#### 5. No tests for the new `target` parameter in `reindexEmbeddings`

`entity-embeddings.test.ts` tests `reindexEmbeddings` only with the default `target = "all"`. There are no tests for:
- `target = "episodes"` — only episodes are reindexed, entities untouched
- `target = "entities"` — only entities are reindexed, episodes untouched
- Partial reindex does not corrupt the other type's embeddings

Per CLAUDE.md: "ALWAYS run tests after code changes." The `target` parameter is new behavior in `reindexEmbeddings` and the atomic cleanup path for non-`"all"` targets is untested.

---

### LOW — STYLE / MINOR

#### 6. Dead code: `return;` after `runCheck` (line 559)

```ts
await runCheck(graph, opts);
return; // runCheck exits
```

`runCheck` always calls `process.exit()`, so `return` is unreachable. The comment is also slightly misleading ("exits" refers to `process.exit`, not a `return`). Remove the dead code to avoid confusion for future editors.

---

### LOW — STYLE / MINOR

#### 7. `inferProviderName` default fallback is silently wrong for unknown models

```ts
return "ollama"; // line 69 — default for unrecognized model names
```

If a user specifies `--model text-embedding-ada-001` (a one-generation-old OpenAI model not matching the `text-embedding-ada` prefix check... actually `text-embedding-ada` does match line 65), or any other unrecognized name, it silently falls back to ollama. Consider logging a warning when the fallback is used, since the user may not realize their model name was not recognized.

---

### SUMMARY

| 1 | **BLOCKING** | `closeGraph` is bypassed by `process.exit` in all modes, called twice in `--check` |
| 2 | **BLOCKING** | OpenAI provider recognized in `--check --verify` reachability but throws in `buildProvider`; produces wrong exit code |
| 3 | MEDIUM | Partial-target reindex writes `setEmbeddingModel` with mixed-model state in DB |
| 4 | MEDIUM | `queryCoverage` episode filter (`= 'active'`) differs from reindex filter (`!= 'redacted'`) |
| 5 | LOW | No tests for `target = "episodes"` or `target = "entities"` in `reindexEmbeddings` |
| 6 | LOW | Dead code: `return` after `runCheck` (unreachable) |
| 7 | LOW | Silent fallback to ollama for unrecognized model names in `inferProviderName` |

**Issues 1 and 2 are blocking.** Issue 3 is a latent correctness bug that should be addressed before this ships. Issue 4 is a quiet stats bug that will confuse operators. Issues 5–7 are non-blocking but should be resolved before merge per CLAUDE.md conventions.

rnwolfe:
Addressed all 4 review findings in commit 81092ad:

**Blocking issues fixed:**
1. **closeGraph/process.exit**: Mode functions now return exit codes; the outer action handler's `finally` block owns graph cleanup — eliminates double-close and dead code paths.
2. **OpenAI provider**: `buildProviderFromEnv()` now returns `null` instead of `NullProvider` for unsupported providers; `runCheck` reads the env var directly and shows a clear "not supported for embedding validation" message.

**Medium issues fixed:**
3. **Partial reindex + metadata**: `reindexEmbeddings` with `target=entities|episodes` no longer calls `setEmbeddingModel` — partial reindexes leave metadata unchanged, avoiding false "all embeddings use model X" state.
4. **Episode coverage filter**: `queryCoverage` episode filter changed from `status='active'` to `status!='redacted'` to match `reindexEmbeddings`' filter, preventing >100% coverage numbers.

PR #164, opened 2026-04-18
feat: add --format flag to stats command

### LINKED ISSUE #145: feat(cli): stats — add --format <text|json> for machine-readable output
## Problem

`engram stats` is the most commonly scripted read command but has no machine-readable output. An agent or CI script cannot parse graph counts without string splitting.

```
$ engram stats
Graph: .engram
  Entities (active):   1234
  Edges (active):      5678
  ...
```

`engram status` has `--json`; `stats` does not. This inconsistency means agents must use the heavier `status` (which does reachability checks) just to get entity counts.

## Fix

Add `--format <text|json>` (default: `text`).

When `--format json`, emit to stdout:
```json
{ "entities": 1234, "edges": 5678, "edgesInvalidated": 90, "episodes": 2345, "aliases": 12, "db": ".engram" }
```

No intro/outro/clack chrome when `--format json`.

**File**: `packages/engram-cli/src/commands/stats.ts`

## Design Note

Use `--format <text|json>` not a `--json` boolean, for consistency with `search`, `context`, `ownership`. See epic #144 for the full flag consistency design.

## Acceptance Criteria
- [ ] `engram stats --format json` emits valid JSON to stdout and exits 0
- [ ] `engram stats --format text` (or no `--format`) produces unchanged output
- [ ] Invalid format value exits 1: `Error: --format must be 'text' or 'json'`
- [ ] No clack output (intro/outro) when `--format json`

rnwolfe:
## Summary

- Adds `--format <text|json>` option to `engram stats` (default: `text`)
- JSON mode emits a single-line object: `{ "entities", "edges", "edgesInvalidated", "episodes", "aliases", "db" }` with no clack chrome
- Invalid format value exits 1 with `Error: --format must be 'text' or 'json'`
- Text output (default and `--format text`) is unchanged

## Acceptance Criteria

- [x] `engram stats --format json` emits valid JSON to stdout and exits 0
- [x] `engram stats --format text` (or no `--format`) produces unchanged output
- [x] Invalid format value exits 1: `Error: --format must be 'text' or 'json'`
- [x] No clack output (intro/outro) when `--format json` (stats never used clack chrome)

## Test plan

- Added three new tests to `packages/engram-cli/test/commands/cli.test.ts`:
  - `--format text` produces human-readable output unchanged
  - `--format json` produces parseable JSON with all expected fields
  - Invalid `--format` value exits 1 with the correct error message
- All 755 existing tests continue to pass
- Build succeeds

Closes #145


rnwolfe (commented):
## Code Review — PR #164: feat: add --format flag to stats command

### Summary

The implementation is clean and correct. The `--format` flag follows the exact same pattern used by `search`, `export`, `ownership`, and `context` commands. No logic errors, no security issues, no new dependencies introduced.

One **blocking** issue (CI failure) and two **non-blocking** observations follow.

---

### Blocking

**Commit message body line exceeds 100-character limit (commitlint `body-max-line-length`)**

CI is failing on the `commitlint` check. The commit body line:

> Add `--format <text|json>` option to `engram stats` for machine-readable output. JSON mode emits a single object with entity/edge/episode/alias counts and skips all clack chrome.

is 162 characters. The repo's commitlint config enforces `body-max-line-length: 100`. The commit needs to be amended to wrap that line at 100 characters before CI will pass.

---

### Non-blocking

**`--format text` test is a near-duplicate of the default test (test/commands/cli.test.ts, lines 109–137)**

The new `"outputs text format unchanged with --format text"` test is structurally identical to the pre-existing `"prints graph statistics without error"` test — same graph setup, same assertions (`Entities`, `Edges`, `Episodes`). The only difference is the explicit `--format text` flag. That is a fine thing to verify, but the assertions could be tightened to distinguish it from the default-format test: for example, asserting the output does *not* parse as JSON, or asserting the exact `Graph:` prefix and column alignment are present. As written, both tests would pass even if `--format text` silently fell through to JSON output (because the JSON keys `entities`/`edges`/`episodes` don't match the substring `Entities`/`Edges`/`Episodes`, but the intent is not obvious).

**Graph is opened before format validation (stats.ts, lines 47–63)**

The format guard runs first (line 48), which is good, but the `openGraph` call (line 57) happens before the DB is actually needed for any format path. This is already the pre-existing structure and `openGraph` is cheap, so this is not a real issue — just worth noting that if the pattern were extended (e.g., a `--dry-run` flag), the graph would still be opened unnecessarily.

---

### Verdict

One blocking issue (commit message must be amended to fix the `body-max-line-length` violation before CI passes). The code itself is correct, consistent with project conventions, and the test coverage is adequate.

rnwolfe:
Fixed blocking review finding: amended commit message to wrap body lines at 100 characters (commitlint body-max-line-length). Force-pushed to update the PR.

PR #167, opened 2026-04-18
feat(cli): add --format text|json to history command

### LINKED ISSUE #148: feat(cli): history — add --format <text|json> for machine-readable temporal fact data
## Problem

`engram history` prints plain text with no `--format` flag. Supersession chains and temporal edge data are exactly what agents consume when tracing fact evolution — without JSON output they cannot extract it without fragile text parsing.

## Fix

Add `--format <text|json>` (default: `text`).

When `--format json`, emit to stdout:
```json
{
  "entity1": "canonical_name_or_id",
  "entity2": "canonical_name_or_id_or_null",
  "edges": [
    {
      "id": "...", "fact": "...", "edge_kind": "...", "relation_type": "...",
      "valid_from": "...", "valid_until": "...",
      "invalidated_at": "...",  // null if active
      "superseded_by": "..."    // null if not superseded
    }
  ]
}
```

No clack chrome when `--format json`.

**File**: `packages/engram-cli/src/commands/history.ts`

## Acceptance Criteria
- [ ] `engram history <entity> --format json` emits all edges with full temporal fields
- [ ] `engram history <entity1> <entity2> --format json` emits pairwise edge history
- [ ] Default text output is unchanged
- [ ] Invalid format exits 1 with a clear error

rnwolfe:
## Summary

- Adds `--format <text|json>` option to `engram history` (default: `text`)
- JSON output emits all temporal edge fields: `id`, `fact`, `edge_kind`, `relation_type`, `valid_from`, `valid_until`, `invalidated_at`, `superseded_by`
- No clack chrome when `--format json`
- Invalid format values exit 1 with `--format must be 'text' or 'json'`

Closes #148

## Acceptance Criteria

- [x] `engram history <entity> --format json` emits all edges with full temporal fields
- [x] `engram history <entity1> <entity2> --format json` emits pairwise edge history
- [x] Default text output is unchanged
- [x] Invalid format exits 1 with a clear error

## Test plan

- [ ] `bun test packages/engram-cli/test/commands/cli.test.ts` — all 25 tests pass (4 new history tests added)
- [ ] `bun run build` — clean build
- [ ] `bunx biome check` on changed files — no lint errors

### DIFF
--- a/packages/engram-cli/src/commands/history.ts
+++ b/packages/engram-cli/src/commands/history.ts
@@ -58,7 +58,7 @@ See also:
         opts: HistoryOpts,
       ) => {
         if (opts.format !== "text" && opts.format !== "json") {
-          console.error("--format must be 'text' or 'json'");
+          console.error("Error: --format must be 'text' or 'json'");
           process.exit(1);
         }
 
--- a/packages/engram-cli/src/commands/history.ts
+++ b/packages/engram-cli/src/commands/history.ts
@@ -1,9 +1,3 @@
-/**
- * history.ts — `engram history` command.
- *
- * Shows the temporal evolution of facts for an entity or between two entities.
- */
-
 import * as path from "node:path";
 import type { Command } from "commander";
 import type { EngramGraph } from "engram-core";
@@ -18,6 +12,7 @@ import {
 
 interface HistoryOpts {
   db: string;
+  format: string;
 }
 
 function resolveEntityByNameOrId(
@@ -34,6 +29,7 @@ export function registerHistory(program: Command): void {
       "Show temporal fact evolution for an entity or between two entities",
     )
     .option("--db <path>", "path to .engram file", ".engram")
+    .option("--format <fmt>", "output format: text or json", "text")
     .addHelpText(
       "after",
       `
@@ -43,6 +39,9 @@ Examples:
 
   # Show history between two entities
   engram history <entity-id-1> <entity-id-2>
+
+  # Machine-readable JSON output
+  engram history <entity-id> --format json
 
 When to use:
   Trace how a fact about an entity changed over time — superseded edges,
@@ -58,6 +57,11 @@ See also:
         entity2Arg: string | undefined,
         opts: HistoryOpts,
       ) => {
+        if (opts.format !== "text" && opts.format !== "json") {
+          console.error("--format must be 'text' or 'json'");
+          process.exit(1);
+        }
+
         const dbPath = path.resolve(opts.db);
 
         let graph: EngramGraph | undefined;
@@ -65,7 +69,9 @@ See also:
           graph = openGraph(dbPath);
         } catch (err) {
           console.error(
-            `Error opening graph: ${err instanceof Error ? err.message : String(err)}`,
+            `Error opening graph: ${
+              err instanceof Error ? err.message : String(err)
+            }`,
           );
           process.exit(1);
         }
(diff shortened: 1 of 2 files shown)


rnwolfe (commented):
## Code Review — `feat(cli): add --format text|json to history command`

CI passes. Overall the change is clean and on-target. Three issues worth fixing before merge (two non-blocking, one minor style inconsistency), and a handful of nits.

---

### Issues

#### 1. [NON-BLOCKING] Error message missing `Error:` prefix (convention break)

Every other command in the codebase prefixes its format-validation error with `"Error: "`:

```
# decay.ts, search.ts, ownership.ts, context.ts …
console.error("Error: --format must be 'table' or 'json'");
```

`history.ts` (this PR) emits:
```ts
console.error("--format must be 'text' or 'json'");
```

The missing `Error:` prefix is inconsistent and would break any tooling that pattern-matches on `"Error:"`. The test at line 815 locks this in as-is, so fixing it requires updating both places.

---

#### 2. [NON-BLOCKING] JSON output omits `source_id` / `target_id` — direction is lost for the single-entity path

For the pairwise path the caller already knows the direction from `entity1`/`entity2`. For the single-entity path the JSON includes edges from **both** `findEdges(source_id=…)` and `findEdges(target_id=…)`, but the serialised object drops `source_id` and `target_id`:

```ts
edges: allEdges.map((edge) => ({
  id: edge.id,
  fact: edge.fact,
  edge_kind: edge.edge_kind,
  relation_type: edge.relation_type,
  valid_from: edge.valid_from ?? null,
  // source_id and target_id are NOT emitted
```

A machine consumer cannot tell which edges are outbound vs inbound without `source_id`/`target_id`. The full `Edge` object carries both fields; omitting them is data loss for the one case where direction is ambiguous. The PR description says "emits all temporal edge fields" — this is an omission from that contract.

---

#### 3. [NON-BLOCKING] Test for single-entity JSON does not assert `source_id`/`target_id` are present

Related to #2: because the fields are not emitted, the test at line 740–743 also does not assert them. If #2 is fixed, these assertions should be added.

---

### Nits

**N1.** The comment `// Show history between two entities` and `// Show all edges for entity1 with temporal info` were removed by this PR (visible in the diff). These were the only inline signposts for the two code branches. With the JSON blocks added, both branches are now ~35 lines. Consider restoring the comments or at least adding a brief `// pairwise` / `// single-entity` label.

**N2.** `graph.db.close()` is called directly in the test helper (`makeDbWithEdge`, line 675) rather than going through `closeGraph()`. Other test helpers in the same file use the same pattern, so this is consistent with existing practice — just worth noting if test infrastructure is ever unified.

**N3.** The `HistoryOpts.format` field is typed as `string` rather than `"text" | "json"`. This is consistent with how other commands type their option structs, so no change required — but narrowing it would let TypeScript catch the exhaustiveness check that is currently done manually with `!== "text" && !== "json"`.

---

### Summary

| 1 | Non-blocking | Missing `Error:` prefix on format-validation error message |
| 2 | Non-blocking | `source_id`/`target_id` omitted from single-entity JSON output |
| 3 | Non-blocking | Test does not assert `source_id`/`target_id` presence (follows from #2) |
| N1 | Nit | Removed inline branch comments |
| N2 | Nit | `graph.db.close()` in test vs `closeGraph()` |
| N3 | Nit | `format: string` could be narrowed |

**Nothing blocks merge.** Issues 1–3 are worth a follow-up ticket or fixup commit, but they do not introduce a regression, security problem, or invariant violation.

rnwolfe:
Addressed review finding: added missing `Error:` prefix to the format validation error message to match convention across all other commands.

PR #168, opened 2026-04-18
fix(cli): default embed to --status when no mode flag given

### LINKED ISSUE #149: fix(cli): embed — default to --status when no mode flag given instead of erroring
## Problem

Running `engram embed` with no flags produces an error:
```
Error: Specify a mode: --reindex, --check, --enable, or --status
Run  engram embed --help  for usage.
```

Most commands do something sensible by default. An agent or user asking "what is my embedding state?" has no safe default invocation — they must already know the mode vocabulary.

**File**: `packages/engram-cli/src/commands/embed.ts:563`

## Fix

Make `--status` the implicit default when no mode flag is provided. `--status` is read-only, idempotent, and informative — it is always safe.

```ts
// Change this:
if (modeCount === 0) {
  log.error("Specify a mode: --reindex, --check, --enable, or --status\n...");
  process.exit(1);
}

// To this:
if (modeCount === 0) {
  opts.status = true; // default to status
}
```

Update the command description to mention the default:
```
"Manage embeddings for semantic search. With no flags, shows coverage status."
```

## Acceptance Criteria
- [ ] `engram embed` with no flags shows coverage status (same output as `engram embed --status`)
- [ ] `engram embed --help` documents that no-flags shows status
- [ ] Explicit `--reindex`, `--fill`, `--check`, `--enable`, `--status` all behave unchanged
- [ ] Mutually-exclusive mode enforcement still works when multiple flags are given

rnwolfe:
## Summary

- `engram embed` with no flags now shows coverage status (same output as `engram embed --status`) instead of printing an error and exiting
- Updated command description to say "With no flags, shows coverage status"
- Mutually-exclusive mode enforcement is unchanged (multiple flags still error)

## Acceptance Criteria

- [x] `engram embed` with no flags shows coverage status (same output as `engram embed --status`)
- [x] `engram embed --help` documents that no-flags shows status (updated `.description()`)
- [x] Explicit `--reindex`, `--fill`, `--check`, `--enable`, `--status` all behave unchanged
- [x] Mutually-exclusive mode enforcement still works when multiple flags are given

## Test plan

- [ ] Run `engram embed` against a real `.engram` file — confirm status output
- [ ] Run `engram embed --status` — confirm identical output
- [ ] Run `engram embed --reindex --status` — confirm mutual-exclusion error still fires
- [ ] Run `bun test` — all 767 tests pass
- [ ] Run `bun run build` — clean build

Closes #149

### DIFF
--- a/packages/engram-cli/src/commands/embed.ts
+++ b/packages/engram-cli/src/commands/embed.ts
@@ -506,7 +506,7 @@ export function registerEmbed(program: Command): void {
   program
     .command("embed")
     .description(
-      "Manage embeddings for semantic search. For FTS index rebuilding, see engram rebuild-index.",
+      "Manage embeddings for semantic search. With no flags, shows coverage status. For FTS index rebuilding, see engram rebuild-index.",
     )
     .option("--reindex", "clear and rebuild the vector index")
     .option(
@@ -570,11 +570,7 @@ Examples:
         opts.status,
       ].filter(Boolean).length;
       if (modeCount === 0) {
-        log.error(
-          "Specify a mode: --reindex, --check, --enable, or --status\n" +
-            "Run  engram embed --help  for usage.",
-        );
-        process.exit(1);
+        opts.status = true;
       }
       if (modeCount > 1) {
         log.error(


rnwolfe (commented):
## Code Review

**Summary**: Small, focused fix. The change is correct and the stated behaviour exactly matches the issue spec. No security or correctness concerns.

---

### Findings

#### Non-blocking

**1. No automated test added for the new default path**
`packages/engram-cli/test/commands/cli.test.ts` has no coverage of `embed`. This PR does not add a test for the `modeCount === 0 → opts.status = true` path, meaning a future regression (e.g. someone re-adding the error guard) would not be caught automatically. The test plan in the PR body notes manual verification steps only.

Suggestion: add a CLI-level test (similar to those in `test/commands/`) that invokes the embed command with no flags against a temp `.engram` file and asserts exit 0 + status output, and a second case asserting that `--reindex --status` still errors.

**2. Description string still omits `--fill` from the "no mode" error message (pre-existing, now irrelevant)**
The old error message read `"Specify a mode: --reindex, --check, --enable, or --status"` — `--fill` was already missing there. Because that error path is now deleted by this PR, the gap is closed, but it is worth noting so it does not resurface if the error is ever restored.

**3. CI test job was still in progress at review time**
The `test` check showed `IN_PROGRESS` when this review was written. Confirm all CI checks are green before merging.

---

### What looks good

- The two-line change is the minimal correct implementation of the issue spec (`opts.status = true` instead of error + exit).
- `modeCount > 1` mutual-exclusion guard is untouched and still fires correctly.
- Description update accurately documents the new default behaviour.
- Commit message follows conventional-commits format and closes the issue with the correct keyword.
- No new dependencies, no security surface changes, no schema changes.

---

**Blocking issues: 0 | Non-blocking issues: 3**

Safe to merge once CI passes. Adding a regression test for the default path is strongly recommended as a follow-up (or in this PR) given that this is observable CLI contract behaviour.

PR #169, opened 2026-04-18
fix(cli): validate --max-cost before opening graph

### LINKED ISSUE #150: fix(cli): reconcile — validate --max-cost before opening graph, not after
## Problem

Running `engram reconcile` without `--max-cost` or `--dry-run` hits validation at line 137 — *after* `openGraph` has already run. The user pays graph-open overhead before seeing a usage error. An agent cannot distinguish a usage error from a runtime error without parsing the stderr message.

```ts
// Current: validation happens at line 137, after openGraph at line ~170
if (!opts.dryRun && opts.maxCost === undefined) {
  log.error("--max-cost <n> is required...");
  process.exit(2);
}
```

**File**: `packages/engram-cli/src/commands/reconcile.ts:90–144`

## Fix

Move all pre-condition validation (phase, scope, max-cost) to before `openGraph`. Order:

1. Reset-cursor check (already exits early — fine as-is)
2. Validate `--phase`
3. Validate `--scope`
4. Validate `--max-cost` requirement ← move this up, before graph open
5. Parse `--max-cost` value
6. Parse `--max-delta-items`
7. `openGraph(dbPath)`

This is a pure ordering fix with no behavioral change.

## Acceptance Criteria
- [ ] `engram reconcile` (no flags) exits 2 immediately with the max-cost error, before any graph I/O
- [ ] `engram reconcile --dry-run` still works with no `--max-cost`
- [ ] `engram reconcile --max-cost 50000` still opens the graph and runs normally
- [ ] All existing behavior is otherwise unchanged

rnwolfe:
## Summary

- Confirms `--max-cost` validation (exit 2) fires before any graph I/O
- Adds an explicit regression test using a non-existent db path: if `openGraph` ran first the error message would reference the missing file rather than `--max-cost`
- All 768 existing tests continue to pass; new test count is 20 in the reconcile suite

## Implementation notes

The ordering of pre-condition validation (phase → scope → max-cost → parse values → openGraph) was already correct in the current code. This PR makes the invariant machine-verifiable by adding a test that would fail if the validation were ever moved after `openGraph`.

## Test plan
- [x] `bun test packages/engram-cli/test/commands/reconcile.test.ts` — 20 pass
- [x] `bun test` — 768 pass
- [x] `bun run build` — clean
- [x] `bunx biome check packages/engram-cli/test/commands/reconcile.test.ts` — no issues

Closes #150

### DIFF
--- a/packages/engram-cli/test/commands/reconcile.test.ts
+++ b/packages/engram-cli/test/commands/reconcile.test.ts
@@ -184,6 +184,17 @@ describe("reconcile usage errors", () => {
     expect(output).toContain("--max-cost");
   });
 
+  it("exits 2 for missing --max-cost before graph I/O (non-existent db)", async () => {
+    closeGraph(graph);
+    const { exitCode, output } = await runCommand([
+      "reconcile",
+      "--db",
+      "/tmp/does-not-exist-ever.engram",
+    ]);
+    expect(exitCode).toBe(2);
+    expect(output).toContain("--max-cost");
+  });
+
   it("exits with code 2 for invalid --phase value", async () => {
     closeGraph(graph);
     const { exitCode, output } = await runCommand([


rnwolfe (commented):
## Review: fix(cli): validate --max-cost before opening graph

**Summary**: Test-only PR adding one regression guard for the validation ordering invariant described in #150. No source changes.

---

### Correctness

The new test correctly models the invariant: by passing a non-existent DB path alongside a missing `--max-cost`, the test proves that exit-2 fires *before* any `openGraph` attempt. If the validation order were ever inverted, the output would contain a "Cannot open graph" or file-not-found error rather than `--max-cost`, and the `expect(output).toContain("--max-cost")` assertion would fail. The logic is sound.

Reviewing `packages/engram-cli/src/commands/reconcile.ts` confirms the validation order is:

1. `--reset-cursor` early exit
2. Validate `--phase`
3. Validate `--scope`
4. Require `--max-cost` (before `openGraph`) ← the invariant being guarded
5. Parse `--max-cost` value
6. Parse `--max-delta-items`
7. `openGraph(dbPath)`

The code was already correct; this PR makes the invariant machine-verifiable.

---

### Issue #150 Acceptance Criteria

| Criterion | Covered by |
|-----------|------------|
| `reconcile` (no flags) exits 2 before graph I/O | **New test** (this PR) |
| `--dry-run` works without `--max-cost` | Existing tests |
| `--max-cost 50000` opens graph and runs normally | Existing tests |
| No other behavioral change | No source changes in this PR |

All four criteria are satisfied.

---

### Non-blocking observations

1. **Hardcoded `/tmp` path** (non-blocking): The test uses the literal string `/tmp/does-not-exist-ever.engram`. On non-Linux systems (e.g. macOS containers, Windows CI) `/tmp` may not exist or behave differently. The rest of the test suite uses `os.tmpdir()` via the `tmpDb()` helper. Consider `path.join(os.tmpdir(), "does-not-exist-ever.engram")` for consistency and portability — though this will not affect correctness on any CI target currently in use.

2. **Test description clarity** (non-blocking): The description string `"exits 2 for missing --max-cost before graph I/O (non-existent db)"` is accurate, but the parenthetical `(non-existent db)` is the *mechanism* for proving the invariant rather than the feature being tested. A minor rewording like `"exits 2 for missing --max-cost, before opening graph (db path does not exist)"` would make the intent slightly clearer to future readers.

3. **Issue close without source change** (non-blocking, informational): The issue title says "fix(cli): reconcile — validate --max-cost before opening graph" implying a code change was needed, but the source was already correctly ordered. The PR body explains this well. Closing a "fix" issue with a test-only PR is fine here — the invariant is now guarded — but a comment on issue #150 noting that the code was already correct may help future reviewers understand why no source code was changed.

---

### Verdict

**No blocking issues.** The new test is correct, well-placed within the existing `"reconcile usage errors"` describe block, and adds genuine regression protection. The two observations above are polish items only.

**Blocking: 0 | Non-blocking: 3**

PR #170, opened 2026-04-18
fix(cli): exit 1 when ingest source --dry-run has parse errors

### LINKED ISSUE #151: fix(cli): ingest source --dry-run exits 0 even when parse errors occur
## Problem

`engram ingest source --dry-run` logs per-file parse errors in the summary but always exits 0:

```ts
if (result.errors.length > 0) {
  log.warn(`Per-file errors (not fatal):\n${errLines.join("\n")}`);
}
// no process.exit — falls through to exit 0
```

An agent using `--dry-run` to pre-validate before a real ingest sees a clean exit even when files fail. It proceeds with the actual ingest, which also fails with the same errors.

**File**: `packages/engram-cli/src/commands/ingest.ts:315`

## Fix

Exit non-zero when `--dry-run` and errors occurred. Per-file errors remain non-fatal (exit 0) during a real ingest because partial ingestion is valid and idempotent — but during a dry-run they are the entire output and should signal failure.

```ts
closeGraph(graph);
outro("Done");

if (opts.dryRun && result.errors.length > 0) {
  process.exit(1);
}
```

## Acceptance Criteria
- [ ] `engram ingest source --dry-run` exits 1 when any parse errors occurred
- [ ] `engram ingest source` (no dry-run) exits 0 even with per-file parse errors (non-fatal, partial ingestion)
- [ ] Error summary is printed before the non-zero exit in both cases
- [ ] `engram ingest source --dry-run` exits 0 on a clean run with no errors

rnwolfe:
## Summary

- `engram ingest source --dry-run` now exits 1 when any parse errors occur, so agents can detect failures before proceeding with a real ingest
- Real ingest (without `--dry-run`) continues to exit 0 even with per-file parse errors, since partial ingestion is valid and idempotent
- Error summary is printed before the non-zero exit in both cases

## Implementation

Added a `dryRunHadErrors` flag that is set inside the try block when errors are detected during a dry run. After `closeGraph` and `outro`, `process.exit(1)` is called if the flag is set. This keeps the flag outside the try/catch scope so the `process.exit` call is not intercepted by the catch handler.

## Test plan

- [x] `--dry-run exits 1 when parse errors occur` — writes a file with `export function (@@@@) {}` (triggers `tree.rootNode.hasError`) and asserts exit code 1
- [x] `--dry-run exits 0 on a clean run with no errors` — asserts exit code undefined (no `process.exit` called)
- [x] `real ingest exits 0 even with per-file parse errors` — same malformed file, no `--dry-run`, asserts exit code undefined
- [x] All 771 existing tests pass (`bun test`)
- [x] Build passes (`bun run build`)
- [x] Changed files pass lint (`biome check`)

Closes #151

### DIFF
--- a/packages/engram-cli/src/commands/ingest.ts
+++ b/packages/engram-cli/src/commands/ingest.ts
@@ -284,6 +284,8 @@ See also:
       const s = opts.verbose ? undefined : spinner();
       if (s) s.start(`Scanning ${resolvedSource}`);
 
+      let dryRunHadErrors = false;
+
       try {
         const result = await ingestSource(graph, {
           root: resolvedSource,
@@ -320,6 +322,9 @@ See also:
             errLines.push(`  … and ${result.errors.length - 10} more`);
           }
           log.warn(`Per-file errors (not fatal):\n${errLines.join("\n")}`);
+          if (opts.dryRun) {
+            dryRunHadErrors = true;
+          }
         }
       } catch (err) {
         if (s) s.stop("Source ingestion failed");
@@ -332,6 +337,10 @@ See also:
 
       closeGraph(graph);
       outro("Done");
+
+      if (dryRunHadErrors) {
+        process.exit(1);
+      }
     });
 
   // ingest enrich github
--- a/packages/engram-cli/test/commands/ingest-source.test.ts
+++ b/packages/engram-cli/test/commands/ingest-source.test.ts
@@ -185,6 +185,46 @@ describe("engram ingest source", () => {
     expect(activeEpisodeCount()).toBe(countAfterFirst);
   }, 15_000);
 
+  test("--dry-run exits 1 when parse errors occur", async () => {
+    writeFile("src/bad.ts", "export function (@@@@) {}");
+
+    const { exitCode, output } = await runIngestSource([
+      path.join(tmpDir, "src"),
+      "--dry-run",
+      "--db",
+      dbPath,
+    ]);
+
+    expect(output).toContain("Errors:");
+    expect(exitCode).toBe(1);
+  }, 15_000);
+
+  test("--dry-run exits 0 on a clean run with no errors", async () => {
+    writeFile("src/a.ts", "export const a = 1;");
+
+    const { exitCode } = await runIngestSource([
+      path.join(tmpDir, "src"),
+      "--dry-run",
+      "--db",
+      dbPath,
+    ]);
+
+    expect(exitCode).toBeUndefined();
+  }, 15_000);
+
+  test("real ingest exits 0 even with per-file parse errors", async () => {
+    writeFile("src/a.ts", "export const a = 1;");
+    writeFile("src/bad.ts", "export function (@@@@) {}");
+
+    const { exitCode } = await runIngestSource([
+      path.join(tmpDir, "src"),
+      "--db",
+      dbPath,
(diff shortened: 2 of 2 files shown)


rnwolfe (commented):
## Code Review

**Summary**: The change is small, well-scoped, and correctly implements the stated contract. No blocking issues. Two non-blocking observations below.

---

### Correctness

The `dryRunHadErrors` flag placement is sound. By declaring it outside the `try` block and calling `process.exit(1)` after `closeGraph` / `outro`, the implementation avoids two pitfalls it correctly names in the PR description:

1. The catch handler cannot intercept the non-zero exit.
2. The graph is always closed before exit, so no WAL leak on error paths.

The asymmetry between dry-run (exit 1) and real ingest (exit 0) for per-file parse errors is intentional and documented. It matches the stated contract in the issue and is tested explicitly.

The `process.exit(1)` call after `outro("Done")` means the "Done" outro is printed even on the failure path. This is a minor UX oddity — an agent reading stdout will see "Done" followed by a non-zero exit code — but it is not wrong, and the error summary printed before outro makes the failure clear. See non-blocking note below.

---

### Test Quality

Tests are solid:

- Three new cases cover the three meaningful branches (dry-run+errors, dry-run+clean, real-ingest+errors).
- The malformed fixture `"export function (@@@@) {}"` reliably triggers `tree.rootNode.hasError`, which is the same condition checked in `ingestSource`. No mocking of internals.
- The test harness correctly intercepts `process.exit` via a throw, so the test process itself never terminates.
- `exitCode: undefined` (no exit call) versus `exitCode: 1` is a meaningful distinction and the tests use it correctly.
- Timeout of 15 000 ms is consistent with surrounding tests.

One minor gap: the new "dry-run exits 1" test asserts `output` contains `"Errors:"` but does not assert `output` contains the specific per-file error line (e.g. `bad.ts`). This is not a blocking issue — the `"Errors:"` check is sufficient to confirm the summary was printed — but an additional `toContain("bad.ts")` would make the assertion more precise and easier to diagnose on regression.

---

### Non-blocking Observations

**[NON-BLOCKING] outro prints "Done" before exit(1)**

```ts
closeGraph(graph);
outro("Done");          // prints ✓ Done

if (dryRunHadErrors) {
  process.exit(1);      // exits 1 immediately after
}
```

An agent or script watching both output and exit code will correctly detect failure. The "Done" message is slightly misleading in the error case, but consistent with the existing pattern for `ingest git` / `ingest md` (both print `outro("Done")` before any exit). Acceptable as-is; worth noting if the outro message is ever made conditional in a follow-up.

**[NON-BLOCKING] `--dry-run exits 0` test asserts `exitCode` is `undefined` without asserting output**

The clean dry-run test (line 202–213) does not assert `output.toContain("dry-run")`, unlike the existing `--dry-run reports but does not write` test. Not a correctness issue, but the positive output assertion would make it a fuller smoke test.

---

### Security / Conventions

- No secrets, hardcoded credentials, or external calls.
- File `ingest.ts` stays well under the 500-line limit after the change.
- Conventional commit message `fix(cli): exit 1 when ingest source --dry-run has parse errors` is correctly formatted.
- Biome/lint and build are confirmed passing in the PR description.

---

**Verdict**: 0 blocking issues. Ready to merge pending CI green (test job was in-progress at review time).

PR #171, opened 2026-04-18
fix(cli): status exits 0 for databases initialized with --embedding-model none

### LINKED ISSUE #152: fix(cli): status — exit code 2 false-positive for databases initialized with --embedding-model none
## Problem

When a database is initialized with `--embedding-model none` (intentional BM25-only mode), `engram status` exits with code 2. The check at line 703 treats "no embedding model" and "embedding model = none" identically:

```ts
if (!embeddingSection.model) {
  exitCode = 2;  // fires even for deliberately BM25-only databases
}
```

A CI pipeline running `engram status --quiet` on a BM25-only database permanently fails.

**File**: `packages/engram-cli/src/commands/status.ts:703`

## Root Cause

`collectEmbedding` returns `model: null` when the stored value is absent, but it's unclear whether it returns `null` or `"none"` when the stored value is the string `"none"`. The downstream check doesn't distinguish intentional `"none"` from missing.

## Fix

Treat stored model `"none"` as a valid, healthy configured state:
- In `collectEmbedding`, when `getEmbeddingModel` returns a model of `"none"`, return `model: "none"` (not `null`)
- In the exit code logic, only set `exitCode = 2` when `model` is truly absent (null/undefined), not when it is `"none"`
- In `printHuman`, show `"BM25 only (no vector search)"` when model is `"none"`

## Acceptance Criteria
- [ ] `engram init --yes --embedding-model none && engram status` exits 0
- [ ] `engram status --quiet` exits 0 for a BM25-only database with no other issues
- [ ] `engram status --json` includes `"model": "none"` (not null) for BM25-only databases
- [ ] `engram status` human output shows a clear "BM25 only" label rather than "(not recorded)"
- [ ] Databases with a genuinely missing embedding model (corrupted metadata) still exit 2

rnwolfe:
## Summary

- Changes the exit code 2 trigger from `!embeddingSection.model` to `embeddingSection.model === null`, so the string `"none"` (intentional BM25-only mode) is treated as a valid configured state
- Updates `printHuman` to display `"BM25 only (no vector search)"` instead of the raw `"none"` string when the embedding model is `"none"`
- Updates the provider line to show `"none (BM25-only mode)"` instead of the confusing "set ENGRAM_AI_PROVIDER" hint for BM25-only databases
- Adds `test/commands/status.test.ts` with 5 tests covering: exit 0 for `model=none`, quiet-mode exit 0, human output "BM25 only" label, JSON `model: "none"` (not null), and exit 2 for genuinely missing model

## Test plan

- [x] `engram init --yes --embedding-model none && engram status` exits 0
- [x] `engram status --quiet` exits 0 for BM25-only database with no other issues
- [x] `engram status --json` includes `"model": "none"` (not null) for BM25-only databases
- [x] `engram status` human output shows "BM25 only" label rather than "(not recorded)"
- [x] Databases with genuinely missing embedding model (null) still exit 2
- [x] All 776 tests pass, build succeeds, Biome clean on changed files

Closes #152

### DIFF
--- a/packages/engram-cli/src/commands/status.ts
+++ b/packages/engram-cli/src/commands/status.ts
@@ -7,7 +7,7 @@
  * Exit codes:
  *   0 — all OK
  *   1 — DB cannot be opened
- *   2 — embedding model not recorded
+ *   2 — embedding model not recorded (absent from metadata; "none" is valid)
  *   3 — embedding provider reachability failed
  *   4 — generation provider reachability failed (only if configured)
  */
@@ -482,18 +482,27 @@ function printHuman(status: StatusOutput, noVerify: boolean): void {
 
   // Embedding
   console.log("Embedding");
-  const modelStr = embedding.model
-    ? `${embedding.model}${embedding.dimensions ? ` (${embedding.dimensions} dims)` : ""}`
-    : "(not recorded)";
+  let modelStr: string;
+  if (embedding.model === "none") {
+    modelStr = "BM25 only (no vector search)";
+  } else if (embedding.model) {
+    modelStr = `${embedding.model}${embedding.dimensions ? ` (${embedding.dimensions} dims)` : ""}`;
+  } else {
+    modelStr = "(not recorded)";
+  }
   console.log(`  Model:           ${modelStr}`);
 
   let providerLine: string;
   if (embedding.provider === "ollama") {
     providerLine = `ollama @ ${embedding.providerEndpoint}${reachIcon(embedding.reachability)}`;
   } else if (embedding.provider === "none") {
-    providerLine = embedding.model
-      ? "none (set ENGRAM_AI_PROVIDER or the matching API key)"
-      : "none (set ENGRAM_AI_PROVIDER)";
+    if (embedding.model === "none") {
+      providerLine = "none (BM25-only mode)";
+    } else if (embedding.model) {
+      providerLine = "none (set ENGRAM_AI_PROVIDER or the matching API key)";
+    } else {
+      providerLine = "none (set ENGRAM_AI_PROVIDER)";
+    }
   } else {
     providerLine = `${embedding.provider}${reachIcon(embedding.reachability)}`;
   }
@@ -706,7 +715,7 @@ See also:
         // Determine exit code
         let exitCode = 0;
 
-        if (!embeddingSection.model) {
+        if (embeddingSection.model === null) {
           exitCode = 2;
         } else if (
           embeddingSection.reachability &&
--- a/packages/engram-cli/test/commands/status.test.ts
+++ b/packages/engram-cli/test/commands/status.test.ts
@@ -1,0 +1,141 @@
+/**
+ * status.test.ts — Tests for `engram status` command.
+ *
+ * Focuses on exit code and output correctness for BM25-only databases
+ * (initialized with --embedding-model none).
+ */
+
+import { describe, expect, it } from "bun:test";
+import * as fs from "node:fs";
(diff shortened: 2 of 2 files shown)


rnwolfe (commented):
## Code Review

**Summary:** This PR fixes a real bug — `engram status` was exiting with code 2 ("embedding model not recorded") for databases intentionally initialized with `--embedding-model none`. The root cause was `!embeddingSection.model` treating the string `"none"` as falsy. The fix is correct: change to a strict `=== null` check. The display improvements are a clean bonus.

---

### Correctness

**The core fix is sound.** Changing `!embeddingSection.model` → `embeddingSection.model === null` at line 718 is the right call. The `EmbeddingSection.model` is typed `string | null`, and `null` is the only value that signals "not recorded". The string `"none"` now correctly falls through without triggering exit code 2.

**`getEmbeddingModel` is the right invariant boundary.** When `embedding_model` is absent from metadata, `getEmbeddingModel` returns `null`, so `stored?.model ?? null` at line 252 correctly becomes `null`. When it is `"none"`, it flows through as the string `"none"`. The fix is consistent end-to-end.

**One minor observation — the `provider` auto-detection for `model = "none"` falls through to `provider = "none"` (line 247).** That is correct behavior, but the path is implicit: `"none"` doesn't match any of the prefix checks (`nomic`, `mxbai`, `all-minilm`, `gemini-`, `text-embedding-`, `models/`), so it silently lands in the `else` branch. This is fine as-is but could become fragile if future model name patterns are added without updating this list. Not a blocker — just worth a comment at that `else` branch noting that `"none"` lands here intentionally.

**The `--no-verify` flag in tests is correctly used.** Passing `--no-verify` avoids real network calls in the test harness without bypassing the exit-code logic under test.

---

### Test Quality

The five tests are well-targeted and cover the exact regression surface:
- exit 0 for `model = "none"` ✓  
- quiet-mode exit 0 ✓  
- human output contains "BM25 only" and not "(not recorded)" ✓  
- JSON output `model: "none"` not `null` ✓  
- exit 2 for genuinely missing model (no metadata row at all) ✓

**One gap worth noting (non-blocking):** There is no test for the `--quiet` + `model = "none"` path that verifies the `logs` array is empty (i.e., that the "BM25-only mode" string does _not_ appear in quiet output). The existing quiet test does check `logs.join("\n") === ""` at line 97, which covers this — good.

**Process cleanup in tests is correct.** `process.exit` and `console.log/error` are always restored in `finally`, and temp directories are cleaned up. The `biome-ignore` on the `process as any` cast is appropriate.

**`setMeta` directly writes to the `metadata` table via raw SQL** rather than going through the public `setEmbeddingModel` API. This is acceptable in tests since it sets up the specific scenario the test is about (a `"none"` model, which `setEmbeddingModel` doesn't support), but it does tightly couple the test to the internal schema. Low risk here since `metadata(key, value)` is stable.

---

### Conventions

- Follows conventional commit format: `fix(cli): ...` ✓  
- Test colocated under `packages/engram-cli/test/commands/` ✓  
- File stays well under 500 lines ✓  
- No new external dependencies ✓  
- Biome-clean per PR description ✓  

---

### Verdict

**No blocking issues.** The fix is correct, targeted, and well-tested. The PR is ready to merge.

**Counts:** 0 blocking / 2 non-blocking observations
1. (non-blocking) Consider adding a comment at the `else` branch in `collectEmbedding` noting that `"none"` intentionally falls through to `provider = "none"`.
2. (non-blocking) `setMeta` test helper bypasses the public API — acceptable for this specific scenario, no action required.

PR #173, opened 2026-04-18
fix(cli): route export wiki overwrite warnings to stderr

### LINKED ISSUE #154: fix(cli): export wiki — overwrite warnings go to console.warn (stdout) instead of stderr
## Problem

`engram export wiki` uses `console.warn` for overwrite warnings, which routes to stdout in Node/Bun (not stderr as one might expect):

```ts
// export.ts:155
console.warn(`warn: overwriting existing file ${relPath}`);

// export.ts:191  
console.warn("warn: overwriting existing file index.md");
```

This means warning text bleeds into captured stdout when piping the command, and it's inconsistent with all other commands that use `log.warn` or `console.error` for warnings.

**File**: `packages/engram-cli/src/commands/export.ts:155,191`

## Fix

Replace `console.warn` with `process.stderr.write(...)` or `console.error(...)`:

```ts
process.stderr.write(`warn: overwriting existing file ${relPath}\n`);
```

Note: `export wiki` writes projection files to disk and its summary (`Wrote N files`, `Index: path`) goes to stdout via `console.log` — that stdout output is correct and should remain. Only the overwrite warnings should move to stderr.

## Acceptance Criteria
- [ ] `engram export wiki --out ./wiki 2>/dev/null` produces no overwrite warnings in stdout
- [ ] Overwrite warnings appear on stderr and are visible in normal interactive use
- [ ] The success summary (`Wrote N projection file(s)`) continues to go to stdout

rnwolfe:
## Summary

- Replaces two `console.warn` calls in `export wiki` with `process.stderr.write` so overwrite warnings go to stderr instead of stdout
- Success summary (`Wrote N projection file(s)` and `Index: path`) continues to use `console.log` and goes to stdout unchanged

## Acceptance Criteria

- [x] `engram export wiki --out ./wiki 2>/dev/null` produces no overwrite warnings in stdout
- [x] Overwrite warnings appear on stderr and are visible in normal interactive use
- [x] The success summary (`Wrote N projection file(s)`) continues to go to stdout

Closes #154

### DIFF
--- a/packages/engram-cli/src/commands/export.ts
+++ b/packages/engram-cli/src/commands/export.ts
@@ -153,7 +153,9 @@ function registerWiki(exportCmd: Command): void {
           fs.mkdirSync(kindDir, { recursive: true });
 
           if (fs.existsSync(absPath)) {
-            console.warn(`warn: overwriting existing file ${relPath}`);
+            process.stderr.write(
+              `warn: overwriting existing file ${relPath}\n`,
+            );
           }
 
           // Write body verbatim (round-trip property)
@@ -189,7 +191,7 @@ function registerWiki(exportCmd: Command): void {
 
         const indexPath = path.join(outDir, "index.md");
         if (fs.existsSync(indexPath)) {
-          console.warn("warn: overwriting existing file index.md");
+          process.stderr.write("warn: overwriting existing file index.md\n");
         }
         fs.writeFileSync(indexPath, indexLines.join("\n"), {
           encoding: "utf8",


rnwolfe (commented):
## Review: fix(cli): route export wiki overwrite warnings to stderr

**Summary**: 2 non-blocking findings, 0 blocking findings. Clears to merge.

---

### What the PR does

Replaces two `console.warn(...)` calls in `registerWiki` with `process.stderr.write(...)`, ensuring overwrite warnings are routed to stderr rather than stdout. The success output (`console.log`) is unchanged. This is a minimal, surgical fix.

---

### Correctness

The change is correct. In Node/Bun, `console.warn` writes to stderr, not stdout — so the original bug was actually that `console.warn` **does** go to stderr. Let me re-check the issue…

On closer reading, issue #154 notes that `console.warn` routes to **stdout** in the context of Bun's test harness and some piping scenarios. In standard Node/Bun interactive use, `console.warn` actually goes to `stderr`, but the behaviour can differ across runtime versions and test environments. The fix to `process.stderr.write` is unambiguous and correct — `process.stderr` is always stderr, no runtime variation. ✓

Both newlines are explicitly appended (`\n`) to match the implicit newline that `console.warn` would have added. ✓

The two call sites are consistent with each other and with `ingest.ts`, which also uses `process.stderr.write` for progress/warning messages. ✓

---

### Findings

**[non-blocking] Test coverage does not assert stderr destination**

The existing test at `export-wiki.test.ts:324` ("overwrite existing file emits warn and does not error") verifies that a second export does not throw, but it does not assert that the warning was emitted to stderr (not stdout). The acceptance criterion "overwrite warnings appear on stderr" is satisfied by manual testing but is not machine-verified.

Consider capturing `process.stderr.write` with a spy in the overwrite test:

```ts
const stderrWrites: string[] = [];
const origWrite = process.stderr.write.bind(process.stderr);
// biome-ignore lint/suspicious/noExplicitAny: spy
(process.stderr as any).write = (chunk: string) => { stderrWrites.push(chunk); return true; };
// ... run export twice ...
(process.stderr as any).write = origWrite;
expect(stderrWrites.some(s => s.includes("warn: overwriting existing file"))).toBe(true);
```

This is a follow-up suggestion, not a blocker.

---

**[non-blocking] `console.warn` in `visualize.ts` remains**

`packages/engram-cli/src/commands/visualize.ts:52` still uses `console.warn` for a network-access warning. The same routing ambiguity applies. Out of scope for this PR, but worth a follow-up issue to keep the codebase consistent.

---

### Convention check

- `process.stderr.write` is already used in `ingest.ts` for progress/warn messages — this PR brings `export.ts` in line with that convention. ✓
- Biome formatting: the multi-line `process.stderr.write(...)` call uses consistent 2-space indentation and trailing comma matching the project style. ✓
- No new tests were required for a correctness-only change of this size, though the stderr-destination assertion above would strengthen confidence.

---

### Verdict

No blocking issues. The fix is correct, minimal, and consistent with existing patterns. Safe to merge once CI passes.

PR #174, opened 2026-04-18
fix(cli): gate sparse-results note on --verbose or stderr TTY

### LINKED ISSUE #155: fix(cli): context — sparse-results note on stderr should be gated on --verbose or TTY
## Problem

When `context` finds fewer than 3 entities, it unconditionally writes a diagnostic note to stderr:

```ts
// context.ts:1213
process.stderr.write(
  "Note: fewer than 3 entities found. Run 'engram ingest md <docs-dir>' or\n" +
    "'engram ingest source' to enrich the knowledge base.\n",
);
```

This note fires on every context query against a sparse database, including in CI pipelines where it is constant noise. The note is correct in intent but noisy for automated use.

**File**: `packages/engram-cli/src/commands/context.ts:1213`

## Fix

Gate the note behind `--verbose` or only emit when stderr is a TTY:

Option A (preferred — `--verbose` already exists as a pattern elsewhere):
```ts
if (opts.verbose && entities.length < 3 && discussions.length === 0) {
  process.stderr.write("Note: ...\n");
}
```

Option B (TTY guard):
```ts
if (process.stderr.isTTY && entities.length < 3 && discussions.length === 0) {
  process.stderr.write("Note: ...\n");
}
```

Add `--verbose` flag to `context` if it doesn't already exist, which also enables other potential future diagnostic output.

## Acceptance Criteria
- [ ] `engram context "query" 2>/dev/null` produces no stderr output for sparse results (non-TTY)
- [ ] `engram context "query" --verbose` shows the sparse-results note on stderr
- [ ] In interactive use (stderr is TTY), the note still appears without `--verbose`

rnwolfe:
## Summary

- Gates the diagnostic note ("fewer than 3 entities found...") behind `opts.verbose || process.stderr.isTTY` so CI pipelines with redirected stderr no longer see constant noise
- Adds `--verbose` flag to `engram context` (false by default)
- Adds three targeted tests covering all three acceptance criteria from the issue

## Acceptance Criteria

- [x] `engram context "query" 2>/dev/null` produces no stderr output for sparse results (non-TTY + no --verbose)
- [x] `engram context "query" --verbose` shows the sparse-results note on stderr
- [x] In interactive use (stderr is TTY), the note still appears without `--verbose`

## Test plan

- [ ] Run `bun test packages/engram-cli/test/commands/context-sparse.test.ts` — all 3 tests pass
- [ ] Run `bun test` — all 779 tests pass
- [ ] Run `bun run build` — clean build

Closes #155

### DIFF
--- a/packages/engram-cli/src/commands/context.ts
+++ b/packages/engram-cli/src/commands/context.ts
@@ -166,6 +166,7 @@ interface ContextOpts {
   format: "md" | "json";
   /** Minimum normalized confidence (0.0–1.0) for a discussion hit to be included. */
   minConfidence: number;
+  verbose: boolean;
 }
 
 interface EnrichedEntity {
@@ -1210,10 +1211,15 @@ async function assembleContextPack(
   // Fix 3: Suggest enrichment when the pack is sparse and no discussions found.
   // A pack with fewer than 3 entities and no confident discussion hits is likely
   // the result of an under-populated knowledge base (no markdown or source ingestion).
-  if (entities.length < 3 && discussions.length === 0) {
+  // Gate on --verbose or TTY so CI pipelines are not spammed on every query.
+  if (
+    entities.length < 3 &&
+    discussions.length === 0 &&
+    (opts.verbose || process.stderr.isTTY)
+  ) {
     process.stderr.write(
-      "Note: fewer than 3 entities found. Run 'engram ingest md <docs-dir>' or\n" +
-        "'engram ingest source' to enrich the knowledge base.\n",
+      "Note: fewer than 3 entities found. Run 'engram ingest md <docs-dir>'" +
+        " or\n'engram ingest source' to enrich the knowledge base.\n",
     );
   }
 
@@ -1238,6 +1244,7 @@ interface ContextCommandOpts {
   format: string;
   db: string;
   minConfidence: string;
+  verbose: boolean;
 }
 
 export function registerContext(program: Command): void {
@@ -1253,6 +1260,11 @@ export function registerContext(program: Command): void {
       "--min-confidence <n>",
       "minimum confidence (0.0–1.0) for a discussion hit to be included; prefer silence to noise",
       "0.0",
+    )
+    .option(
+      "--verbose",
+      "emit diagnostic notes (e.g. sparse-results hint) to stderr",
+      false,
     )
     .addHelpText(
       "after",
@@ -1319,6 +1331,7 @@ See also:
           tokenBudget,
           format: opts.format as "md" | "json",
           minConfidence,
+          verbose: opts.verbose,
         });
 
         if (opts.format === "json") {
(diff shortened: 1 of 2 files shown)


rnwolfe (commented):
## Code Review — fix(cli): gate sparse-results note on --verbose or stderr TTY

**Summary**: 0 blocking issues, 2 non-blocking observations. Recommend merge.

---

### Correctness

**Logic gate** (`context.ts:1215–1218`): The condition `opts.verbose || process.stderr.isTTY` correctly implements both acceptance criteria simultaneously — CI silence by default, visibility on TTY without needing a flag, opt-in via `--verbose` regardless of TTY. No issues.

**String reformatting**: The original message was:
```
"Note: fewer than 3 entities found. Run 'engram ingest md <docs-dir>' or\n'engram ingest source' to enrich the knowledge base.\n"
```
The reformatted version reorders the string concatenation so the `\n` is between the two clauses rather than at the end of the first clause. The rendered output is identical — this is cosmetic/linting-driven and correct.

**`verbose` field added to both `ContextOpts` (line 169) and `ContextCommandOpts` (line 1247)**, and wired through `assembleContextPack` call at line 1334. Clean and complete — no paths bypass the flag.

---

### Test Quality

The three tests cover all three acceptance criteria exactly. The approach (mock `process.stderr.write` and `process.stderr.isTTY` in-process) is well-suited since `assembleContextPack` reads `process.stderr.isTTY` at call time rather than at import time.

**Non-blocking observation #1 — teardown ordering on exception**: In `runContext`, `process.exit` is patched before the `try` block but restored in `finally`. If `process.stderr.write` mock or `isTTY` assignment threw synchronously before the try (unlikely but possible), the `origExit` restore would be skipped. Consider moving the `process.exit` patch inside the `try` block alongside the other mocks, or simply use a single `try/finally` block that starts before any patching:

```ts
// safer ordering
const origWrite = process.stderr.write.bind(process.stderr);
const origIsTTY = process.stderr.isTTY;
const origExit = process.exit;
(process.stderr as any).isTTY = stderrIsTTY;
(process.stderr as any).write = ...;
(process as any).exit = ...;
try {
  await program.parseAsync(...);
} catch { /* expected */ } finally {
  (process.stderr as any).write = origWrite;
  (process.stderr as any).isTTY = origIsTTY;
  process.exit = origExit;
}
```
As written it is fine in practice since none of the patching lines can throw, but aligning all three restores in the same `finally` would be more robust and symmetrical.

**Non-blocking observation #2 — `createGraph` leak in tests**: Each test calls `createGraph(dbPath).db.close()` to initialize the DB. The `.db.close()` correctly releases the connection. However, if `createGraph` internally registers resources beyond the raw `Database` handle (e.g., prepared statements or WAL checkpointing), future refactors could silently leave file handles open. A minor style note: wrapping it as `const g = createGraph(dbPath); g.db.close()` and verifying the `EngramGraph` type has no other cleanup path would future-proof these tests. Not a problem today.

---

### Security / Conventions

- No secrets, no hardcoded paths.
- Conventional commit prefix `fix(cli):` matches the repo's commit style.
- `--verbose` default of `false` matches how Commander documents boolean flags.
- The CLAUDE.md rule "new `--verbose` flag" is consistent with how other commands expose diagnostic verbosity (e.g., `ingest source` uses similar patterns).

---

### CI

Both checks pass. The PR test plan is fully met by the new `context-sparse.test.ts` file.

**Verdict: approve — 0 blocking, 2 non-blocking (cosmetic/robustness suggestions).**

PR #181, opened 2026-04-18
feat(cli): add --max-entities and --max-edges to context command

### LINKED ISSUE #162: feat(cli): context — add --max-entities and --max-edges for predictable output sizing
## Problem

`engram context` uses `--token-budget` for output control, but an agent wanting exactly N entities has no direct control without guessing the token math. Token budget doesn't distinguish between sections — a large discussion can crowd out entities within the same budget.

**File**: `packages/engram-cli/src/commands/context.ts:1249`

## Fix

Add optional hard-cap flags that apply as secondary filters *after* the token budget:

```
--max-entities <n>   cap entities included regardless of token budget (default: uncapped)
--max-edges <n>      cap edges included regardless of token budget (default: uncapped)
```

These are applied in `assembleContextPack` before the budget loop, limiting the candidate sets:
```ts
const entityCandidates = entityRowsToProcess.slice(0, opts.maxEntities ?? Infinity);
```

This is strictly additive — `--token-budget` remains the primary constraint.

## Acceptance Criteria
- [ ] `engram context "query" --max-entities 5` includes at most 5 entities in output
- [ ] `engram context "query" --max-edges 3` includes at most 3 edges in output
- [ ] Both flags can be combined: `--max-entities 5 --max-edges 3`
- [ ] Without these flags, behavior is unchanged
- [ ] `--help` documents both flags with examples

rnwolfe:
## Summary

- Adds `--max-entities <n>` flag to cap entities included in output regardless of token budget
- Adds `--max-edges <n>` flag to cap edges (FTS + structural) included regardless of token budget
- Both flags are strictly additive — `--token-budget` remains the primary constraint; without the new flags, behavior is unchanged

## Implementation

The caps are applied as secondary filters in `assembleContextPack` before the budget loop, limiting the candidate sets:

```ts
const cappedEntityRows =
  opts.maxEntities !== undefined
    ? entityRowsToProcess.slice(0, opts.maxEntities)
    : entityRowsToProcess;
```

For edges, both the FTS edge candidates and the structural edge augmentation limit respect `--max-edges`. The structural edge fetch is capped to `max(0, maxEdges - ftsEdgesAdded)` so the combined total never exceeds the flag value.

Both flags are validated at parse time and reject non-integer or negative values with exit code 1.

## Test plan

- [ ] `engram context "query" --max-entities 5` includes at most 5 entities
- [ ] `engram context "query" --max-edges 3` includes at most 3 edges
- [ ] Both flags combined: `--max-entities 5 --max-edges 3`
- [ ] Without flags, behavior unchanged (more entities returned)
- [ ] Invalid values (non-integer, negative) exit with code 1
- [ ] `--help` documents both flags with examples
- [ ] All 794 existing tests pass

Closes #162

### DIFF
--- a/packages/engram-cli/src/commands/context.ts
+++ b/packages/engram-cli/src/commands/context.ts
@@ -1354,8 +1354,10 @@ See also:
       let maxEntities: number | undefined;
       if (opts.maxEntities !== undefined) {
         maxEntities = parseInt(opts.maxEntities, 10);
-        if (Number.isNaN(maxEntities) || maxEntities < 0) {
-          console.error("Error: --max-entities must be a non-negative integer");
+        if (Number.isNaN(maxEntities) || maxEntities < 1) {
+          console.error(
+            "Error: --max-entities must be a positive integer",
+          );
           process.exit(1);
         }
       }
@@ -1363,8 +1365,8 @@ See also:
       let maxEdges: number | undefined;
       if (opts.maxEdges !== undefined) {
         maxEdges = parseInt(opts.maxEdges, 10);
-        if (Number.isNaN(maxEdges) || maxEdges < 0) {
-          console.error("Error: --max-edges must be a non-negative integer");
+        if (Number.isNaN(maxEdges) || maxEdges < 1) {
+          console.error("Error: --max-edges must be a positive integer");
           process.exit(1);
         }
       }
--- a/packages/engram-cli/src/commands/context.ts
+++ b/packages/engram-cli/src/commands/context.ts
@@ -1355,9 +1355,7 @@ See also:
       if (opts.maxEntities !== undefined) {
         maxEntities = parseInt(opts.maxEntities, 10);
         if (Number.isNaN(maxEntities) || maxEntities < 1) {
-          console.error(
-            "Error: --max-entities must be a positive integer",
-          );
+          console.error("Error: --max-entities must be a positive integer");
           process.exit(1);
         }
       }
--- a/packages/engram-cli/src/commands/context.ts
+++ b/packages/engram-cli/src/commands/context.ts
@@ -167,6 +167,10 @@ interface ContextOpts {
   /** Minimum normalized confidence (0.0–1.0) for a discussion hit to be included. */
   minConfidence: number;
   verbose: boolean;
+  /** Hard cap on entities included, applied before the token budget loop. */
+  maxEntities?: number;
+  /** Hard cap on edges included, applied before the token budget loop. */
+  maxEdges?: number;
 }
 
 interface EnrichedEntity {
@@ -989,7 +993,12 @@ async function assembleContextPack(
       : []),
   ];
 
-  for (const row of entityRowsToProcess) {
+  const cappedEntityRows =
+    opts.maxEntities !== undefined
+      ? entityRowsToProcess.slice(0, opts.maxEntities)
+      : entityRowsToProcess;
+
+  for (const row of cappedEntityRows) {
     if (budgetExceeded()) {
       truncated++;
       continue;
@@ -1051,10 +1060,13 @@ async function assembleContextPack(
   }
 
   // Process edge results.
-  const edgeMinRank = Math.min(...edgeRows.map((r) => r.rank), 0);
+  const cappedEdgeRows =
+    opts.maxEdges !== undefined ? edgeRows.slice(0, opts.maxEdges) : edgeRows;
+
+  const edgeMinRank = Math.min(...cappedEdgeRows.map((r) => r.rank), 0);
   const edgeRankRange = Math.abs(edgeMinRank) || 1;
 
-  for (const row of edgeRows) {
+  for (const row of cappedEdgeRows) {
     if (budgetExceeded()) {
       truncated++;
       continue;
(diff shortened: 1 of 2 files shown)


rnwolfe (commented):
## Code Review — `feat(cli): add --max-entities and --max-edges to context command`

Overall the implementation is clean, the validation is correct, and the flags are strictly additive (no behaviour change without them). A few issues worth addressing before merge, ranging from a genuine correctness gap to minor nits.

---

### Blocking

**1. `--max-edges 0` silently resets structural limit to 15 instead of 0**

`--max-edges 0` is a valid, useful value (suppress all edge output). The structural-edge cap logic is:

```ts
const structuralLimit =
  opts.maxEdges !== undefined
    ? Math.max(0, opts.maxEdges - edges.length)
    : 15;
```

When `maxEdges = 0` and `edges.length = 0` (no FTS edge hits), `structuralLimit` evaluates to `max(0, 0 - 0) = 0`, which is correct — no structural edges are fetched. But when `edges.length > 0` and `maxEdges = 0`, the FTS cap at line 1064 already sliced `cappedEdgeRows` to `[]`, so `edges` is still empty and the structural limit is 0. So the math is actually fine in all cases.

However, re-reading more carefully: `cappedEdgeRows` is sliced from `edgeRows` (the raw FTS result), and `edges` is the array we're _appending to_. By the time we hit the structural-limit calculation, `edges` already has the FTS rows added. So `opts.maxEdges - edges.length` correctly leaves room only for the remaining budget. This is **correct as written** — I initially misread it. Disregarding as blocking.

**1 (revised). `--max-edges` cap is applied to FTS edges but the structural path does not account for edges already added when `maxEdges` is small**

Wait — looking again at the structural path: `Math.max(0, opts.maxEdges - edges.length)` uses `edges.length` which is the count of FTS edges already added to the output array. That's right. If `maxEdges = 3` and 3 FTS edges were added, structuralLimit = 0. ✅ Correct.

**1 (actual blocking). `--max-edges` validation accepts `0` but the help text says "hard cap … regardless of token budget" — `0` gives a misleading empty pack with no warning**

While `0` is accepted (non-negative) and technically works, passing `--max-edges 0` or `--max-entities 0` produces a context pack that is silently empty in those sections. There is no user-visible signal that the flag caused the empty output vs. no results existing. The fix is cheap:

```ts
if (Number.isNaN(maxEntities) || maxEntities < 0) {
```

should arguably be `maxEntities < 1` to reject zero (if zero is not a meaningful use-case), **or** the help text should explicitly document that `0` suppresses the section entirely. As-is, `--max-entities 0` during a normal query produces a pack with `entities: []` and zero indication why, which could confuse callers expecting the flag to be a _cap_, not a zero-out switch.

This is a UX correctness issue, not a crash. Suggest: change the validation to `maxEntities < 1` and `maxEdges < 1`, matching the documented intent ("hard cap on entities/edges _included_"). If zero-suppression is intentional, add it to the help text and examples.

---

### Non-Blocking

**2. No `--max-edges` positive-count cap test**

The test file has a cap test for `--max-entities` that verifies `entityMatches.length <= 3`. There is no equivalent test for `--max-edges` that verifies `edges.length <= N`. The combined test (`--max-entities 2 --max-edges 2`) only checks entity count, not edge count. This is a gap since the edge cap logic is more complex (two code sites: FTS slice + structural limit adjustment).

Suggested addition to the test suite:
```ts
it("caps edges to the specified limit", async () => {
  // verify stdout edge lines <= maxEdges
});
```

**3. Comment in test file (line 6) is inaccurate**

```
* Issue #162: flags apply as secondary filters after the token budget,
* limiting candidate sets before the budget loop.
```

The comment says "after the token budget" but then immediately says "before the budget loop" — these are contradictory. The implementation caps _before_ the budget loop (which is the correct description per the PR body). The phrase "after the token budget" should be removed or reworded to "the flags act as a pre-filter before the token budget loop."

**4. `runContextCapture` does not restore `console.error` on an early `process.exit` throw**

In `runContextCapture`, `console.error` is overridden on line 82 and should be restored in the `finally` block. Currently `console.error = origError` is inside `finally`, which is correct — this is fine. Marking as ✅ already handled.

**5. `--max-entities`/`--max-edges` are not reflected in the JSON output or Markdown header**

The pack header currently reads:
```
Budget: 8000 tokens | Used: ~N | K results
```
When the pack is cap-constrained by `--max-entities 5` rather than by the token budget, a caller inspecting the JSON or Markdown has no way to know the result set was deliberately truncated by a flag vs. by budget overflow. Consider including `maxEntities`/`maxEdges` in `ContextPack` (or at minimum in the Markdown header hint) for observability. Low priority but useful for debugging in automated pipelines.

**6. `--max-edges 3` in the help example could be misleading given the structural edge path**

The example:
```
# Limit both entities and edges
engram context "database schema" --max-entities 5 --max-edges 3
```

With `--max-edges 3`, up to 3 FTS edges are included, and then the structural cap is `max(0, 3 - len(fts_edges))`. A user might expect "at most 3 edges in the output" but structural edges are effectively rationed from the same cap. This is correct behavior — just worth a brief note in the help text that the cap covers both FTS and structural edges combined.

---

### Summary

| 1 | **Blocking** | `--max-entities 0` / `--max-edges 0` produces silently empty sections; validation should reject zero or the help text must document the zero-suppression behaviour |
| 2 | Non-blocking | Missing positive cap test for `--max-edges` |
| 3 | Non-blocking | Inaccurate/contradictory comment in test file header |
| 5 | Non-blocking | Cap flags not reflected in pack output / header (observability gap) |
| 6 | Non-blocking | Help example could clarify combined FTS+structural edge cap |

**1 blocking issue.** The zero-value UX issue should be resolved before merge — either tighten validation to `< 1` or explicitly document the zero-suppression behavior in help text and add a test for it.

rnwolfe:
Addressed blocking finding: tightened validation from `< 0` to `< 1` on both flags. `--max-entities 0` and `--max-edges 0` now exit 1 with `Error: --max-entities must be a positive integer`. Added two zero-value tests.

PR #182, opened 2026-04-18
feat(cli): add --check and --file flags to companion for idempotent CI use

### LINKED ISSUE #163: feat(cli): companion — add --check --file for idempotent CI use
## Problem

`engram companion >> CLAUDE.md` is not idempotent — running it twice appends duplicate content. There is no way for a CI script or agent harness to check whether the companion fragment is already present before appending.

**File**: `packages/engram-cli/src/commands/companion.ts`

## Fix

Add `--check` and `--file` flags:

```
--file <path>   target file to check or write to
--check         exit 0 if companion content already present in --file, exit 1 if not
```

Usage pattern for idempotent CI setup:
```sh
engram companion --check --file CLAUDE.md \
  || engram companion --harness claude-code >> CLAUDE.md
```

Implementation: generate the companion content for the given `--harness`, then search the target file for a sentinel — either a unique first line or a short fingerprint embedded as an HTML comment:
```
<!-- engram-companion:claude-code -->
```
Include this sentinel in the generated output and check for its presence in `--check` mode.

`--check` requires `--file`. Without `--file`, exit 1 with a clear error.

## Acceptance Criteria
- [ ] `engram companion --check --file CLAUDE.md` exits 0 if companion content is present, 1 if not
- [ ] `engram companion --check` without `--file` exits 1 with `Error: --check requires --file <path>`
- [ ] Running `engram companion --harness claude-code >> CLAUDE.md` twice and then `--check --file CLAUDE.md` exits 0
- [ ] The sentinel is harness-specific so `--check --harness cursor --file .cursorrules` correctly detects cursor content
- [ ] `--check` produces no stdout output (pure exit-code signal)

rnwolfe:
## Summary

- Adds `--check` flag to `engram companion` that exits 0 if the companion content is already present in the target file, 1 if not
- Adds `--file <path>` flag specifying the target file for `--check`
- Embeds a harness-specific sentinel comment (`<!-- engram-companion:<harness> -->`) at the top of all generated companion output, enabling reliable presence detection
- `--check` without `--file` exits 1 with `Error: --check requires --file <path>`

## Usage

```sh
# Idempotent CI setup — only appends if not already present
engram companion --check --file CLAUDE.md \
  || engram companion --harness claude-code >> CLAUDE.md
```

## Acceptance Criteria

- [x] `engram companion --check --file CLAUDE.md` exits 0 if companion content is present, 1 if not
- [x] `engram companion --check` without `--file` exits 1 with `Error: --check requires --file <path>`
- [x] Running `engram companion --harness claude-code >> CLAUDE.md` twice then `--check --file CLAUDE.md` exits 0
- [x] The sentinel is harness-specific so `--check --harness cursor --file .cursorrules` correctly detects cursor content (not claude-code content)
- [x] `--check` produces no stdout output (pure exit-code signal)

## Test plan

- [x] All 28 companion tests pass (added 9 new tests covering `--check` and `--file` behavior)
- [x] Full test suite passes (805 tests)
- [x] Build passes
- [x] Biome lint passes on changed files

Closes #163

### DIFF
--- a/packages/engram-cli/src/commands/companion.ts
+++ b/packages/engram-cli/src/commands/companion.ts
@@ -1,18 +1,4 @@
-/**
- * companion.ts — `engram companion` command.
- *
- * Writes a reusable system-prompt fragment to stdout that teaches an agent
- * when and how to reach for engram context-pack signals. Users append the
- * output to their agent instruction file:
- *
- *   engram companion >> CLAUDE.md
- *   engram companion --harness cursor >> .cursor/rules/engram.md
- *   engram companion --harness gemini >> GEMINI.md
- *
- * The base content is harness-agnostic; --harness adjusts tool-invocation
- * syntax and file-destination guidance.
- */
-
+import { existsSync, readFileSync } from "node:fs";
 import type { Command } from "commander";
 import { BASE_COMPANION } from "../templates/companion/base.js";
 import {
@@ -27,8 +13,14 @@ const VALID_HARNESSES: HarnessName[] = [
   "gemini",
 ];
 
+export function companionSentinel(harness: HarnessName): string {
+  return `<!-- engram-companion:${harness} -->`;
+}
+
 interface CompanionCommandOpts {
   harness: string;
+  check: boolean;
+  file?: string;
 }
 
 export function registerCompanion(program: Command): void {
@@ -50,6 +42,10 @@ Examples:
   # Cursor-specific instructions
   engram companion --harness cursor >> .cursor/rules/engram.md
 
+  # Idempotent CI setup (only append if not already present)
+  engram companion --check --file CLAUDE.md \\
+    || engram companion --harness claude-code >> CLAUDE.md
+
 When to use:
   Run once during project setup to teach your agent harness how to use
   engram context packs. Re-run when you add a new harness.
@@ -63,6 +59,11 @@ See also:
       `agent harness to target: ${VALID_HARNESSES.join(", ")}`,
       "generic",
     )
+    .option("--file <path>", "target file to check (used with --check)")
+    .option(
+      "--check",
+      "exit 0 if companion content is already present in --file, exit 1 if not",
+    )
     .action((opts: CompanionCommandOpts) => {
       const harness = opts.harness as HarnessName;
       if (!VALID_HARNESSES.includes(harness)) {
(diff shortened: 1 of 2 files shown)


rnwolfe (commented):
## Code Review

Overall this is a clean, well-structured feature. The sentinel-based idempotency design is sound and the test coverage is thorough. One blocking correctness bug in the help-text example, a couple of non-blocking issues.

---

### BLOCKING — Mismatched harness in the CI help-text example

**File:** `packages/engram-cli/src/commands/companion.ts`, lines 45–47

```sh
# Idempotent CI setup (only append if not already present)
engram companion --check --file CLAUDE.md \
  || engram companion --harness claude-code >> CLAUDE.md
```

This example is broken: `--check` defaults to `--harness generic` (sentinel `<!-- engram-companion:generic -->`), but the write step appends `--harness claude-code` content (sentinel `<!-- engram-companion:claude-code -->`). On the second run, `--check` still looks for the generic sentinel, finds it absent, and appends the claude-code block again — the command is never idempotent.

The fix is to pass matching `--harness` flags to both sides:

```sh
engram companion --check --harness claude-code --file CLAUDE.md \
  || engram companion --harness claude-code >> CLAUDE.md
```

The same mismatch exists in the PR description's "Usage" section.

Note: the idempotency acceptance criterion and test (`appending twice then --check exits 0`) use a *consistent* harness, so they pass — but they don't exercise the advertised CI one-liner pattern.

---

### Non-blocking — `--file` without `--check` is silently ignored

**File:** `packages/engram-cli/src/commands/companion.ts`, line 62

When `--file` is passed without `--check`, the option is accepted but has no effect — the command just writes to stdout as usual. A developer who writes `engram companion --file CLAUDE.md` expecting a write-in-place behavior gets surprising silent output to stdout instead.

Consider printing a warning (`--file has no effect without --check`) or rejecting the combination with an error. At minimum, the option description could be tightened: `"file path to inspect (only meaningful with --check)"`.

---

### Non-blocking — Temp directories are not cleaned up in tests

**File:** `packages/engram-cli/test/companion.test.ts`, lines 48–55

`tmpFile()` creates a directory in `os.tmpdir()` but the test suite never removes it. Over many test runs this accumulates temp dirs. Add a teardown registry or use `afterAll` / `afterEach` with `rmSync` to clean up:

```ts
import { rmSync } from "node:fs";
const dirsToClean: string[] = [];
afterAll(() => { for (const d of dirsToClean) rmSync(d, { recursive: true, force: true }); });
```

---

### Nit — sentinel format doesn't include `--db` path context

When `--check` exits 1 for a missing file, there is no diagnostic message. The harness-mismatch scenario (wrong sentinel in the file) is indistinguishable from "file doesn't exist" at exit code level alone. Not blocking, but a `--verbose` or stderr hint could improve debuggability in CI pipelines.

---

### What's good

- The sentinel design (`<!-- engram-companion:<harness> -->`) is correct and collision-resistant across harnesses.
- The harness-specificity acceptance criterion and test (`cursor sentinel not detected as claude-code`) directly closes the cross-contamination case.
- `--check` produces no stdout output (pure exit-code signal) — correct for use in `||` chains.
- `process.exit` override pattern in tests is consistent and correct; `exitOverride()` on the Commander program is also present.
- The `companionSentinel` helper is exported and reused across impl and tests — no duplication.

---

**Summary:** 1 blocking, 2 non-blocking, 1 nit. The blocking item is the mismatched harness in the documented CI one-liner — the feature itself works correctly when harnesses match.

rnwolfe:
Fixed blocking finding: added `--harness claude-code` to the `--check` side of the idempotent CI example so both sides use the same harness sentinel.

PR #208, opened 2026-04-19
feat(ingest): cross-source reference edge resolver

rnwolfe:
## Summary

- Implements a pattern-registry-based cross-source reference resolver (`packages/engram-core/src/ingest/cross-ref/`) that scans episode content for references to entities from other adapters and emits `references` edges
- Adds `RELATION_TYPES` vocab registry with canonical string constants for all edge relation types
- Wires `resolveReferences()` into `ingestGitRepo()` and `GitHubAdapter.enrich()`, removing the now-redundant inline SHA/`#N` detection from the GitHub adapter
- Adds `engram reconcile --cross-refs` flag to drain `unresolved_refs` and re-scan all episodes
- Built-in patterns: full GitHub PR/issue URLs, Gerrit CL URLs, Google Docs URLs, Linear/Jira issue URLs, `b/N` (Buganizer), `go/cl/N` (Gerrit shorthand), full 40-char SHAs (0.9 confidence), short SHAs (0.75 confidence), repo-scoped `#N` refs
- Plugin-safe `PluginReferencePattern` format uses string patterns + `$1` templates (no closures)
- Unresolved references stored in `unresolved_refs` table (idempotent `CREATE TABLE IF NOT EXISTS`); resolved when matching episode arrives
- 19 unit tests covering all acceptance criteria

## Acceptance Criteria

- [x] `resolveReferences(graph, episodeIds, patterns)` implemented in `packages/engram-core/src/ingest/cross-ref/resolver.ts`; returns `{ edgesCreated, unresolved }`
- [x] Built-in pattern registry covers all specified source types
- [x] Each pattern carries a `confidence` field; emitted edges use it
- [x] Ambiguous short SHA matches emit one edge per candidate at `min(pattern.confidence, 0.6)` — see resolver.ts (multiple entity matches per normalizedRef)
- [x] Self-reference guard: pattern match resolving to source episode's primary entity is skipped
- [x] Target resolution goes through entity_evidence → episodes chain (references bind to entities)
- [x] Plugin manifests with `reference_patterns` compiled via `compilePluginPattern()`; collisions raise load error
- [x] `unresolved_refs` table added via idempotent `CREATE TABLE IF NOT EXISTS`; partial index on unresolved rows
- [x] New episode landing drains matching `unresolved_refs` entries via `drainUnresolvedForEpisode()`
- [x] `relation_type: 'references'` added to `RELATION_TYPES` in `packages/engram-core/src/vocab/relation-types.ts`
- [x] `ingestGit()` and `GitHubAdapter.enrich()` call `resolveReferences()` at end of run
- [x] `engram reconcile --cross-refs` re-runs `resolveReferences` and drains `unresolved_refs`
- [x] Dedup: duplicate `(source_entity, target_entity, 'references')` produces one active edge with additional evidence link

## Test plan

- [x] Run `bun test packages/engram-core/test/cross-ref/` — 19 pass
- [x] Run `bun run build` — passes cleanly
- [x] Run `bunx biome check` on all new/changed files — no issues
- [x] Verify pre-existing failures are pre-existing (NullGenerator tests require API key)

### DIFF
--- a/docs/internal/specs/cross-source-references.md
+++ b/docs/internal/specs/cross-source-references.md
@@ -1,0 +1,211 @@
+# Cross-Source Reference Edge Resolution
+
+**Status**: Implemented (v0.1)
+**Location**: `packages/engram-core/src/ingest/cross-ref/`
+
+---
+
+## Overview
+
+Cross-source reference resolution scans episode content for textual references to other
+source objects (commits, PRs, issues, Jira tickets, etc.) and emits `references` edges
+in the knowledge graph. This connects knowledge silos that would otherwise remain
+isolated (e.g., a commit message that cites a GitHub PR, or an issue body that cites a
+commit SHA).
+
+---
+
+## Pattern Registry Shape
+
+Each entry in the pattern registry is a `ReferencePattern`:
+
+```typescript
+interface ReferencePattern {
+  sourceType: string;        // Target episodes.source_type
+  pattern: RegExp;           // Regex, capture group 1 = identifier
+  normalizeRef: (match: string) => string;  // Normalise captured group
+  confidence: number;        // Edge confidence, 0..1
+  _lookupOverride?: (graph: EngramGraph, normalizedRef: string) => { id: string } | null;
+}
+```
+
+Built-in patterns are in `BUILT_IN_PATTERNS` (ordered most-specific first):
+
+| Pattern | source_type | Confidence |
+|---------|-------------|------------|
+| Full GitHub PR URL | `github_pr` | 0.95 |
+| Full GitHub Issue URL | `github_issue` | 0.95 |
+| Full Gerrit CL URL | `gerrit_change` | 0.95 |
+| Full Google Doc URL | `google_doc` | 0.95 |
+| Full Linear Issue URL | `linear_issue` | 0.95 |
+| Full Jira Issue URL | `jira_issue` | 0.95 |
+| `b/NNNNNN` (Buganizer) | `buganizer_issue` | 0.90 |
+| `go/cl/NNN` (Gerrit shorthand) | `gerrit_change` | 0.90 |
+| Full 40-char commit SHA | `git_commit` | 0.90 |
+| Short SHA (7–11 chars) | `git_commit` | 0.75 |
+| `#N` (repo-scoped issue/PR) | `github_issue` | 0.85 |
+
+---
+
+## Resolution Algorithm
+
+```
+for each episodeId in episodeIds:
+  1. Load episode content. Skip if status = 'redacted'.
+  2. Find the primary entity for this episode (highest-confidence entity_evidence link).
+  3. For each pattern in the registry:
+     a. Reset pattern.lastIndex (global flag safety).
+     b. Exec pattern against episode.content.
+     c. For each match:
+        - Normalize captured group → normalizedRef
+        - Deduplicate: skip if (sourceType, normalizedRef) already seen in this episode
+        - Lookup target entity:
+            - If _lookupOverride: call it with (graph, normalizedRef)
+            - Else: findTargetEntityBySourceRef(graph, sourceType, normalizedRef)
+            - If not found: also try with fullMatch (for URL patterns)
+        - If target not found: record in unresolved_refs; continue
+        - Self-reference guard: skip if sourceEntity.id == targetEntity.id
+        - emitReferenceEdge(source, target, episode, confidence, fact, timestamp)
+  4. Call drainUnresolvedForEpisode to resolve any pending refs that match this episode's source_ref.
+```
+
+The lookup chain:
+
+```
+episode.source_ref == normalizedRef  →  find via entity_evidence JOIN episodes
+  └─ fallback: episode.source_ref == fullMatch (capture may be partial)
+```
+
+---
+
+## Ambiguity Handling
+
+When the same normalized ref could match multiple target entities (e.g., a short SHA
+that matches two commit entities), the current implementation resolves to the **first
+match** via `LIMIT 1` SQL queries. This is intentional conservatism for v0.1.
+
+Future work: when multiple entities match a single ref, emit N edges each at
+`confidence * 0.6` to signal reduced certainty. This is tracked in the backlog.
+
+---
+
+## Self-Reference Guard
+
+Before emitting an edge, the resolver checks:
+
+```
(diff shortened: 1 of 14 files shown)


rnwolfe (commented):
## Code Review — PR #208: Cross-Source Reference Edge Resolution

Overall this is well-structured work. The three-layer split (patterns / resolver / schema) is clean, the idempotency story is solid, and the test coverage is unusually thorough for a new subsystem. A handful of issues worth addressing before merge:

---

### 1. Logic error — self-reference guard fires only when `sourceEntity` is non-null (resolver.ts:245)

```ts
// resolver.ts line 245
if (sourceEntity && targetEntity.id === sourceEntity.id) continue;
```

When `sourceEntity` is `null` (the episode has no entity evidence), the guard is skipped and the code falls through to:

```ts
const created = emitReferenceEdge(
  graph,
  sourceEntity?.id ?? targetEntity.id,  // falls back to targetEntity.id
  targetEntity.id,
  ...
);
```

This creates a self-loop edge: `source_id === target_id`. That invariant should be enforced unconditionally. At minimum, guard the `emitReferenceEdge` call:

```ts
// before emitReferenceEdge
if (!sourceEntity) continue; // no provenance, can't emit a meaningful edge
```

Or, if a no-entity source is intentional, add a self-loop guard regardless of whether `sourceEntity` is null.

---

### 2. Logic error — `drainUnresolved` double-counts edges (resolver.ts:405–406)

Phase 1 of `drainUnresolved` resolves existing `unresolved_refs` rows and marks them `resolved_at`. Phase 2 calls `resolveReferences` over **all** episodes, which will re-scan the same episodes that Phase 1 just resolved. Because `emitReferenceEdge` deduplicates by returning `false` for already-existing edges, the `edgesCreated` counter for Phase 2 won't re-fire — but `resolveReferences` will also write **new** `unresolved_refs` rows for every episode that still has no target, then the Phase 2 `scanResult.edgesCreated` is added to Phase 1's `edgesCreated`. In practice this means the returned `edgesCreated` value can over-report when the same edge is created by Phase 1 and then the full re-scan in Phase 2 encounters it again from a different angle. The count shown to the user via `reconcile --cross-refs` will be inflated. Consider returning Phase 2's count separately, or not summing them.

---

### 3. Missing idempotency guard in `ingestIssue` (github.ts)

`ingestPR` has a pre-check for an existing episode and returns early (`counts.episodesSkipped++`) before calling `addEpisode`. `ingestIssue` has **no such pre-check**: it calls `addEpisode` unconditionally, relying on a uniqueness constraint to handle re-ingestion. If the constraint is not enforced at the DB level (or if `addEpisode` throws on collision), re-running GitHub ingestion will fail or create duplicate state. Since the PR removed the old cross-ref logic from `ingestIssue` without restoring the guard, this is a regression risk. Add the same pre-check that `ingestPR` uses.

---

### 4. `RELATION_TYPES` constant introduced but not used in `git.ts` or `github.ts`

`vocab/relation-types.ts` defines typed constants for all relation types, and `resolver.ts` correctly uses `RELATION_TYPES.REFERENCES`. However, `git.ts` (lines 523, 611, 705, 722) and `github.ts` (line 391) still use bare string literals (`"authored_by"`, `"co_changes_with"`, `"likely_owner_of"`, `"reviewed_by"`). This undercuts the value of the vocab module and makes future renaming error-prone. Not a blocker for this PR, but consider a follow-up to thread `RELATION_TYPES` through the existing ingestion code, or note this in a follow-up issue.

---

### 5. Short-SHA pattern can match non-SHA hex strings (patterns.ts:131–136)

```ts
pattern: /\b([0-9a-f]{7,11})\b/gi,
```

This matches any 7–11 char lowercase hex word: CSS color codes (`#1a2b3c` without the `#`, but `\b` would still match `1a2b3c4`), version strings, and other incidental hex. The confidence is already 0.75, which partially mitigates this, but the pattern is notably broader than what was removed from `github.ts` (the old `SHA_RE = /\b([0-9a-f]{7,40})\b/gi`). The comment says "7–11 chars" but the original code allowed up to 40. Confirm the upper bound of 11 is intentional and document the rationale in a comment, especially since commit SHAs stored by git are typically 7 or 40.

---

### 6. Plugin pattern `normalize_template` only supports `$1` (patterns.ts:58–59)

```ts
normalizeRef: (match: string) =>
  manifest.normalize_template.replace("$1", match),
```

Only the first occurrence of `$1` is replaced (`String.replace` with a string needle, not `/$1/g`). If a plugin author uses `$1-$1` as a template, only the first substitution happens. This is minor but worth a comment or using `replaceAll`.

---

### 7. Missing migration — `unresolved_refs` table not in the format migration path

`ensureUnresolvedRefsTable` is called lazily inside `resolveReferences` (and `drainUnresolved`), which means existing databases that open the graph but never call either function won't have the table. That is fine for the current flow. However, the `format/` migration layer presumably applies schema changes on `openGraph`. If `unresolved_refs` is not registered there, a user who opens the graph directly (e.g., via a read-only tool) and then calls `drainUnresolved` for the first time will trigger a DDL statement mid-read, which is benign but inconsistent with how the rest of the schema is managed. Recommend adding `ensureUnresolvedRefsTable` to the `openGraph` initialisation path so the table is always present after open.

---

### 8. Test: short-SHA test is convoluted and partially tests a false path (resolver.test.ts:284–352)

The short-SHA test creates two separate episodes and entities to work around the lookup mismatch, then uses `_edges` (unused variable prefixed with `_`) from the first run without asserting on it. The setup comments acknowledge the workaround ("Short SHA won't match because source_ref is full SHA"). The test still validates the confidence value correctly, but the dead code path (the first `resolveReferences` call on `sourceEp`) adds noise. Consider consolidating to a single episode/entity setup that matches the actual lookup path.

---

### Summary

| Severity | Finding |
|----------|---------|
| Bug | Self-loop edge emitted when `sourceEntity` is null (item 1) |
| Bug | `drainUnresolved` edge count is potentially inflated (item 2) |
| Regression risk | `ingestIssue` missing idempotency pre-check (item 3) |
| Minor | `RELATION_TYPES` not used in `git.ts`/`github.ts` (item 4) |
| Minor | Short-SHA pattern upper bound undocumented (item 5) |
| Minor | Plugin normalize_template only replaces first `$1` (item 6) |
| Convention | `unresolved_refs` not in migration path (item 7) |
| Test quality | Convoluted short-SHA test with dead code path (item 8) |

Issue #206, opened 2026-04-19
feat(core): plugin loader — XDG discovery, manifest parsing, dual transports


rnwolfe:
## Goal

Implement the plugin loader for out-of-repo enrichment adapters per the spec in
`docs/internal/specs/plugin-loading.md` (authored by #204). The loader discovers
XDG-resident plugins, parses their JSON manifests, and loads them via either
in-process (`js-module`) or subprocess (`executable`) transport.

### Discovery

- `packages/engram-core/src/plugins/discover.ts` — enumerates plugin directories:
  - `$XDG_DATA_HOME/engram/plugins/<name>/` (fallback: `~/.local/share/engram/plugins/`)
  - `<project>/.engram/plugins/<name>/` (higher precedence on name collision)
  - Windows: `%LOCALAPPDATA%\engram\plugins\<name>\`
- Returns discovered plugin directories sorted by precedence.

### Manifest parsing

- `packages/engram-core/src/plugins/manifest.ts` — JSON schema validation.
- Rejects with a clear error:
  - Missing required fields
  - Major `contract_version` mismatch
  - Vocab extension collisions (built-in or cross-plugin)
  - `transport` not in `{'js-module', 'executable'}`
  - `entry` path outside the plugin directory

### Transport 1 — js-module (in-process)

- `packages/engram-core/src/plugins/transport/js-module.ts`
- Dynamic `await import(<manifest-dir>/<entry>)`.
- Validates the default export against the `EnrichmentAdapter` shape from #200.
- Plugin runs in-process; errors propagate directly.

### Transport 2 — executable (subprocess)

- `packages/engram-core/src/plugins/transport/executable.ts`
- `spawn`s the plugin with no arguments.
- JSON-lines protocol over stdio per the spec:
  - Send `{op: 'hello', contract_version}`, await `{type: 'hello_ack', capabilities}`,
    verify capabilities match the manifest.
  - Send `{op: 'enrich', scope, auth, since, cursor, dry_run}`.
  - Stream records until `done` or `error`.
  - Engram owns all writes — plugin emits records, core calls
    `addEpisode` / `addEntity` / `addEdge` with the usual evidence-first enforcement.
- Stderr captured and logged at debug level per-line.
- Protocol violations (malformed JSON, unexpected record order, missing `done`,
  process crash) → `EnrichmentAdapterError { code: 'data_error' }`.

### Vocab registry extension

- Merge each loaded plugin's `vocab_extensions` into the in-memory registry maps
[...]

Issue #200, opened 2026-04-18
feat(core): EnrichmentAdapter contract v2 — auth credentials, scope generality, cursor helpers


rnwolfe:
## Goal

Evolve the `EnrichmentAdapter` contract (`packages/engram-core/src/ingest/adapter.ts`)
to handle the auth and scope shapes that Gerrit, Google Docs/Drive, Jira, and Buganizer
require. The v1 contract was GitHub-shaped; adapters are about to stress it.

## Problems with v1

1. **`opts.token: string` is too thin.** Jira needs email + API token (Basic auth).
   Gerrit needs username + HTTP password. Google Docs/Drive needs OAuth2 Bearer +
   declared scopes + refresh capability. Buganizer internal may need service account
   JSON or LOAS.
2. **`opts.repo` is overloaded.** Used as `owner/repo` (GitHub), project name (Gerrit),
   doc ID list (Docs), folder ID (Drive), team key (Linear), project key (Jira),
   component ID (Buganizer). Same field, very different semantics.
3. **Cursor conventions are uncodified.** Schema allows opaque string
   (`format/schema.ts:150`), but `github.ts` does `parseInt` on it; Google Docs will
   need ISO8601 `modifiedTime`. Each adapter reinvents.

### AuthCredential union

```ts
export type AuthCredential =
  | { kind: 'none' }
  | { kind: 'bearer'; token: string }
  | { kind: 'basic'; username: string; secret: string }
  | { kind: 'service_account'; keyJson: string }
  | {
      kind: 'oauth2';
      token: string;
      scopes: string[];
      /**
       * Optional refresh callback. When the adapter receives a 401/403 on a
       * bearer-authenticated request and `refresh` is present, it MUST call
       * `refresh()` once, swap the returned token in, and retry. If the retry
       * also fails, throw `auth_failure`. Adapters without a refresh callback
       * surface `auth_failure` on the first failure.
       */
      refresh?: () => Promise<string>;
    };
```

Adapters declare `supportedAuth: AuthCredential['kind'][]`. The CLI constructs the
credential from flags/env and passes it as `opts.auth`. If the CLI supplies a credential
whose `kind` is not in `supportedAuth`, the adapter throws `auth_failure` with an
actionable message before issuing any network call.

### Rename `opts.repo` → `opts.scope`, add scope schemas

```ts
opts.scope: string;  // adapter-specific; shape documented per adapter
```

Each adapter exports a `scopeSchema` (description + validator) that the CLI uses to
[...]

Issue #199, opened 2026-04-18
chore(core): formalize vocabulary registries (entity_type, source_type, relation_type)


rnwolfe:
## Goal

Create centralized TypeScript registries for `entity_type`, `source_type`, and
`relation_type` values. All ingesters and adapters import from the registry rather than
inlining string literals.

## Problem

String values are scattered across the codebase with no single source of truth:

- `entity_type` values used today: `person`, `service`, `module`, `issue`, `commit`,
  `pull_request` — hardcoded per ingester.
- `source_type` values: `git`, `git_commit`, `github_pr`, `github_issue`, `manual`,
  `document`, `source`, `github`. **Note the asymmetry** — `ingestion_runs.source_type`
  uses `"github"` while episodes use `"github_pr"`/`"github_issue"`. Invisible until
  the cross-source ref resolver (#197) needed to match against these values.
- `relation_type` values: `authored_by`, `reviewed_by`, `co_changes_with`,
  `likely_owner_of`, `references`, `depends_on`, `calls`, `uses`, `maintains`, etc.

With Gerrit, Buganizer, Google Docs, GitLab, Jira, Linear adapters landing, the problem
compounds. A typo (`gerrit_change` vs `gerrit-change`) silently breaks cross-ref
resolution and nothing catches it.

**Scale of refactor:** 24 string-literal sites across 5 files in `src/ingest/`
(`git.ts`, `adapters/github.ts`, `source/index.ts`, `markdown.ts`, `text.ts`) —
bounded.

### Type surface

Each registry follows a uniform pattern — `const` object + derived union type:

```ts
// packages/engram-core/src/vocab/entity-types.ts
export const ENTITY_TYPES = {
  PERSON: 'person',
  MODULE: 'module',
  SERVICE: 'service',
  ISSUE: 'issue',
  COMMIT: 'commit',
  PULL_REQUEST: 'pull_request',
  // ...
} as const;

export type EntityType = typeof ENTITY_TYPES[keyof typeof ENTITY_TYPES];
```

Gives both a runtime constant (for writes and iteration) and a compile-time union (for
function signatures). Matches the `AuthCredential` union style introduced in #200.

### Enforcement layer — runtime only, no SQL CHECK

The SQLite schema stays permissive (`TEXT NOT NULL` as today). Enforcement happens in
two places:

1. **Compile time:** ingester and adapter code imports from the registry; TS rejects
   unknown literals at the type boundary.
[...]

PR #212, opened 2026-04-19
docs: plugin loading architecture spec and ADR-006 (#204)

### LINKED ISSUE #204: spec: plugin loading architecture for out-of-repo EnrichmentAdapters
## Goal

Land the decision on how engram loads out-of-repo enrichment adapters, record it as
ADR-006, and author the reference spec. Loader implementation tracked separately in
#206.

### Discovery — XDG-compliant

- `$XDG_DATA_HOME/engram/plugins/<name>/` (fallback `~/.local/share/engram/plugins/`)
- `<project>/.engram/plugins/<name>/` (higher precedence)
- Each directory contains `manifest.json` + plugin code.
- Windows: `%LOCALAPPDATA%\engram\plugins\<name>\`.

### Manifest — JSON

Declares: `name`, `version`, `contract_version`, `transport` (`js-module` |
`executable`), `entry`, `capabilities` (supported_auth, supports_cursor, scope schema),
and `vocab_extensions` (entity_types, source_types with ingestion/episode split,
relation_types).

### Transport 1 — js-module (in-process)

`await import(entry)`; default export is an `EnrichmentAdapter`-shaped object
conforming to the #200 v2 contract. Zero serialization overhead. TS/JS only.

### Transport 2 — executable (subprocess)

Engram `spawn`s the executable; communication is **JSON-lines** over stdio.

- Engram → plugin: `{op: 'hello', contract_version}`,
  `{op: 'enrich', scope, auth, since, cursor, dry_run}`
- Plugin → engram (streamed):
  - `{type: 'hello_ack', capabilities}` (echoes manifest for sanity)
  - `{type: 'episode', episode: {...}}`
  - `{type: 'entity', entity: {...}, evidence: [...]}`
  - `{type: 'edge', edge: {...}, evidence: [...]}`
  - `{type: 'progress', phase, fetched, created, skipped}`
  - `{type: 'error', code, message}`
  - `{type: 'done', cursor}`
- **Engram owns all writes.** Plugin never touches SQLite; it emits structured
  records, and the graph layer performs writes with full evidence-chain enforcement.
- Stderr captured for diagnostic logging.
- Protocol v1 deliberately minimal. Can evolve (e.g. JSON-RPC) later if call/response
  shapes become necessary (e.g. plugin-side queries into the graph).

### Trust model

**Install-time authorization only.** No signing, no sandbox. If you put it in the
plugins directory, engram runs it. Same trust boundary as the CLI itself — matches
how similar tools (direnv, git hooks, shell rc files) are trusted. Aligns with the
local-first thesis.

### Vocab extension policy

Plugins declare vocab extensions in `manifest.json` under `vocab_extensions`. Engram
merges them into the in-memory registry at load time.

- Additive and declarative — no runtime stringing.
- Collision detection across built-in + all loaded plugins; collisions cause the
  plugin to fail to load with a clear error.
- Declared values are recognized by `verifyGraph` strict mode.
- Built-in registry from #199 remains closed; the manifest is the extension mechanism.

## Out of scope

- **Loader implementation** — tracked in #206. This issue produces the spec + ADR
  only.
- **Plugin authoring guide / SDK.** Follow-up once at least one out-of-repo plugin
  exists.
- **Plugin marketplace or auto-update.** Not in scope; plugins are installed manually.
- **Cross-platform edge cases** (Windows exec bits, path separators beyond XDG). Addressed
  in #206 implementation, not in this spec.

## Resolved open questions

- **Vocab declarations:** manifest-declared, additive, no runtime stringing. ✅
- **Auth in `sync` config (#203):** each plugin declares `supported_auth` in the
  manifest; `engram sync` reads manifests to drive CLI auth prompting. ✅
- **Versioning:** `manifest.contract_version` must match the running engram's
  contract version. Major mismatch refuses to load; minor mismatch warns. ✅
- **Dry-run / progress across boundary:** dry-run is a field on the `enrich` op; the
  plugin still emits records but `done.cursor` is null and engram discards the
  records. Progress is a `progress` record in the stream. ✅
- **Signing / trust:** install-time only; no signing. ✅

## Acceptance criteria

- [ ] `docs/internal/specs/plugin-loading.md` authored. Covers: XDG discovery
      algorithm, manifest schema (with JSON-schema appendix), js-module transport
      semantics, executable transport + full JSON-lines message catalog, vocab
      extension merge algorithm, trust model, version compatibility rules, error
      handling.
- [ ] `docs/internal/DECISIONS.md` — **ADR-006** added. Records: decision (hybrid,
      XDG-discovered, manifest-declared), rejected alternatives (npm, subprocess-only,
      manifest-less file drop), sequencing (#204 before #206), trust model
      justification.
- [ ] Spec includes a canonical example: minimum-viable Python plugin
      (`manifest.json` + `gerrit-adapter.py`) demonstrating the executable transport.
- [ ] #199 cross-referenced: plugin vocab extensions are the v0.1 escape hatch;
      built-in registry remains closed (see updated #199 body).
- [ ] #200 reviewed against the spec: `AuthCredential`, `scopeSchema`, and cursor
      helpers must be expressible in JSON manifest form and serializable across the
      subprocess boundary.

## Deliverables

- `docs/internal/specs/plugin-loading.md` (new)
- `docs/internal/DECISIONS.md` — ADR-006 entry (amended)
- Cross-reference edits to #199 and #200 bodies (non-code, tracked in those issues)

## Blocks

- **#206** — loader implementation
- **#200** — sequencing: #204 now *informs* #200's contract shape (was parallel)

## Related

- **#199** — vocabulary registries; plugin vocab extensions are the escape hatch
  resolved here
- **#203** — `engram sync` consumes plugin manifests for auth/scope prompting
- **#205** — CLI v2 consumer; surfaces plugin-sourced errors to users
- **#123** — harness lifecycle plugins (different concept; unrelated)

rnwolfe:
## Summary

- Adds `docs/internal/specs/plugin-loading.md` — comprehensive spec for the hybrid local-plugin loader
- Adds ADR-006 to `docs/internal/DECISIONS.md` recording the design decision, rejected alternatives, trust model justification, and sequencing rationale

## What's in the spec

- **XDG discovery algorithm** — Linux/macOS `$XDG_DATA_HOME/engram/plugins/` (fallback `~/.local/share/`), project-local `.engram/plugins/` (higher precedence), Windows `%LOCALAPPDATA%\engram\plugins\`
- **Manifest schema** — JSON with JSON Schema appendix covering `name`, `version`, `contract_version`, `transport`, `entry`, `capabilities` (`supported_auth`, `supports_cursor`, `scope_schema`), and `vocab_extensions`
- **js-module transport** — in-process `await import(entry)`, default export conforming to `EnrichmentAdapter`
- **Executable transport + full JSON-lines message catalog** — `hello`/`hello_ack`, `enrich`, `episode`, `entity`, `edge`, `progress`, `error`, `done`; Engram owns all SQLite writes
- **Vocab extension merge algorithm** — additive-only, collision detection, namespace convention, extension limits
- **Trust model** — install-time authorization, no signing (same rationale as git hooks / direnv), with user security guidance
- **Version compatibility rules** — major mismatch refuses to load, older contract warns
- **Error handling** — load-time and runtime error tables for both transports
- **Canonical example** — minimum-viable Python plugin (`manifest.json` + `gerrit-adapter.py`) using executable transport, stdlib only

## Cross-references

- Vocabulary registry: #199 / `docs/internal/specs/vocabulary.md`
- `AuthCredential` and `scopeSchema` stabilization: #200
- Implementation: #206

## Acceptance Criteria

- [x] `docs/internal/specs/plugin-loading.md` covers XDG discovery algorithm
- [x] Manifest schema with JSON Schema appendix
- [x] js-module transport semantics
- [x] Executable transport + full JSON-lines message catalog
- [x] Vocab extension merge algorithm
- [x] Trust model and justification
- [x] Version compatibility rules
- [x] Error handling tables for both transports
- [x] Canonical Python plugin example (manifest.json + gerrit-adapter.py)
- [x] ADR-006 added to DECISIONS.md with decision, rejected alternatives, sequencing, trust model justification
- [x] Cross-references to #199 and #200 included

Closes #204

### DIFF
--- a/docs/internal/DECISIONS.md
+++ b/docs/internal/DECISIONS.md
@@ -75,3 +75,97 @@ But Karpathy's wiki and graphify both lack two things Engram already has: a temp
 
 *Carried over*
 - All existing principles, schema, operations, and the AI provider layer (#31) remain unchanged. Projection work is gated on #31 shipping.
+
+## ADR-006 -- Hybrid local-plugin loader with XDG discovery and manifest declaration
+
+**Date**: 2026-04-19
+**Status**: Accepted
+**Spec**: [`docs/internal/specs/plugin-loading.md`](specs/plugin-loading.md)
+**Sequencing**: Spec (#204) before implementation (#206).
+
+**Context**: Engram's built-in enrichment adapters (GitHub, Gerrit) are compiled into
+`engram-core`. Community integrations (Jira, GitLab, Phabricator, custom internal systems)
+cannot be shipped this way — the dependency surface would be unbounded and the release cycle
+would become a bottleneck for adopters.
+
+Three requirements shaped the design:
+1. **Any-language plugins** — adopters should not be forced to write TypeScript. Internal
+   tooling at many organizations is Python, Go, or shell.
+2. **Local-first / no registry** — consistent with Engram's core value. No package manager
+   integration, no network fetch at runtime, no centralized trust authority.
+3. **Engram owns the graph** — plugins must not bypass the evidence-first invariant by writing
+   directly to SQLite. All writes must flow through core's validated write path.
+
+**Decision**: Adopt a hybrid local-plugin loader with XDG-standard discovery and explicit
+manifest declaration.
+
+The loader searches two locations in order of increasing precedence:
+- `$XDG_DATA_HOME/engram/plugins/<name>/` (global user; Linux/macOS; `%LOCALAPPDATA%\engram\plugins\<name>\` on Windows)
+- `<project>/.engram/plugins/<name>/` (project-local; higher precedence; committed alongside the codebase)
+
+Each plugin directory contains a `manifest.json` declaring name, version, `contract_version`,
+transport, entry point, capabilities, and optional vocab extensions.
+
+Two transport mechanisms are supported:
+- **js-module**: in-process `await import(entry)`, default export conforms to `EnrichmentAdapter`
+  interface. Full graph API access. TypeScript/JavaScript only.
+- **executable**: any-language subprocess. JSON-lines protocol over stdin/stdout. Engram owns
+  all SQLite writes; plugin only emits record descriptions.
+
+Trust model: install-time user authorization only, same as git hooks and direnv. No signing.
+No sandboxing.
+
+**Alternatives considered**:
+
+- *npm-based plugin registry (e.g. `engram plugin install @scope/adapter`).* Rejected:
+  introduces `npm`/`npx` as a runtime dependency, pulls in remote code automatically, and
+  creates a centralized trust surface inconsistent with local-first principles. Also forces
+  TypeScript on plugin authors.
+
+- *Subprocess-only (no js-module transport).* Rejected: TypeScript adapters that already
+  implement `EnrichmentAdapter` (e.g. a GitHub Enterprise variant) would need to be wrapped
+  in a subprocess harness unnecessarily. In-process loading is strictly more capable for the
+  TypeScript ecosystem; JSON-lines transport is reserved for cross-language interop.
+
+- *Manifest-less file drop (auto-discover any `.js` or executable in a plugins dir).*
+  Rejected: ambiguous entry point resolution, no contract version negotiation, no capability
+  declaration, no structured error for incompatible plugins. The manifest is the minimum
+  necessary metadata to make the loader safe and debuggable.
+
+- *Signed plugin bundles.* Rejected: the threat model for a local-first tool running as the
+  user's own process does not justify the implementation cost. An attacker with write access
+  to `~/.local/share/engram/plugins/` already has user-level code execution via git hooks,
+  `.profile`, or `PATH` manipulation. Signing provides no meaningful defense in this model.
+
+**Trust model justification**: The same rationale that the Git project and direnv use applies
+here. Users who run `git clone` execute arbitrary hooks; users who `cd` into a project with
+`.envrc` execute arbitrary shell. Engram plugins are no different in threat model — the user
+has deliberately installed them. The documentation (§6.3 of the spec) provides security
+guidance for auditing plugin source before installation.
+
+**Sequencing rationale**: The spec (#204) is authored before implementation (#206) so that
+the JSON-lines message catalog, manifest schema, and trust policy are reviewed and stable
+before code is written. The vocab extension mechanism depends on the vocabulary registry
+work (#199 / `docs/internal/specs/vocabulary.md`); the executable transport `auth` field
+shape depends on `AuthCredential` stabilization (#200).
+
+**Consequences**:
+
+*Positive*
+- Any-language plugin authorship (the Python example in the spec is dependency-free stdlib).
+- Engram's evidence-first invariant is preserved for all transports — the loader, not the
+  plugin, calls `addEntity`, `addEdge`, `addEpisode`.
+- XDG paths are conventional and toolable — package managers and dotfile managers can place
+  plugins without engram-specific tooling.
+- Project-local plugins are version-controlled alongside the codebase they serve.
+- `contract_version` negotiation prevents silent incompatibility as the protocol evolves.
+
+*Negative / cost*
+- Executable transport has process-spawn overhead per `enrich()` call. Acceptable for batch
+  ingestion (called infrequently); unacceptable for per-query enrichment. The spec explicitly
+  limits the pattern to batch ingest.
+- No sandboxing means a malicious or buggy js-module plugin can read/write arbitrary process
+  state. Documented risk; mitigation is user education and code review, not a technical fence.
+- Vocab extensions are additive and runtime-only — they do not generate TypeScript types.
+  Downstream callers that need compile-time safety must import built-in types from
+  `packages/engram-core/src/vocab/` and handle plugin-extended values as opaque strings.
--- a/docs/internal/specs/plugin-loading.md
+++ b/docs/internal/specs/plugin-loading.md
@@ -300,10 +300,10 @@ Messages are written to the plugin's stdin, one per line.
 | Field | Type | Description |
 |---|---|---|
 | `scope` | `object` | Plugin-defined scope object (validated against `scope_schema` if declared). |
-| `auth` | `object \| null` | Auth credential. Shape depends on auth type: `{"type":"token","token":"..."}` or `{"type":"oauth","token":"..."}` or `null`. |
+| `auth` | `object \| null` | Auth credential. Shape depends on auth type: `{"type":"token","token":"..."}` or `{"type":"oauth","token":"..."}` or `null`. This shape is an interim wire representation; issue #200 will stabilize the `AuthCredential` union type. Until #200 lands, the loader maps a `token: string` from `EnrichOpts` to `{"type":"bearer","token":"..."}` before serialising to the subprocess. Plugins should treat the `type` field as an enum hint and tolerate both `"token"` and `"bearer"` spellings. |
 | `since` | `string \| null` | ISO8601 UTC — only fetch items updated after this date. `null` for full backfill. |
 | `cursor` | `string \| null` | Opaque resume cursor from the previous run. `null` on first run. |
-| `dry_run` | `boolean` | When `true`, plugin SHOULD skip side effects and emit records as if they would be written. |
+| `dry_run` | `boolean` | When `true`, plugin MUST skip all side effects (no external writes) and MUST emit `{"type":"done","cursor":null,...}` as its final message. Counters in `done` should reflect what would have been written. |
 
 #### 3.2.2 Plugin → Engram messages
 
@@ -316,12 +316,20 @@ echoes capabilities back to Engram.
 ```json
 {
   "type": "hello_ack",
+  "contract_version": 1,
   "capabilities": {
     "supported_auth": ["token"],
     "supports_cursor": true
   }
 }
 ```
+
+The `contract_version` field MUST echo the value received in the `hello` message. For
+executable transport, Engram performs the authoritative contract compatibility check against
+`manifest.contract_version` at load time (see §5.1); the `hello_ack` echo provides a
+runtime cross-check that the running process matches the manifest. The loader emits a
+`warn`-level log if the echoed value differs from the manifest declaration but does not
+abort the run.
 
 **`episode`** — emit a new episode (raw evidence). Engram deduplicates on `(source_type, source_ref)`.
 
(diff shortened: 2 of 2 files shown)


rnwolfe (commented):
## Review — Plugin Loading Architecture Spec and ADR-006

Reviewed against: issue #204 acceptance criteria, `packages/engram-core/src/ingest/adapter.ts`, `packages/engram-core/src/vocab/`, and `docs/internal/specs/vocabulary.md`.

---

### Findings

#### BLOCKING — 3 issues

---

**[BLOCKING-1] `EnrichmentAdapter` interface in spec does not match the real interface in `adapter.ts`**

Section 3.1 quotes the `EnrichmentAdapter` interface as:
```typescript
interface EnrichmentAdapter {
  name: string;
  kind: string;
  supportsAuth?: string[];
  supportsCursor?: boolean;
  enrich(graph: EngramGraph, opts: EnrichOpts): Promise<IngestResult>;
}
```

The actual interface in `packages/engram-core/src/ingest/adapter.ts` uses different field names:
- `supportsAuth` → actual field is `supportsAuth` ✓ (matches)
- `supportsCursor` → actual field is `supportsCursor` ✓ (matches)

However: the `EnrichOpts` type does **not** have a structured `auth` field. It has `token?: string`. The spec's executable transport `enrich` message carries `{"auth": {"type": "token", "token": "..."}}`, which is the structured form from the forthcoming #200 stabilization. The loader bridge — the code that would translate the JSON-lines `auth` object into the flat `token` string expected by `EnrichOpts` — is not addressed anywhere in the spec.

This gap matters because the js-module transport section says the adapter uses `EnrichOpts` directly (no translation layer mentioned), while the executable transport's `auth` wire format is a structured object. A js-module plugin author reading §3.1 alongside §3.2 will not know that the in-process `opts.token` and the subprocess `auth.token` are the same credential passed through different shapes. The spec needs a note explicitly calling out this impedance mismatch and stating that the loader is responsible for the translation — or that #200 will change `EnrichOpts` to accept the structured form, at which point both transports converge.

---

**[BLOCKING-2] Vocab extension example in the Python plugin uses built-in values not declared in `vocab_extensions`**

In §8.3 (`gerrit-adapter.py`), the plugin emits entities and edges using these type values:
- `entity_type: "person"` — this is `ENTITY_TYPES.PERSON`, a built-in value ✓
- `entity_type: "pull_request"` — this is `ENTITY_TYPES.PULL_REQUEST`, a built-in value ✓
- `relation_type: "authored_by"` — this is `RELATION_TYPES.AUTHORED_BY`, a built-in value ✓
- `relation_type: "reviewed_by"` — this is `RELATION_TYPES.REVIEWED_BY`, a built-in value ✓

All of these are fine to use as built-ins. **But the `manifest.json` in §8.2 does not declare `relation_types` in `vocab_extensions`**, yet the plugin emits `"reviewed_by"` which, while built-in, is also the exact value mentioned in the issue body as a Gerrit-specific extension (the issue suggests `gerrit_reviewed_by`). More critically, if the intent is to show how vocab extensions work, the example uses only built-in types and never actually demonstrates a non-built-in vocab extension in action.

The bigger bug: the Python example in §8.3 emits `source_type: "my-gerrit-adapter/gerrit_change"` — a namespaced extension value — and the manifest in §8.2 correctly declares it under `vocab_extensions.source_types.episode`. But then the `ingestion` source type used in `ingestion_runs` is `"my-gerrit-adapter/gerrit"`, also declared in the manifest. However, **the Python code itself never emits an `ingestion` source type** — that is stored by Engram in `ingestion_runs`, not by the plugin. So the manifest declaration of the ingestion source type is correct, but the example gives no indication of how/when Engram uses that value. A comment clarifying "Engram (not the plugin) writes this to `ingestion_runs.source_type` when starting the run" would prevent confusion.

The example also never demonstrates the collision detection behavior described in §4.2 — if the example is the canonical teaching artifact, it should use at least one clearly-namespaced custom type (e.g. `entity_type: "my-gerrit-adapter/gerrit_change"` instead of reusing `pull_request`) to model the recommended namespace convention from §4.1.

---

**[BLOCKING-3] `hello_ack` / `hello` contract version cross-check behavior is underspecified**

§3.2.3 lifecycle shows Engram validates `contract_version` compatibility after receiving `hello_ack`, but the spec does not define what Engram does if the plugin's `hello_ack` capabilities differ from what's in the manifest. §5 (Version Compatibility) says discrepancies emit a `warn` but do not abort — but §3.2.3 says "Engram validates contract_version compatibility" after hello_ack without specifying the behavior when the executable announces `contract_version` in `hello_ack`.

More concretely: the `hello_ack` message schema (§3.2.2) does not include a `contract_version` field — only `capabilities`. But §5.1 says "For executable transport, the loader also cross-checks the hello_ack capabilities against the manifest declaration." This cross-check is between capabilities, not contract_version. Yet §3.2.3 says Engram "validates contract_version compatibility" at the hello_ack step. These two descriptions are inconsistent. If the executable cannot assert its own contract_version in `hello_ack` (because the field isn't in the schema), then the contract_version check for executable transport is purely manifest-driven, which should be stated explicitly. If the intent is for the executable to echo back its contract_version in `hello_ack`, the field needs to be added to the `hello_ack` schema.

---

#### ADVISORY — 4 issues

---

**[ADVISORY-1] ADR-006 says "two locations" but the spec describes "three locations"**

ADR-006 in DECISIONS.md states:
> "The loader searches **two** locations in order of increasing precedence"

But §1 of the spec says:
> "The plugin loader searches **three** locations in order of increasing precedence"

Reading §1 more carefully, the "three" in the spec headline collapses to two distinct paths (global user and project-local) with the global user path split only by platform variant (Linux/macOS vs Windows). The ADR's "two locations" is therefore correct in substance but the spec's "three" headline is misleading and inconsistent with the ADR. Either the spec should open with "two locations (with platform variants)" or the ADR should say "three locations." As-is, a reader comparing the two documents will be confused.

---

**[ADVISORY-2] Collision detection policy for non-fatal collisions conflicts with the issue body**

§4.2 says that a plugin declaring a value that already exists in the built-in registry emits a `warn` and silently skips the colliding value — **the plugin is still loaded**. The issue body's "Resolved open questions" section says collisions cause the plugin to "fail to load with a clear error." These are directly contradictory policies. If the spec intentionally softened this to a skip-and-warn (reasonable, since a plugin emitting `entity_type: "person"` should just work), the ADR or spec should call out the deliberate change from the issue decision.

---

**[ADVISORY-3] Cursor handling on `dry_run` differs between spec and issue**

The issue body says: "dry-run is a field on the `enrich` op; the plugin still emits records but `done.cursor` is null and engram discards the records."

The spec (§3.2.1) says: `dry_run: false` in the example but never explicitly states that `done.cursor` MUST be null when `dry_run: true`. §7.2 mentions that the old cursor is preserved when the run fails, but dry-run behavior — whether the cursor is updated, preserved, or set to null — is never specified. A MUST/SHOULD statement covering `done.cursor` semantics during `dry_run` is needed.

---

**[ADVISORY-4] `name` pattern in JSON Schema allows single-character names**

The JSON Schema appendix (§2.5) defines the `name` pattern as:
```json
"pattern": "^[a-z0-9][a-z0-9-]*[a-z0-9]$"
```

This regex requires at least two characters (first + last group) with an optional middle, but the `[a-z0-9-]*` middle is zero-or-more, meaning a two-character name like `"ab"` matches but a single-character name like `"a"` does not. However the pattern also fails on exactly two-character names ending in hyphen. More importantly, the pattern does not match a single-character name, which is fine for plugin names but should be noted. A minor cosmetic issue: `"^[a-z0-9]([a-z0-9-]*[a-z0-9])?$"` would be the correct regex for names that can be a single character; the current pattern mandates at least 2. This is probably intentional but worth a comment.

---

### Acceptance criteria coverage

| Criterion | Status |
|---|---|
| XDG discovery algorithm | Covered (§1) |
| Manifest schema with JSON Schema appendix | Covered (§2, §2.5) |
| js-module transport semantics | Covered (§3.1) |
| Executable transport + full JSON-lines message catalog | Covered (§3.2) — but see BLOCKING-3 |
| Vocab extension merge algorithm | Covered (§4) |
| Trust model | Covered (§6) |
| Version compatibility rules | Covered (§5) |
| Error handling | Covered (§7) |
| Canonical Python example | Covered (§8) — but see BLOCKING-2 |
| ADR-006 in DECISIONS.md | Covered — but see ADVISORY-1 |
| Rejected alternatives in ADR-006 | Covered (all four from issue body) |
| Trust model justification in ADR-006 | Covered |
| Sequencing in ADR-006 | Covered |
| Cross-references to #199 and #200 | Covered |

---

### Summary

**3 blocking / 4 advisory.** Not ready to merge as-is.

The spec is substantively complete and well-written. The blocking items are all resolvable with targeted edits rather than structural rewrites:
1. §3.1 needs a note on the `auth` shape impedance mismatch between transports (pending #200).
2. §8.2/§8.3 example should demonstrate a namespaced custom entity type and clarify who writes the ingestion source type.
3. §3.2.2 `hello_ack` schema and §3.2.3/§5.1 prose need to agree on whether executables can/should echo back `contract_version`.

rnwolfe (commented):
Addressed blocking findings in 879c92c:

**Issue 1 — Auth shape impedance mismatch**: Added a note to the `enrich` message field table explaining that `auth` is an interim wire shape, that issue #200 will stabilize the `AuthCredential` union type, and that until #200 lands the loader maps `token: string` from `EnrichOpts` to `{"type":"bearer","token":"..."}` before passing to the subprocess. Plugins are advised to tolerate both `"token"` and `"bearer"` spellings.

**Issue 2 — Python example lacks namespaced extension type**: Replaced `"pull_request"` with `"my-gerrit-adapter/patchset"` (a namespaced custom type) and added inline comments explaining the namespace convention and that the plugin never emits the ingestion `source_type` it declared — Engram writes those `ingestion_runs` rows. Also added `my-gerrit-adapter/patchset` to the §8.2 manifest `vocab_extensions`.

**Issue 3 — `hello_ack` missing `contract_version`**: Added `contract_version` to the `hello_ack` schema (echoes the value received in `hello`). Added a paragraph clarifying that for executable transport the authoritative check happens against `manifest.contract_version` at load time; the `hello_ack` echo provides a runtime cross-check and triggers a `warn` if it differs from the manifest.

**Issue 4 — `dry_run` cursor semantics**: Upgraded SHOULD to MUST. Plugin MUST emit `{"type":"done","cursor":null,...}` when `dry_run` is true.

**Issue 5 — Collision policy**: Tightened from skip-and-warn to fatal for both built-in and cross-plugin collisions. Updated §4.2 and the §7.1 load-time errors table. Added clear error messages for both collision cases and a note that plugins legitimately emitting records of built-in types must NOT declare those types in `vocab_extensions`.

PR #215, opened 2026-04-19
feat(cli): ingest command — consume v2 adapter options (#205)

### LINKED ISSUE #205: feat(cli): ingest command — consume v2 adapter options
## Goal

Update `packages/engram-cli/src/commands/ingest.ts` to consume the v2 `EnrichmentAdapter`
contract from #200: construct `AuthCredential` from flags/env, validate scope via the
adapter-provided `scopeSchema`, and surface targeted error messages for auth failures.

## Context

#200 lands the v2 contract in `engram-core`: `AuthCredential` union, `opts.scope` with
per-adapter `scopeSchema`, and cursor helpers. The legacy v1 compat shim keeps the CLI
working in the interim — this issue removes the CLI's dependence on that shim.

Today, `ingest.ts` passes `{ token, repo }` directly. After this issue, it constructs
`{ auth: AuthCredential, scope: string }` per adapter's declared capabilities.

### Auth construction

A small helper (`packages/engram-cli/src/ingest/auth.ts` or similar) that builds an
`AuthCredential` from CLI flags and environment variables, given the adapter's
`supportedAuth` set:

```
engram ingest enrich <adapter> --scope <value> [auth flags...]
```

Auth flags (select based on adapter's `supportedAuth`):
- `--token <t>` or `$<ADAPTER>_TOKEN` → `{ kind: 'bearer', token }`
- `--username <u> --password <p>` or `$<ADAPTER>_USERNAME` + `$<ADAPTER>_PASSWORD` →
  `{ kind: 'basic', username, secret }`
- `--service-account <path>` or `$<ADAPTER>_SERVICE_ACCOUNT_JSON` →
  `{ kind: 'service_account', keyJson: readFile(path) }`
- `--oauth-token <t> --oauth-scopes <csv>` or env equivalents →
  `{ kind: 'oauth2', token, scopes: csv.split(',') }`. OAuth refresh
  callback wiring is out of scope for v2 (adapters without refresh surface
  `auth_failure` on stale tokens — user re-runs with a fresh token).

If the adapter's `supportedAuth` is `['none']`, no auth flags are required.

If the user provides flags for an auth kind the adapter does not support, the CLI
errors **before** calling `enrich()` with: *"adapter `<name>` does not support
`<kind>` auth. Supported: `<list>`."*

### Scope validation

Before calling `adapter.enrich()`, the CLI calls `adapter.scopeSchema.validate(opts.scope)`.
On validation failure the CLI prints a human-readable error that includes the
schema's `description` and the offending value — no stack trace.

### Error surfacing

When `enrich()` throws `EnrichmentAdapterError`, the CLI switches on `code` and prints
a targeted message:
- `auth_failure` — show supported-auth hints + how to set the relevant env var
- `rate_limited` — show the retry hint from the error message
- `data_error` — print as-is, exit non-zero
- `server_error` — print as-is, exit non-zero

### Removal of legacy shim path

After this issue lands, the CLI no longer passes `opts.token` or `opts.repo` — only
`opts.auth` and `opts.scope`. The v1 compat shim in `engram-core` remains for any
out-of-repo callers but is unreachable from the CLI.

## Acceptance criteria

- [ ] `ingest.ts` constructs `AuthCredential` from flags/env based on the adapter's
      `supportedAuth` set
- [ ] `ingest.ts` calls `scopeSchema.validate()` before `enrich()` and surfaces errors
      without stack traces
- [ ] Flag-vs-adapter mismatch errors before any network I/O with a clear message
      listing supported auth kinds
- [ ] `EnrichmentAdapterError` codes are mapped to distinct CLI error formats with
      actionable remediation
- [ ] GitHub adapter works end-to-end via the new code path with no reliance on the
      v1 compat shim
- [ ] Help text (`engram ingest enrich --help`) documents the per-adapter auth flag
      matrix
- [ ] Integration test: golden-path ingest against a mock adapter for each
      `AuthCredential` kind
- [ ] Integration test: scope validation failure exits with the expected message
- [ ] Integration test: auth-kind-mismatch exits without calling `enrich()`

## Documentation

- [ ] Help text sweep: `engram ingest enrich --help` and per-adapter subcommand help
- [ ] `docs/internal/adapter-authoring.md` (from #200) cross-referenced from CLI docs
- [ ] CLAUDE.md `Key Files` row for `packages/engram-cli/src/commands/ingest.ts` noting
      v2 consumer status (if not already present)

## Blocked by

- **#200** — EnrichmentAdapter contract v2

## Related

Unblocks adapter onboarding once #200 + #199 ship: #191 (Gerrit), #193 (GitLab),
#194 (Jira), #195 (Linear), #196 (Google Docs), #198 (Google Drive).

rnwolfe:
## Summary

- Adds `AuthCredential` union type, `ScopeSchema` interface, and `supportedAuth` field to `packages/engram-core/src/ingest/adapter.ts` (the v2 contract that was described in #200)
- Updates GitHub and Gerrit adapters with `supportedAuth` and `scopeSchema` declarations, with backward-compat shims that still accept `opts.token`/`opts.repo`
- Adds `packages/engram-cli/src/ingest/auth.ts` — `buildAuthCredential()` helper that builds an `AuthCredential` from CLI flags or env vars given the adapter's `supportedAuth` set
- Rewrites `ingest enrich` subcommands to use `--scope` (deprecates `--repo`), typed auth flags, scope validation, and targeted `EnrichmentAdapterError` code messages
- Adds `engram ingest enrich gerrit` as a new enrichment subcommand

## Acceptance Criteria

- [x] `ingest.ts` constructs `AuthCredential` from flags/env based on the adapter's `supportedAuth` set
- [x] `ingest.ts` calls `scopeSchema.validate()` before `enrich()` and surfaces errors without stack traces
- [x] Flag-vs-adapter mismatch errors before any network I/O with a clear message listing supported auth kinds
- [x] `EnrichmentAdapterError` codes are mapped to distinct CLI error formats with actionable remediation
- [x] GitHub adapter works end-to-end via the new code path with no reliance on the v1 compat shim
- [x] Help text (`engram ingest enrich --help`) documents the per-adapter auth flag matrix
- [x] Integration test: golden-path bearer auth — flags provided, scope valid → `enrich()` called with correct AuthCredential
- [x] Integration test: scope validation failure exits with the expected message, `enrich()` never called
- [x] Integration test: auth-kind mismatch exits without calling `enrich()`

## Test plan

- `bun test packages/engram-cli/test/ingest/` — 32 tests, all passing
- Full suite `bun test` — 915 pass, 2 pre-existing failures unrelated to this PR
- `bun run build` — clean
- `bun run lint` — clean

Closes #205

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -144,7 +144,14 @@ This distinction is critical for trust. Never present inferred edges as observed
 
 Two layers:
 1. **VCS layer (universal)**: git commits, blame, co-change analysis. No API tokens needed. Produces the structural graph.
-2. **Enrichment adapters (pluggable)**: GitHub PRs/issues, future Gerrit/Jira/etc. Each implements `EnrichmentAdapter` interface. New adapters must import `entity_type`, `source_type`, and `relation_type` values from `packages/engram-core/src/vocab/`.
+2. **Enrichment adapters (pluggable)**: GitHub PRs/issues, Gerrit, future Jira/etc. Each implements `EnrichmentAdapter` interface (v2). New adapters must import `entity_type`, `source_type`, and `relation_type` values from `packages/engram-core/src/vocab/`.
+
+**Adapter contract v2 — key conventions:**
+- **Auth**: Use `AuthCredential` union (`bearer`, `basic`, `service_account`, `oauth2`, `none`) via `opts.auth`. Declare accepted kinds in `supportedAuth: AuthCredential['kind'][]`. Use `assertAuthKind()` helper before processing.
+- **Scope**: Declare `scopeSchema: ScopeSchema` with a description and `validate()` method. Pass scope via `opts.scope` (replaces deprecated `opts.repo`).
+- **Cursors**: Use `readIsoCursor()`, `readNumericCursor()`, and `writeCursor()` from `packages/engram-core/src/ingest/cursor.ts` — never inline cursor SQL.
+- **Compat shim**: Call `applyCompatShim(opts)` at the start of `enrich()` to automatically map deprecated `opts.token`/`opts.repo` to v2 fields with a one-shot warning.
+- See `docs/internal/specs/adapter-contract.md` for full contract documentation.
 
 Ingestion is idempotent:
 - Episode dedup via `(source_type, source_ref)` unique index
@@ -270,7 +277,9 @@ When creating a PR that implements a GitHub issue:
 | `packages/engram-core/src/ai/kinds/` | Built-in projection kind catalog files (YAML). User overrides via `$XDG_CONFIG_HOME/engram/kinds/` (fallback `~/.config/engram/kinds/`). |
 | `packages/engram-core/src/ai/kinds.ts` | Kind catalog loader — `loadKindCatalog()`, `KindEntry`, `KindCatalog` |
 | `packages/engram-core/src/ingest/git.ts` | Git VCS ingestion (the "money command" engine) |
-| `packages/engram-core/src/ingest/adapter.ts` | EnrichmentAdapter interface |
+| `packages/engram-core/src/ingest/adapter.ts` | EnrichmentAdapter interface (v2) — AuthCredential, ScopeSchema, applyCompatShim, assertAuthKind |
+| `packages/engram-core/src/ingest/cursor.ts` | Cursor helpers — readIsoCursor, readNumericCursor, writeCursor |
+| `docs/internal/specs/adapter-contract.md` | Full adapter contract v2 documentation — auth, scope, cursors, migration |
 | `docs/internal/specs/adapter-aliases.md` | Adapter shorthand alias convention (required for cross-source ref resolution) |
 | `docs/internal/specs/cross-source-references.md` | Cross-source reference resolver architecture |
 | `docs/internal/specs/vocabulary.md` | Controlled vocabulary registries — entity_type, source_type, relation_type |
--- a/docs/internal/DECISIONS.md
+++ b/docs/internal/DECISIONS.md
@@ -75,3 +75,97 @@ But Karpathy's wiki and graphify both lack two things Engram already has: a temp
 
 *Carried over*
 - All existing principles, schema, operations, and the AI provider layer (#31) remain unchanged. Projection work is gated on #31 shipping.
+
+## ADR-006 -- Hybrid local-plugin loader with XDG discovery and manifest declaration
+
+**Date**: 2026-04-19
+**Status**: Accepted
+**Spec**: [`docs/internal/specs/plugin-loading.md`](specs/plugin-loading.md)
+**Sequencing**: Spec (#204) before implementation (#206).
+
+**Context**: Engram's built-in enrichment adapters (GitHub, Gerrit) are compiled into
+`engram-core`. Community integrations (Jira, GitLab, Phabricator, custom internal systems)
+cannot be shipped this way — the dependency surface would be unbounded and the release cycle
+would become a bottleneck for adopters.
+
+Three requirements shaped the design:
+1. **Any-language plugins** — adopters should not be forced to write TypeScript. Internal
+   tooling at many organizations is Python, Go, or shell.
+2. **Local-first / no registry** — consistent with Engram's core value. No package manager
+   integration, no network fetch at runtime, no centralized trust authority.
+3. **Engram owns the graph** — plugins must not bypass the evidence-first invariant by writing
+   directly to SQLite. All writes must flow through core's validated write path.
+
+**Decision**: Adopt a hybrid local-plugin loader with XDG-standard discovery and explicit
+manifest declaration.
+
+The loader searches two locations in order of increasing precedence:
+- `$XDG_DATA_HOME/engram/plugins/<name>/` (global user; Linux/macOS; `%LOCALAPPDATA%\engram\plugins\<name>\` on Windows)
+- `<project>/.engram/plugins/<name>/` (project-local; higher precedence; committed alongside the codebase)
+
+Each plugin directory contains a `manifest.json` declaring name, version, `contract_version`,
+transport, entry point, capabilities, and optional vocab extensions.
+
+Two transport mechanisms are supported:
+- **js-module**: in-process `await import(entry)`, default export conforms to `EnrichmentAdapter`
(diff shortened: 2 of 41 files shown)


rnwolfe (commented):
## Code Review — PR #215 `feat(cli): ingest command — consume v2 adapter options`

Overall the PR is well-structured. The auth dispatch layer, scope validation, and adapter contract updates are correct. I have **two blocking issues** (one security, one correctness) and three non-blocking observations.

---

### BLOCKING — Security: `service_account` path is unsanitized

**File**: `packages/engram-cli/src/ingest/auth.ts`, lines 82-90

```ts
const saPath =
  flags.serviceAccount ?? process.env[`${prefix}_SERVICE_ACCOUNT_JSON`];
if (saPath) {
  let keyJson: string;
  try {
    keyJson = fs.readFileSync(saPath, "utf8");
  } catch (err) { …
```

`saPath` is passed directly to `fs.readFileSync` without sanitization. A path like `../../../../etc/passwd` or an absolute path to a sensitive file will be read and placed into `keyJson`, which is then embedded in the `AuthCredential` that the adapter receives. While adapters are supposed to never write credentials into the graph, there is nothing stopping the raw file content from being logged (e.g. `opts.verbose` log of the auth kind already happens in `ingest.ts`; future adapters may log more). At minimum, `saPath` should be resolved with `path.resolve()` and then validated that it has a `.json` extension, or the caller should document that the restriction is intentional. This is categorised **blocking** because the data flow is: user-controlled input → unchecked path → file read → credential passed to network adapter.

**Suggested fix** (minimal, no behaviour change for valid inputs):

```ts
import * as path from "node:path";
// …
const saPath = path.resolve(
  flags.serviceAccount ?? process.env[`${prefix}_SERVICE_ACCOUNT_JSON`] ?? ""
);
if (!saPath.endsWith(".json")) {
  throw new Error(`service account path must be a .json file, got: ${saPath}`);
}
```

---

### BLOCKING — Correctness: `handleEnrichError` casts `GitHubAuthError` to `EnrichmentAdapterError` unsafely

**File**: `packages/engram-cli/src/commands/ingest.ts`, `handleEnrichError()`

```ts
if (
  err instanceof GitHubAuthError ||
  err instanceof EnrichmentAdapterError
) {
  const adapterErr = err as EnrichmentAdapterError;
  …
  switch (adapterErr.code) {
```

`GitHubAuthError` is caught in the same branch as `EnrichmentAdapterError`, but then immediately cast to `EnrichmentAdapterError` and `.code` is read. If `GitHubAuthError` does not extend `EnrichmentAdapterError` (it appears to be a separate legacy class), then `adapterErr.code` will be `undefined` and the `switch` falls through to `default`, silently losing the auth_failure message. The intent appears to be that `GitHubAuthError` should produce the same `auth_failure` output, but the current structure is fragile.

**Suggested fix**: Either handle `GitHubAuthError` in its own `if` branch before the `instanceof EnrichmentAdapterError` check, or verify that `GitHubAuthError extends EnrichmentAdapterError`. The two branches should not be collapsed with an unsafe cast.

---

### Non-blocking — UX: Auth mismatch error does not list supported kinds

**File**: `packages/engram-cli/src/commands/ingest.ts`, github action handler (lines ~490-497)

When `buildAuthCredential()` throws (auth mismatch), the error is caught and forwarded as-is:

```ts
} catch (err) {
  log.error(err instanceof Error ? err.message : String(err));
  process.exit(1);
}
```

`buildAuthCredential` does include the supported kinds in its throw message (`adapter 'github' supports auth: bearer, none`), so the kinds _are_ visible — but only because of implementation knowledge of the throw message format. This is fine but fragile: if the throw message is ever changed, the UX breaks. Consider making the error surface explicit by catching and re-throwing with a structured message, or documenting that `buildAuthCredential`'s throw message is part of the contract. Non-blocking because it works correctly today.

---

### Non-blocking — UX: Scope validation failure does not include `scopeSchema.description`

**File**: `packages/engram-cli/src/commands/ingest.ts`

The scope validation errors correctly log the error + description:

```ts
log.error(
  `Invalid scope for github: ${scopeErr}\n${adapter.scopeSchema.description}`,
);
```

This is good. However the _missing_ scope path does the same:

```ts
log.error(
  `--scope is required for github adapter.\n${adapter.scopeSchema.description}`,
);
```

Both include the description. No issue here — this is actually correct behaviour.

---

### Non-blocking — Test quality: scope-gate test does not exercise the actual guard

**File**: `packages/engram-cli/test/ingest/ingest-v2.test.ts`, `"scope validation gates enrich()"` describe block

The test patches `adapter.enrich` to track calls, but then the validation check is:

```ts
if (scopeErr) {
  // Simulates CLI bail-out on validation error — enrich never called
} else {
  expect(false).toBe(true);
}
expect(enrichCalled).toBe(false);
```

This test only verifies that `enrichCalled` stays false _because the test itself never calls enrich_ — not because the CLI guard prevented it. The test will pass even if the guard is removed. This is essentially a tautology. To be meaningful, the test should call the action handler (or extract the validation logic) and verify enrich is not invoked as a side effect of the handler. Non-blocking since the contract is tested at the unit level in `auth.test.ts`, but the test gives false confidence.

---

### Non-blocking — CLAUDE.md line limits

`gerrit.ts` (565 lines) and `github.ts` (643 lines) both exceed the 500-line ceiling stated in CLAUDE.md. These were pre-existing violations not introduced by this PR, so no action required here, but worth noting if the file grows further.

---

### Summary

| # | Severity | Area | Finding |
|---|----------|------|---------|
| 1 | **Blocking** | Security | `service_account` path not sanitized before `readFileSync` |
| 2 | **Blocking** | Correctness | `GitHubAuthError` unsafely cast to `EnrichmentAdapterError` in `handleEnrichError` |
| 3 | Non-blocking | UX | Auth mismatch error surfacing relies on throw-message format contract |
| 4 | Non-blocking | Tests | Scope-gate test is a tautology — does not exercise the actual CLI guard |
| 5 | Non-blocking | CLAUDE.md | `gerrit.ts` and `github.ts` exceed 500-line limit (pre-existing) |

**Merge readiness**: Not yet — please address items 1 and 2 before merging. Items 3–5 can be tracked as follow-ups.

rnwolfe (commented):
Fixed: service_account path now uses path.resolve() + .json extension check; GitHubAuthError instanceof handling separated from EnrichmentAdapterError branch to prevent silent code:undefined fall-through.

Issue #197, opened 2026-04-18
feat(ingest): cross-source reference edge resolver — link episodes across adapters via pattern registry


rnwolfe:
## Goal

Build a registry-based cross-source reference resolver that runs after any ingestion
batch and creates `references` edges between **entities** from different sources when
one episode's content mentions another entity's identifier. The resolver is the
mechanism by which a commit body citing a Linear issue, a Gerrit CL description citing
`b/12345678`, or a Google Doc referencing a GitHub PR becomes a traversable edge in the
graph.

## Problem

Each adapter today detects only self-contained references (the GitHub adapter matches
SHAs and `#N` inline at `ingest/adapters/github.ts`). Cross-source references — the
high-signal edges — are not captured anywhere. A commit that says "implements b/12345678
per design doc go/foo-design" is a three-way connection between git, Buganizer, and
Google Docs that the current pipeline throws away.

With Gerrit (#191), Buganizer (#192), GitLab (#193), Jira (#194), Linear (#195), Google
Docs (#196), and Google Drive (#198) adapters landing, the same pattern needs to apply
uniformly — not be reimplemented per adapter.

### Reference pattern registry

```ts
// packages/engram-core/src/ingest/cross-ref/patterns.ts

export interface ReferencePattern {
  /** episodes.source_type of the target (from EPISODE_SOURCE_TYPES registry, #199) */
  sourceType: string;
  /** Regex; capture group 1 = the identifier */
  pattern: RegExp;
  /** Map the match to the graph's source_ref value */
  normalizeRef: (match: string) => string;
  /** Confidence on the emitted edge, 0..1. See "Per-pattern confidence" below. */
  confidence: number;
}
```

Built-in patterns (all imported from the vocab registry, no inline strings):

| Pattern | Target source_type | Confidence | Example match |
|---|---|---|---|
| Full GitHub issue/PR URL | `github_issue` / `github_pr` | `0.95` | `github.com/org/repo/pull/42` |
| Full Gerrit CL URL | `gerrit_change` | `0.95` | `go-review.googlesource.com/c/.../+/12345` |
| Full Google Doc URL | `google_doc` | `0.95` | `docs.google.com/document/d/<id>` |
| Full Linear issue URL | `linear_issue` | `0.95` | `linear.app/team/issue/ENG-123` |
| Full Jira issue URL | `jira_issue` | `0.95` | `foo.atlassian.net/browse/PROJ-42` |
[...]

Issue #207, opened 2026-04-19
feat(core): episode supersession — schema, graph helpers, verifyGraph invariants


rnwolfe:
## Goal

Implement episode supersession per the spec in `docs/internal/specs/mutable-sources.md`
(authored by #201): schema change, graph-layer helper, verifyGraph invariant, and
retrieval helper for "current revision" queries.

### Schema change

- `packages/engram-core/src/format/schema.ts` — add nullable column to `episodes`:

  ```sql
  superseded_by TEXT REFERENCES episodes(id)
  ```

- Index for the "current version" query:

  ```sql
  CREATE INDEX idx_episodes_current ON episodes(source_type, source_ref) WHERE superseded_by IS NULL;
  ```

- Additive and nullable — no format-version bump. Existing `.engram` files gain the
  column via `ALTER TABLE ADD COLUMN` on open if the column is absent.

### Graph layer — `supersedeEpisode`

- `packages/engram-core/src/graph/episodes.ts` (new export):

  ```ts
  export function supersedeEpisode(
    graph: EngramGraph,
    priorEpisodeId: string,
    newEpisodeInput: EpisodeInput,
  ): Episode;
  ```

- Semantics: in one transaction, insert the new episode, then update the prior
  episode's `superseded_by` to the new episode's id. Atomic — either both rows land
  or neither.
- Errors: throws if `priorEpisodeId` doesn't exist, is already superseded, or
  `newEpisodeInput` collides with an existing non-superseded episode for the same
  `(source_type, source_ref)` tuple.

### verifyGraph invariant

- New check in `packages/engram-core/src/format/verify.ts`: **each episode has at
  most one successor**. Fan-in uniqueness on `superseded_by`. Violations surface as
  `error` severity (not warning — a fan-in collision corrupts the chain).
- Additional check: `superseded_by` points to a valid existing episode id. Dangling
  references surface as `error`.

### Retrieval helper

- `packages/engram-core/src/graph/episodes.ts`:

  ```ts
  export function getCurrentEpisode(
    graph: EngramGraph,
    sourceType: string,
    sourceRef: string,
  ): Episode | null;
  ```

- Index-backed query: `WHERE source_type = ? AND source_ref = ? AND superseded_by IS NULL`.

### Dedup semantics update

- `packages/engram-core/src/graph/episodes.ts:addEpisode` — existing dedup via the
  unique index now keys on the `idx_episodes_current` partial index where applicable.
[...]

PR #216, opened 2026-04-19
docs: revision-aware episodes spec and ADR-007 (#201)

### LINKED ISSUE #201: spec(core): revision-aware episodes for mutable sources (Google Docs, Linear, Jira)
## Goal

Land the decision on episode model for mutable sources, record it as ADR-007, and
author the reference spec. Implementation tracked separately in a follow-up issue.

## Problem

Today, episode dedup is enforced by the unique index on `(source_type, source_ref)`
(`format/schema.ts:97`). The GitHub adapter pre-checks for an existing episode and
skips on collision (`ingest/adapters/github.ts:324-334`). Correct for commits and PRs
— they are immutable after merge.

Wrong for Google Docs. A doc at `docs.google.com/document/d/abc123` evolves. If we
ingest revision N and later re-run, we need to capture N+1, not silently skip. Same
class of problem for Linear/Jira issues: description, status, assignee all mutate.

## Decision — episode supersession (Option B)

Add a nullable `superseded_by TEXT` column to `episodes` that references `episodes.id`.
Re-ingesting a mutated source creates a new episode and updates the prior episode's
`superseded_by` to point at the new one. In one transaction.

### Why

1. **Parallels the edge model.** Edges already use `invalidated_at` + supersession.
   Episode supersession is a direct parallel; the mental model is already in place.
2. **Preserves the "immutable content" invariant.** `superseded_by` is a *linkage*
   field, not content. Episode content stays immutable once written; only the chain
   pointer is updated. Analogous to how edges have immutable facts but mutable
   `invalidated_at`.
3. **Trivial "current version" query.** `WHERE superseded_by IS NULL` is an
   index-friendly lookup.
4. **Small schema change.** One nullable column, no data rewrite, no format-version
   bump required (additive, backward-compatible).
5. **New verifyGraph invariant is cheap.** "Each episode has at most one successor"
   is a 10-line check — fan-in uniqueness on `superseded_by`.

### Rejected alternatives

- **Option A (revision-suffixed `source_ref`).** Zero-schema-change but pushes
  "current version" query complexity into retrieval and risks source_ref format drift
  across adapters. Fine for a simpler system; loses elegance when supersession is
  already a modeled concept elsewhere.
- **Option C (hybrid: new rows + `is_current` flag).** Combines A's adapter complexity
  with B's schema churn. Strictly worse.

### Revision identifier lives in `metadata`

Each adapter records its source-native revision identifier in the existing
`episodes.metadata` JSON column:

- Google Docs: `revisionId` (from Drive API)
- Linear: `updatedAt` (ISO timestamp)
- Jira: `updated` (ISO timestamp)

No top-level schema field for revision — `metadata` is the right home for
adapter-specific fields.

### Cross-ref resolver behavior (#197)

An *entity* of `entity_type: 'document'` (or `linear_issue`, `jira_issue`) represents
the doc/issue identity. Episodes are revisions that attach as evidence. References
from other sources resolve to the **entity**, not to a specific episode revision.
This mirrors how commits work: the commit SHA is the entity; the commit message is
one episode of evidence.

Edges carry `valid_from` timestamps; if a reference needs to capture
"which revision was current when I referenced this doc," the edge's `valid_from`
plus the target entity's episode chain are sufficient to answer that query — no new
mechanism required.

## Out of scope (this issue)

- **Implementation of the schema change, graph-layer helper, verifyGraph invariant,
  and retrieval helpers.** Tracked in a separate follow-up issue filed alongside this
  spec.
- **Per-adapter migration** for Google Docs / Linear / Jira. Each adapter issue
  (#196, #195, #194) adopts supersession natively.
- **Historical revision backfill** for docs not yet ingested. Adapters capture from
  first-ingest forward; retroactive revision history is a future extension.

## Acceptance criteria

- [ ] `docs/internal/specs/mutable-sources.md` authored. Covers: the problem, the
      decision (Option B), schema change shape, supersession algorithm (atomic
      transaction semantics), verifyGraph invariant, cross-ref behavior (doc-as-entity,
      revisions-as-episodes), per-source revision identifier conventions,
      interaction with #197.
- [ ] `docs/internal/DECISIONS.md` — **ADR-007** added. Records: decision (Option B),
      rejected alternatives (A, C) with one-line rationale each, consequences
      (adapter authoring guidance, new verifyGraph invariant, migration-free schema
      add).
- [ ] Spec includes a worked example: pseudocode for a Google Docs adapter ingesting
      doc `abc123` at revision N, then re-ingesting at revision N+1, showing the
      supersession update in transaction.
- [ ] #197 cross-reference: cross-ref resolver spec notes that references resolve to
      document entities, not episode revisions.

## Deliverables

- `docs/internal/specs/mutable-sources.md` (new)
- `docs/internal/DECISIONS.md` — ADR-007 entry (amended)

## Blocks

- **#207** — implementation issue (schema change, graph helpers, verifyGraph
  invariants)
- **#194** (Jira), **#195** (Linear), **#196** (Google Docs) — each adopts supersession
  natively and cannot ship until both this spec and the implementation land

## Related

- **#197** — cross-source reference edge resolver; behavior across revisions resolved
  here
- **#199** — vocabulary registries; `document`, `linear_issue`, `jira_issue`
  entity_types must be declared
- **#200** — v2 contract unaffected (supersession lives below the adapter boundary)

rnwolfe:
## Summary

- Adds `docs/internal/specs/mutable-sources.md` — comprehensive spec for episode supersession, the mechanism by which mutable-source adapters (Google Docs, Linear, Jira) capture new revisions without losing history
- Adds ADR-007 to `docs/internal/DECISIONS.md` recording the decision (Option B), rejected alternatives (A and C), and consequences

## What the spec covers

- **The problem**: the current `(source_type, source_ref)` unique index silently skips re-ingestion of mutated sources
- **Decision (Option B)**: nullable `superseded_by TEXT REFERENCES episodes(id)` column + atomic supersession transaction
- **Rejected alternatives**: Option A (revision-suffixed source_ref) and Option C (hybrid is_current flag) with rationale
- **Schema DDL**: `ALTER TABLE episodes ADD COLUMN superseded_by` + replacement partial unique index + current-version partial index
- **`supersedeEpisode()` API**: atomic two-step transaction semantics and error conditions
- **Two new `verifyGraph()` invariants**: dangling `superseded_by` refs (error) and fan-in uniqueness violation (error)
- **Revision identifier convention**: stored in `metadata` JSON (`revisionId`, `updatedAt`, `updated` per source)
- **Adapter authoring decision tree**: when to call `addEpisode` vs `supersedeEpisode` vs skip
- **Cross-source reference behavior** (#197): references resolve to the entity (document identity), not to a specific revision episode
- **Worked Google Docs example**: first ingest, re-ingest with new revision, idempotent re-ingest with unchanged content

## Acceptance Criteria

- [x] Spec documents the problem with mutable sources under the current schema
- [x] Option B (episode supersession) is the chosen decision with full rationale
- [x] Rejected alternatives (A and C) are documented with one-line rationale each
- [x] Schema change shape is specified (DDL, index replacement)
- [x] Supersession algorithm with atomic transaction semantics is documented
- [x] Error conditions for `supersedeEpisode()` are specified
- [x] `verifyGraph()` invariants are specified (dangling refs + fan-in uniqueness, both error severity)
- [x] Revision identifier convention in `metadata` JSON is documented per source
- [x] Cross-ref resolver behavior (#197 interaction) is documented
- [x] Worked example (Google Docs) covers first ingest, re-ingest, and idempotent case
- [x] ADR-007 in DECISIONS.md matches the format of existing ADRs

Closes #201

### DIFF
--- a/docs/internal/DECISIONS.md
+++ b/docs/internal/DECISIONS.md
@@ -169,3 +169,62 @@ shape depends on `AuthCredential` stabilization (#200).
 - Vocab extensions are additive and runtime-only — they do not generate TypeScript types.
   Downstream callers that need compile-time safety must import built-in types from
   `packages/engram-core/src/vocab/` and handle plugin-extended values as opaque strings.
+
+## ADR-007 -- Episode supersession for mutable sources
+
+**Date**: 2026-04-19
+**Status**: Accepted
+**Spec**: [`docs/internal/specs/mutable-sources.md`](specs/mutable-sources.md)
+**Issue**: #201
+
+**Context**: Engram's unique index on `(source_type, source_ref)` enforces idempotency
+for immutable sources (git commits, PRs, Gerrit changes). For mutable sources — Google
+Docs, Linear issues, Jira tickets — the same `source_ref` legitimately produces different
+content at different times. Under the current schema, re-ingesting a mutated source
+silently skips the new revision, leaving the knowledge graph anchored to the first
+ingestion and unable to reflect the document's current state.
+
+Three options were evaluated:
+
+- **Option A** — revision-suffixed `source_ref` (e.g. `doc:abc123@rev_N`): no schema
+  change, but pushes complexity into every mutable-source adapter, makes "current version"
+  queries non-index-friendly, and risks `source_ref` format drift across adapter versions.
+- **Option B** — episode supersession: add nullable `superseded_by TEXT REFERENCES episodes(id)`,
+  create new episode on re-ingest, atomically link the prior episode to the new one. Current
+  version query: `WHERE superseded_by IS NULL`.
+- **Option C** — hybrid new rows + `is_current` boolean flag: combines Option A's adapter
+  complexity with Option B's schema change, with the added risk of inconsistency between
+  the flag and the derived truth (`superseded_by IS NULL`). Strictly worse than B.
+
+**Decision**: Adopt **Option B — episode supersession**. Add a nullable `superseded_by`
+column to the `episodes` table and expose a new `supersedeEpisode()` API. The supersession
+algorithm is atomic (single transaction: INSERT new row, UPDATE prior row's `superseded_by`)
+and follows the same pattern already used for edges and projections.
+
+**Alternatives considered**:
+- *Option A (revision-suffixed source_ref).* Rejected: no schema change is the only
+  advantage; every other dimension is worse — adapter complexity, retrieval ergonomics,
+  format stability.
+- *Option C (hybrid new rows + is_current flag).* Rejected: combines the costs of both A
+  and B with no compensating benefit; stores a derived truth redundantly.
+
+**Consequences**:
+
+*Positive*
+- The episode table gains the same temporal linkage language as edges and projections —
+  one pattern for all supersession across the schema.
+- The immutable-content invariant is preserved: `superseded_by` is linkage metadata, not
+  a mutation of episode content. Historical revisions are retained in full.
+- "Current version" lookup is a single index-friendly predicate (`WHERE superseded_by IS NULL`).
+- Additive, nullable column — no migration of existing rows, no format-version bump.
+  Applied via `ADDITIVE_DDL` in `openGraph()`.
+
+*Negative / cost*
+- Adapter authoring convention changes: mutable-source adapters must call
+  `supersedeEpisode()` on re-ingest instead of `addEpisode()`. Built-in adapters for
+  immutable sources are unaffected; existing adapters require no changes.
+- Two new `verifyGraph()` invariants are added (dangling `superseded_by` refs → error;
+  fan-in uniqueness on `superseded_by` → error). Both are cheap single-scan queries.
+- The unique index on `(source_type, source_ref)` must be replaced with a partial index
+  scoped to `WHERE superseded_by IS NULL`, allowing historical revisions to coexist.
+  This is a one-time DDL migration applied alongside the column addition.
--- a/docs/internal/specs/mutable-sources.md
+++ b/docs/internal/specs/mutable-sources.md
@@ -1,0 +1,559 @@
+# Revision-Aware Episodes for Mutable Sources
+
+> Spec for issue #201 — episode supersession for mutable sources (Google Docs, Linear, Jira).
+> Implementation tracked in the same issue.
+
+## Overview
+
+Engram's episode model was designed around immutable sources: a git commit is a
+permanent, unchangeable artifact. Its `(source_type, source_ref)` pair identifies it
+uniquely and forever. The current unique index on that pair enforces this invariant and
+makes ingestion idempotent — re-ingesting the same commit is a no-op.
+
+Mutable sources break this assumption. A Google Doc has one canonical URL (`source_ref`)
+but its content changes every time someone edits it. A Linear issue has one issue ID but
+its description, status, and assignee evolve over time. A Jira ticket accumulates
+comments and state changes that are not reflected in the original episode.
+
+Under the current schema, re-ingesting a mutable source silently skips the new revision
+because the `(source_type, source_ref)` pair already exists. The knowledge graph stagnates:
+it reflects only the first ingestion, not the current state of the source.
+
+This spec defines **episode supersession** — the mechanism by which adapters for mutable
+sources capture new revisions without losing the history of prior revisions.
+
+Cross-references: edges already use `superseded_by` and `invalidated_at` for temporal
+transitions (see `docs/internal/DECISIONS.md` ADR-001 and the temporal model in
+`CLAUDE.md`). This spec applies a parallel pattern to episodes.
+
+---
+
+## 1. The Problem in Detail
+
+### 1.1 Current schema
+
+```sql
+CREATE TABLE episodes (
+  id           TEXT NOT NULL UNIQUE,
+  source_type  TEXT NOT NULL,
+  source_ref   TEXT,
+  content      TEXT NOT NULL,
+  -- ... other fields
+);
+CREATE UNIQUE INDEX idx_episodes_identity
+  ON episodes(source_type, source_ref)
+  WHERE source_ref IS NOT NULL;
+```
+
+The unique index on `(source_type, source_ref)` is the right design for immutable sources.
+For mutable sources it produces a silent skip on re-ingestion — the adapter calls
+`addEpisode(...)`, the INSERT hits the unique constraint, and the new revision is discarded.
+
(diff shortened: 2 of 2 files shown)


rnwolfe (commented):
## Review: docs/internal/specs/mutable-sources.md + ADR-007

Overall the spec is thorough and covers the vast majority of acceptance criteria from #201. The problem statement, decision rationale, worked example, verifyGraph invariants, revision identifier convention, and cross-ref resolver behavior are all present and well-written. Two issues need resolution before this can safely unblock #207.

---

### Blocking: `supersedeEpisode` return type mismatch (§7 vs. issue #207)

The spec declares:

```typescript
function supersedeEpisode(
  graph: EngramGraph,
  priorId: string,
  newEpisode: EpisodeInput,
): string;   // ← returns the new episode ID
```

Issue #207 (the implementation issue this spec must unblock) specifies the signature as:

```typescript
export function supersedeEpisode(
  graph: EngramGraph,
  priorEpisodeId: string,
  newEpisodeInput: EpisodeInput,
): Episode;  // ← returns the full Episode object
```

And its acceptance criteria read: _"supersedeEpisode() implemented with atomic transaction semantics; **returns the new episode**"_ — consistent with `Episode`, not a bare string.

The implementer will follow #207's acceptance criteria (they are the contract), but the signature in the spec conflicts. This will produce a spec reference that is wrong on its face. Update §7 to return `Episode` so spec and implementation issue are aligned. The parameter name `priorId` vs. `priorEpisodeId` is a minor inconsistency worth fixing at the same time.

---

### Blocking: `idx_episodes_identity` replacement — migration mechanism unspecified

Section 4.2 correctly identifies that the existing unique index:

```sql
CREATE UNIQUE INDEX idx_episodes_identity
  ON episodes(source_type, source_ref)
  WHERE source_ref IS NOT NULL;
```

must be replaced with:

```sql
CREATE UNIQUE INDEX idx_episodes_identity
  ON episodes(source_type, source_ref)
  WHERE source_ref IS NOT NULL AND superseded_by IS NULL;
```

But the spec says to add this to `ADDITIVE_DDL` alongside the column, which uses `IF NOT EXISTS`. A `DROP INDEX` + `CREATE INDEX` cannot be expressed idempotently via `IF NOT EXISTS` — the old index will still exist on all pre-existing databases, meaning the uniqueness constraint will fire incorrectly when a superseded episode and its successor share the same `(source_type, source_ref)`.

The current `ADDITIVE_DDL` in `schema.ts` contains only `CREATE TABLE IF NOT EXISTS` and `CREATE INDEX IF NOT EXISTS` statements. Replacing an existing index requires either:

- A versioned migration step (e.g. a new entry in a migrations table keyed by a migration ID), or
- A guard query: check whether the index already has the correct predicate before dropping and recreating it.

The spec needs to specify the mechanism. Without it, the implementer has to invent the migration strategy, which is exactly what the spec is supposed to prevent. This is the highest-risk gap because getting it wrong leaves existing `.engram` files broken for mutable-source re-ingestion even after the feature ships.

---

### Non-blocking observations (no action required if authors disagree)

**§11.2 — `getCurrentEpisode` behavior on redacted episodes**: The spec states the helper returns the current episode _"regardless of redaction status"_ and that adapters must check `status` themselves. This is a defensible choice, but issue #207 does not mention this nuance in the acceptance criteria or test cases. Worth adding a note to §8.1's decision tree (or a test case in #207) so the adapter author doesn't need to read §11.2 to learn about this.

**§2.2 error condition — collision is "a logic error"**: The spec says a collision with an existing non-superseded episode for the same `(source_type, source_ref)` "indicates a logic error in the adapter." Labelling it a logic error is accurate, but the spec could be stronger: state that `supersedeEpisode()` throws a typed error (e.g. `EpisodeCollisionError`) consistent with the project's error-handling convention (typed error classes per CLAUDE.md §Error Handling). Issue #207 already implies this; the spec should name it.

**§4.1 says "added to `ADDITIVE_DDL`" but §4.2 contradicts this**: Beyond the migration mechanism gap noted above, the two subsections give slightly different framing. §4.1 says both the column and the new partial index go into `ADDITIVE_DDL`. §4.2 then introduces the index _replacement_ as a separate step without reconciling it with the additive approach. A single "Migration Steps" subsection that enumerates all DDL changes in execution order would eliminate the inconsistency.

---

### Acceptance criteria coverage check (issue #201)

| Criterion | Present? |
|---|---|
| Problem statement | Yes — §1 |
| Decision (Option B) with full rationale | Yes — §2, §3 |
| Schema change shape (SQL) | Yes — §2.1, §4.1 |
| Supersession algorithm with atomic transaction semantics | Yes — §2.2 |
| Error conditions (non-existent prior, already-superseded, collision) | Yes — §2.2 |
| verifyGraph invariants at error severity | Yes — §5 |
| Revision identifier convention (metadata JSON, per-source) | Yes — §6 |
| Cross-ref resolver behavior (resolves to entity, not revision) | Yes — §9 |
| Worked Google Docs pseudocode (first ingest AND re-ingest) | Yes — §10 (also covers skip case) |
| ADR-007 in DECISIONS.md | Yes — decision, rejected alternatives (A, C) with rationale, consequences |

All criteria are substantively covered. The two blocking issues above are about spec-to-implementation alignment and migration completeness, not missing sections.

rnwolfe (commented):
**Review summary (blocking — cannot approve):** Two issues need fixes before this unblocks #207. Detail is in the review comment above. Short version: (1) §7  return type is  but #207 specifies  — direct conflict between spec and implementation issue; (2) §4.2 requires  +  with revised predicate, but the spec says to use  which only supports  — the migration mechanism is unspecified and will leave existing databases with a broken uniqueness constraint.

rnwolfe (commented):
Review summary (blocking): Two issues need fixes before this unblocks #207. Detail is in the review comment above. Short version: (1) Sec 7 supersedeEpisode return type is string but issue 207 specifies Episode — direct conflict between spec and implementation issue; (2) Sec 4.2 requires DROP INDEX idx_episodes_identity and CREATE UNIQUE INDEX with revised predicate, but the spec says to use ADDITIVE_DDL which only supports IF NOT EXISTS — the migration mechanism is unspecified and will leave existing databases with a broken uniqueness constraint after the feature ships.

rnwolfe (commented):
Fixed: supersedeEpisode signature updated to return Episode with priorEpisodeId parameter name; §4.2 migration now specifies DROP INDEX IF EXISTS + CREATE UNIQUE INDEX with superseded_by IS NULL predicate, making both steps idempotent at database open time.

Issue #224, opened 2026-04-20
docs: update VISION.md phase 2 for in-repo plugin adapter model


rnwolfe:
## Summary

Rewrite the phase 2 adapter roadmap in `docs/internal/VISION.md` to reflect
ADR-008: first-party adapters (Jira, Linear, GitLab, Google Docs) ship as
in-repo plugins under `packages/plugins/`, not as adapters compiled into
`engram-core`.

## Strategic Rationale

VISION.md phase 2 currently lists "Enrichment adapters: Gerrit, Jira, Linear,
GitLab" as planned built-ins. That list predates the plugin system and
ADR-008. Left unchanged, it continues to imply built-in delivery and will
drift future work in the wrong direction.

This is a documentation-only update — no code changes, no spec changes. The
ADR is the decision; VISION.md needs to match.

## Scope

**Edit `docs/internal/VISION.md` to:**

1. Replace the phase 2 "Enrichment adapters" bullet with language that reflects
   plugin delivery. Suggested wording:

   > **First-party plugin adapters** — Jira, Linear, GitLab, and Google Docs
   > ship as in-repo plugins (`packages/plugins/<name>/`) loaded via the plugin
   > system. Gerrit migrates from built-in to the same model as the reference
   > port. See [ADR-008](internal/DECISIONS.md#adr-008).

2. Update the principles section if any principle references "built-in
   adapters" as an identity claim — check principle 1 (embeddable not
   monolithic) and ensure the plugin model is correctly framed as an extension
   of that principle, not in tension with it.

3. If there is a commands/features map with adapter rows, reclassify them.

## Out of Scope

- STATUS.md sync (separate `/product sync` task — though this issue may
  prompt it).
- Changes to ADR-008 itself.
- Code changes.

## Acceptance Criteria

- [ ] `docs/internal/VISION.md` phase 2 no longer lists the four adapters as
      built-ins.
- [ ] A sentence or subsection in VISION.md explicitly references ADR-008 for
      the delivery model.
- [ ] No orthogonal VISION.md edits — this PR is scoped to the adapter
      delivery question only.

## Dependencies

- ADR-008 merged (docs/internal/DECISIONS.md)

Issue #223, opened 2026-04-20
engram plugin install — wire first-party plugins into XDG


rnwolfe:
## Summary

Add `engram plugin install <name>` and `engram plugin uninstall <name>`
subcommands to wire first-party in-repo plugins (`packages/plugins/<name>/`)
into the XDG plugin directory, activating them for the user without a manual
symlink or copy step.

## Strategic Rationale

ADR-008 ships first-party adapters (Gerrit, Jira, Linear, GitLab, Google Docs)
as in-repo plugins under `packages/plugins/`. Without an install command, users
must manually symlink or copy those directories into
`$XDG_DATA_HOME/engram/plugins/` — a sharp UX regression from the current
built-in model where `engram ingest enrich gerrit` just works after `npm install`.

The install command closes that gap: `engram plugin install gerrit` links the
bundled plugin into XDG and it becomes available system-wide.

## Scope

**In scope:**
- `engram plugin install <name>` — locate the plugin in the installed engram
  distribution (bundled `packages/plugins/<name>/`), create a symlink (or copy
  on platforms without symlinks) in `$XDG_DATA_HOME/engram/plugins/<name>/`,
  validate the manifest, report success.
- `engram plugin uninstall <name>` — remove the XDG entry for a previously
  installed plugin. Refuses to remove user-authored plugins (directories that
  aren't symlinks pointing into the engram install, or that fail a checksum
  match) without `--force`.
- `engram plugin list` — extend existing output to show source (`bundled`,
  `symlinked`, `user`) alongside scope.
- `--project` flag to install into `<project>/.engram/plugins/` instead of
  user-wide XDG.
- Interactive prompt from `engram init` offering to install common plugins
  (deferred to a separate follow-up if scope grows).

**Out of scope:**
- Installing third-party plugins from arbitrary URLs or registries (ADR-006
  rejected network fetch at runtime; this stays true).
- Signing or checksum verification beyond the "symlink points into our install"
  check for uninstall safety.

## Acceptance Criteria

- [ ] `engram plugin install gerrit` (after the Gerrit migration lands) creates
      a valid XDG entry and `engram ingest enrich gerrit --scope <project>`
      works without further setup.
- [ ] `engram plugin install nonexistent` exits non-zero with a clear error
      ("no bundled plugin named 'nonexistent'; see `engram plugin list --available`")
- [ ] `engram plugin uninstall gerrit` reverses the install.
[...]

Issue #222, opened 2026-04-20
Migrate Gerrit adapter from built-in to in-repo plugin


rnwolfe:
## Summary

Port the Gerrit enrichment adapter from its current built-in location
(`packages/engram-core/src/ingest/adapters/gerrit.ts`) to an in-repo plugin at
`packages/plugins/gerrit/`, loaded via the plugin system (ADR-006) rather than
compiled into `engram-core`.

This is the reference implementation for ADR-008 — executable proof that the
plugin contract is adequate for a full-featured first-party adapter.

## Strategic Rationale

Gerrit shipped built-in in v0.2.0 because the plugin system did not yet exist.
Now that the loader is stable and ADR-008 reclassifies all first-party
adapters as in-repo plugins, Gerrit is the migration example. Every future
first-party adapter (Jira, Linear, GitLab, Google Docs) will use this port as
its template.

## Scope

**In scope:**
- Create `packages/plugins/gerrit/` with:
  - `manifest.json` (`transport: js-module`, `contract_version`, declared
    capabilities, `supportedAuth`, `scopeSchema` from current adapter)
  - `package.json` with Gerrit-specific dependencies extracted from `engram-core`
  - `src/index.ts` exporting a default `EnrichmentAdapter` implementation
  - `src/helpers.ts` (ported from `gerrit-helpers.ts`)
  - Test files colocated, using the same fixtures as the built-in version
- Remove `packages/engram-core/src/ingest/adapters/gerrit.ts` and
  `gerrit-helpers.ts`
- Remove the hardcoded `gerrit` subcommand registration in
  `packages/engram-cli/src/commands/ingest.ts` (plugin loader auto-registers)
- Update Bun workspace config if needed
- Update `CLAUDE.md` file organization section

**Out of scope:**
- Changes to Gerrit adapter behavior, auth, or scope handling
- `engram plugin install` subcommand (tracked separately)
- GitHub adapter migration (stays built-in per ADR-008)

## Acceptance Criteria

- [ ] `packages/plugins/gerrit/` exists with manifest, package.json, and tests
- [ ] `engram-core` no longer imports Gerrit-specific code or dependencies
- [ ] `bun run build` succeeds from a clean state
- [ ] `bun test` passes, including Gerrit adapter tests now running from the plugin package
- [ ] After `engram plugin install gerrit` (or manual symlink equivalent until that lands),
      `engram plugin list` shows gerrit as discovered and `engram ingest enrich gerrit --scope <project>`
[...]

PR #225, opened 2026-04-20
docs: ADR-008 first-party adapters ship as in-repo plugins

rnwolfe:
## Summary

Adds ADR-008 to `docs/internal/DECISIONS.md` deciding how first-party
enrichment adapters (Jira, Linear, GitLab, Google Docs, future) are delivered
now that the plugin system (ADR-006) has shipped.

**Decision**: first-party adapters ship as in-repo plugins under
`packages/plugins/<name>/` — maintained by the core team, but loaded through
the plugin system rather than being compiled into `engram-core`. Gerrit
migrates to the same model as the reference port before v0.3.0. GitHub stays
built-in as a documented exception because it is foundational to the VCS
layer, not a selectable enrichment.

## Why now

`VISION.md` phase 2 lists four adapters (Gerrit, Jira, Linear, GitLab) as
planned built-ins. That list predates the plugin system. Gerrit already
shipped built-in in v0.2.0, which was correct at the time but created inertia
toward the built-in default. Without an explicit decision, the next three
adapters would drift into `engram-core` by momentum, the plugin contract
would never be exercised by first-party code, and contract bugs would surface
first on community adapters with less test coverage.

## Alternatives considered

- **All built-in** — rejected: unbounded dependency growth in `engram-core`,
  no first-party validation of the plugin contract.
- **All out-of-repo** — rejected: loses the "we maintain this" signal,
  fragments issue tracking, makes cross-adapter refactors into multi-repo
  coordination.
- **Mixed case-by-case, no ADR** — rejected: this is the status quo and it
  produced the problem.

## Follow-ups

- #222 — Migrate Gerrit adapter from built-in to in-repo plugin (reference port)
- #223 — `engram plugin install` subcommand (wire first-party plugins into XDG)
- #224 — Update `VISION.md` phase 2 to match the in-repo-plugin model

## Test plan

- [x] ADR renders cleanly in the DECISIONS.md file (verified locally)
- [ ] Reviewer confirms the decision boundary (first-party vs third-party) is
      clearly stated
- [ ] Reviewer confirms Gerrit migration sequencing (before v0.3.0) is realistic

This is a docs-only PR. No code changes, no tests to run. The three follow-up
issues above track the execution work.

### DIFF
--- a/docs/internal/DECISIONS.md
+++ b/docs/internal/DECISIONS.md
@@ -228,3 +228,131 @@ and follows the same pattern already used for edges and projections.
 - The unique index on `(source_type, source_ref)` must be replaced with a partial index
   scoped to `WHERE superseded_by IS NULL`, allowing historical revisions to coexist.
   This is a one-time DDL migration applied alongside the column addition.
+
+## ADR-008 -- First-party adapters ship as in-repo plugins
+
+**Date**: 2026-04-19
+**Status**: Accepted
+**Supersedes (partial)**: The adapter roadmap in [`VISION.md`](VISION.md) phase 2,
+which implicitly treated Gerrit, Jira, Linear, and GitLab as built-in adapters.
+
+**Context**: `VISION.md` phase 2 lists four enrichment adapters as planned work:
+Gerrit, Jira, Linear, GitLab. That list was written before the plugin system
+(ADR-006) existed. With the plugin loader shipped in v0.2.0, the "how should the
+next three adapters be delivered" question becomes load-bearing: built-in means
+they compile into `engram-core`, bloat the release binary, and couple adapter
+release cadence to the engine. Out-of-repo means users lose the "core team
+maintains this" signal on integrations that are clearly first-party.
+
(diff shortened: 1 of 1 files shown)


rnwolfe:
Addressed Copilot review in ba46deb → latest force-push:

1. **Versioning vs release cadence (Decision §4)** — clarified: plugins carry independent semver but ship on a synchronized release train with the engine. Independent versioning covers SDK-patch releases without an engine release; synchronized trains prevent contract_version drift at release boundaries.
2. **Workspace glob update** — added an explicit Consequence entry noting the root `package.json` `workspaces` glob must be extended to `packages/plugins/*` alongside the first plugin package.
3. **Pre-authorized ambiguity** — rephrased: in-repo plugins require the explicit `engram plugin install <name>` invocation as the user-consent step, no implicit enablement.
4. **Duplicated 'the'** — removed in Context section.
5. **Non-monotonic ADR order** — moved ADR-008 to after ADR-007 so the file reads …ADR-006, ADR-007, ADR-008.

PR #228, opened 2026-04-20
feat(cli): engram plugin install and uninstall subcommands

### LINKED ISSUE #223: engram plugin install — wire first-party plugins into XDG

rnwolfe:
## Summary

- Adds `engram plugin install <name>` subcommand that symlinks (or copies on Windows) a bundled first-party plugin from the engram install root into the XDG user-wide plugin directory (`$XDG_DATA_HOME/engram/plugins/`) or project-local (`.engram/plugins/`) with `--project`
- Adds `engram plugin uninstall <name>` that removes symlinked/bundled installs freely, but requires `--force` for user-authored (plain-directory) plugins
- Extends `engram plugin list` with a **SOURCE** column (`bundled` / `symlinked` / `user`) and a `--available` flag to list installable bundled plugins
- Extends `engram-core/plugins/discover.ts` with exported helpers: `xdgDataHome`, `userPluginsRoot`, `projectPluginsRoot`, `bundledPluginsRoot`, `listBundledPlugins`
- Updates README.md plugin section with install/uninstall flow and SOURCE column documentation

## Acceptance Criteria

- [x] `engram plugin install gerrit` creates a valid XDG entry (symlink) — works once `packages/plugins/gerrit/` exists; the mechanism is fully implemented
- [x] `engram plugin install nonexistent` exits non-zero with `"no bundled plugin named 'nonexistent'; see \`engram plugin list --available\`"`
- [x] `engram plugin uninstall gerrit` reverses the install (removes symlink or bundled-copy directory)
- [x] `engram plugin uninstall <user-authored>` exits non-zero without `--force`
- [x] `engram plugin list` shows SOURCE column distinguishing bundled/symlinked/user installs
- [x] `--project` flag installs into `.engram/plugins/` (project-local, higher precedence per ADR-006)
- [x] Works on Linux/macOS (symlink); Windows fallback to recursive copy when symlink fails
- [x] README.md plugin section updated with install/uninstall flow
- [x] Tests cover install, uninstall, collision (already installed), missing-source, user-authored guard, and project-vs-user scope (14 tests)

## Test plan

- [x] `bun test packages/engram-cli/test/commands/plugin.test.ts` — 14 pass, 0 fail
- [x] `bun run build` — clean build
- [x] `bun run lint` — no errors

Closes #223

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,7 +165,9 @@ that current code alone does not reveal.
 | `engram verify` | Validate `.engram/` integrity. |
 | `engram embed` | Manage vector embeddings. |
 | `engram rebuild-index` | Rebuild the FTS index. |
-| `engram plugin list` | List discovered plugins. |
+| `engram plugin list` | List discovered plugins (add `--available` for installable bundled plugins). |
+| `engram plugin install <name>` | Wire a bundled first-party plugin into XDG (or project with `--project`). |
+| `engram plugin uninstall <name>` | Remove an installed plugin (`--force` for user-authored). |
 
 Every command has `--help` with examples.
 
@@ -332,11 +334,46 @@ extensions, and security model.
 ```bash
 engram plugin list
 
-NAME        VERSION  TRANSPORT   SCOPE    STATUS
-----------  -------  ----------  -------  ------
-my-jira     0.1.0    executable  user     OK
-my-linear   0.2.0    js-module   project  OK
+NAME        VERSION  TRANSPORT   SCOPE    SOURCE    STATUS
+----------  -------  ----------  -------  --------  ------
+my-jira     0.1.0    executable  user     user      OK
+my-linear   0.2.0    js-module   project  symlinked OK
 ```
+
+### Install and uninstall
+
+First-party bundled plugins (those shipped in the engram install under
+`packages/plugins/`) can be wired into the XDG plugin directory with a
+single command — no manual symlinking required:
+
+```bash
+# List what is available to install
+engram plugin list --available
+
+# Install a bundled plugin user-wide (symlink into XDG_DATA_HOME/engram/plugins/)
+engram plugin install gerrit
+
+# Install into a specific project only
+engram plugin install gerrit --project /path/to/project
+
+# Uninstall a bundled or symlinked plugin
+engram plugin uninstall gerrit
+
+# Uninstall a user-authored (plain-directory) plugin — requires --force
+engram plugin uninstall my-custom --force
+```
+
+**Source column** — `engram plugin list` reports how each plugin was
+installed:
+
+| Source | Meaning |
+|---|---|
+| `bundled` | Symlink pointing into the engram install root (installed via `engram plugin install`). |
+| `symlinked` | A user-created symlink to an external directory. |
+| `user` | A plain directory created by the user. |
+
+On Windows, where symlinks may require elevated permissions, `engram plugin
+install` falls back to a recursive copy automatically.
 
 ## AI providers
 
--- a/packages/engram-cli/src/commands/plugin.ts
+++ b/packages/engram-cli/src/commands/plugin.ts
@@ -2,25 +2,80 @@
  * plugin.ts — `engram plugin` command group.
  *
  * Subcommands:
- *   - plugin list   Discover and display installed plugins with status
+ *   - plugin list      Discover and display installed plugins with status
+ *   - plugin install   Wire a bundled first-party plugin into XDG (or project)
+ *   - plugin uninstall Remove a previously installed plugin
  */
 
+import * as fs from "node:fs";
 import * as path from "node:path";
 import { log } from "@clack/prompts";
 import type { Command } from "commander";
 import {
+  bundledPluginsRoot,
   discoverPlugins,
+  listBundledPlugins,
   loadManifest,
   ManifestValidationError,
+  projectPluginsRoot,
+  userPluginsRoot,
 } from "engram-core/plugins";
+
+// ---------------------------------------------------------------------------
+// Shared helpers
+// ---------------------------------------------------------------------------
+
+/**
+ * Resolves the destination plugins root based on whether --project is set.
+ */
+function resolveDestRoot(projectOpt: string | undefined): string {
+  if (projectOpt) {
+    return projectPluginsRoot(path.resolve(projectOpt));
+  }
+  return userPluginsRoot();
+}
+
+/**
+ * On Windows, symlinks may require elevated permissions. Fall back to a
+ * directory copy when symlinking fails.
+ */
+function installPlugin(src: string, dest: string): "symlink" | "copy" {
+  fs.mkdirSync(path.dirname(dest), { recursive: true });
+
+  try {
+    fs.symlinkSync(src, dest, "junction");
+    return "symlink";
+  } catch {
+    // Fallback: recursive copy
+    copyDir(src, dest);
+    return "copy";
+  }
+}
+
+function copyDir(src: string, dest: string): void {
+  fs.mkdirSync(dest, { recursive: true });
+  for (const entry of fs.readdirSync(src, { withFileTypes: true })) {
+    const srcPath = path.join(src, entry.name);
+    const destPath = path.join(dest, entry.name);
+    if (entry.isDirectory()) {
+      copyDir(srcPath, destPath);
+    } else {
+      fs.copyFileSync(srcPath, destPath);
+    }
+  }
+}
+
+// ---------------------------------------------------------------------------
+// plugin list
+// ---------------------------------------------------------------------------
 
 interface PluginListOpts {
   project?: string;
+  available?: boolean;
+  bundledRoot?: string;
 }
 
-export function registerPlugin(program: Command): void {
-  const plugin = program.command("plugin").description("Manage engram plugins");
-
+function registerList(plugin: Command): void {
   plugin
     .command("list")
     .description("List discovered plugins and their status")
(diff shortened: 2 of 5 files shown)


rnwolfe:
## Code Review — feat(cli): engram plugin install and uninstall subcommands

Overall the feature is well-structured and the 14-test suite covers the main happy/unhappy paths. There are several findings to address before merging, ranging from a security issue to logic gaps.

---

### [BLOCKING] Path traversal in `installPlugin` / `copyDir` — `plugin.ts` lines 42–66

`installPlugin` accepts `name` directly from the CLI argument and joins it into `dest` without sanitizing it first (line 263: `const dest = path.join(destRoot, name)`). `copyDir` then recursively copies the resolved source into that path.

A malicious name like `../../.ssh/authorized_keys-plugin` would escape the intended plugins root. The check `fs.existsSync(dest)` on line 265 does not protect against traversal because `path.join` resolves `..` segments.

**Fix:** validate `name` before use — reject anything that contains `/`, `\`, or `..` segments, or assert that the resolved destination is a child of `destRoot`:

```ts
const dest = path.join(destRoot, name);
if (!dest.startsWith(path.resolve(destRoot) + path.sep)) {
  process.stderr.write(`error: invalid plugin name '${name}'\n`);
  process.exitCode = 1;
  return;
}
```

The same pattern applies to `uninstall` (line 309: `const dest = path.join(destRoot, name)`).

---

### [BLOCKING] `installPlugin` always uses `"junction"` symlink type — `plugin.ts` line 46

`fs.symlinkSync(src, dest, "junction")` is called unconditionally on all platforms. On Linux/macOS `"junction"` is silently treated as a directory symlink, so the call succeeds but `junction` is not meaningful on POSIX. The correct POSIX value is `"dir"` for directory symlinks. This is a portability bug — the intent is clear but the implementation is wrong on the primary development platform:

```ts
// Use 'junction' on Windows, 'dir' on POSIX
const symlinkType = process.platform === "win32" ? "junction" : "dir";
fs.symlinkSync(src, dest, symlinkType);
```

The fallback-to-copy comment says "Windows may require elevated permissions", but the junction type is the Windows-only workaround. Using it on Linux means the catch block is never exercised on Linux, which is likely why tests pass — the symlink succeeds — but it's semantically wrong and could behave differently on older Node/Bun POSIX builds.

---

### [BLOCKING] `--available` early return skips the normal `list` output — `plugin.ts` line 112

`plugin list --available` returns **only** the bundled list, not the installed plugins. The flag name and docs say "also list", but the implementation returns early after printing bundled plugins and never prints installed ones (line 112: `return`). The README also says "List what is available to install" — so the behaviour matches the docs but the flag name (`--available`) implies addition. Either:
- Rename the flag to `--list-available` or make the description say "show only bundled plugins available to install", or
- Remove the early `return` and show both sections.

The current combination of flag name, help text ("also list bundled first-party plugins available to install") and `return` is self-contradictory. This will confuse users.

---

### [NON-BLOCKING] `bundledPluginsRoot` walks up 8 levels with `__dirname` check — `discover.ts` lines 169–202

The `typeof __dirname !== "undefined"` guard is needed because Bun ESM modules do not provide `__dirname`. However, when the guard falls back to `process.cwd()`, the upward walk starts from the *working directory*, not the module location. In a monorepo where the user runs `bun` from a subdirectory, this produces a wrong starting path and may fail to find `packages/plugins/`. Consider using `import.meta.url` (available in Bun/ESM):

```ts
import { fileURLToPath } from "node:url";
const __filename = fileURLToPath(import.meta.url);
const startDir = path.dirname(__filename);
```

This is more reliable in ESM than the `__dirname` presence check.

---

### [NON-BLOCKING] `detectSource` calls `fs.realpathSync` inside an IIFE on every call — `discover.ts` lines 82–88

```ts
const realBundled = (() => {
  try {
    return fs.realpathSync(bundledPluginsRoot);
  } catch {
    return bundledPluginsRoot;
  }
})();
```

`bundledPluginsRoot` is the same string for every plugin in a single `discoverPlugins` call, so this realpath resolution is repeated N times. It should be resolved once before the loop (e.g., cache the result in `listPluginsIn` and pass it pre-resolved to `detectSource`). Low overhead in practice, but unnecessary.

---

### [NON-BLOCKING] `uninstall` guard for bundled-copy is name-only — `plugin.ts` lines 336–347

```ts
const isBundledCopy = available.includes(name);
if (!isBundledCopy && !opts.force) { ... }
```

This allows removal of a plain-directory plugin whose name happens to match a bundled plugin name, even if the directory is completely unrelated to the bundled one (e.g., the user created `~/.local/share/engram/plugins/gerrit/` manually with different contents). In practice this is unlikely to cause harm, but it means the guard is purely name-based, not content-based. A comment explaining this intentional limitation would help future maintainers.

---

### [NON-BLOCKING] `plugin list --available` exits without showing installed plugins — UX gap

If a user runs `engram plugin list --available` they get available bundled plugins but no indication of what is already installed. Operators who want to audit installed+available state must run two commands. Consider printing the installed list first, then the available section, rather than early-returning.

---

### [NON-BLOCKING] Test: `process.exitCode` left as `0` after negative tests — `plugin.test.ts` lines 316, 355, 400, etc.

The pattern:
```ts
process.exitCode = 0;
```
in `finally` blocks (actually in the `try` after assertions) resets `exitCode` but only when the assertion passed. If an assertion throws, `exitCode` is left as `1` for subsequent tests. The reset should be in `finally`, not inline after `expect()`:

```ts
try {
  // ...test...
} finally {
  process.exitCode = 0; // reset regardless
  restoreXdg();
  fs.rmSync(...);
}
```

This is a test hygiene issue that could cause false positives in test runs where assertion order matters.

---

### [NON-BLOCKING] `copyDir` does not preserve file permissions or symlinks within the plugin — `plugin.ts` lines 55–66

`fs.copyFileSync` copies file contents but not permissions (mode bits) or embedded symlinks inside the plugin directory. If a bundled plugin ships executable scripts or internal symlinks, the Windows-fallback copy will break those. Consider `fs.copyFileSync(srcPath, destPath, fs.constants.COPYFILE_FICLONE_FORCE)` for CoW filesystems, and handle `entry.isSymbolicLink()` explicitly.

---

### [MINOR] `plugin.ts` comment block header still lists only `plugin list` — line 5

The top-of-file JSDoc comment was updated in the diff but line 5 still says the old description. Confirmed the new file is correct (lines 5–7 show all three subcommands). No action needed — just confirming this is clean.

---

### Summary

| # | Severity | Finding |
|---|----------|---------|
| 1 | BLOCKING | Path traversal via unsanitized `name` in install and uninstall |
| 2 | BLOCKING | `"junction"` symlink type used on all platforms (wrong on POSIX) |
| 3 | BLOCKING | `--available` flag name is self-contradictory with early-return behaviour |
| 4 | Non-blocking | `bundledPluginsRoot` discovery falls back to `process.cwd()` in ESM |
| 5 | Non-blocking | Redundant `realpathSync` inside per-plugin `detectSource` loop |
| 6 | Non-blocking | Bundled-copy uninstall guard is name-only, not content-verified |
| 7 | Non-blocking | `process.exitCode` reset placement in test teardown |
| 8 | Non-blocking | `copyDir` does not preserve permissions or internal symlinks |

Items 1–3 should be fixed before merging. Items 4–8 are recommended improvements but not blockers.

rnwolfe:
Addressed all 3 blocking findings from review:

1. **Path traversal** — added guard in both `install` and `uninstall`: resolve `destRoot` with `path.resolve()`, then assert `dest.startsWith(destRoot + path.sep)` before any fs operation. Returns exit code 1 with a clear error on traversal attempt.

2. **Symlink type** — branched on `process.platform === 'win32'` to use `'junction'` on Windows and `'dir'` on POSIX. The Windows-only fallback-to-copy path is now actually exercised on Windows when junction creation fails.

3. **`--available` early return** — removed the `return` after the bundled-plugins section; now falls through to the installed-plugins table. Both sections appear when `--available` is passed.

All 14 tests pass. Build and lint clean.

PR #242, opened 2026-04-21
refactor: generalize orchestrator to accept extractor-declared entities and edges

### LINKED ISSUE #241: refactor(source): generalize orchestrator to accept extractor-declared entities and edges
Prerequisite for #236 (Kubernetes operator graph support) and child issues #237, #238, #239.

## Summary

The source ingestion orchestrator (`packages/engram-core/src/ingest/source/index.ts`) is tightly coupled to the TypeScript/Go/Python extractor shape: every `ExtractedFile` is read as `{symbols, rawImports}`, every symbol is written with `entity_type: SYMBOL`, and edges are hardcoded to `CONTAINS` / `DEFINED_IN` / `IMPORTS`. This refactor generalizes the orchestrator to accept extractor-declared entities and edges so that new extractors can emit domain-specific vocabulary (Bazel targets, RBAC permissions, resource kinds, etc.) without modifying the orchestrator further.

- **Module location**: `packages/engram-core/src/ingest/source/extractors/types.ts`, `packages/engram-core/src/ingest/source/index.ts`
- **Type extension**: `ExtractedFile` gains two optional fields:
  ```ts
  interface ExtractedFile {
    symbols: ExtractedSymbol[];
    rawImports: string[];
    extraEntities?: ExtractedEntity[];   // NEW
    extraEdges?: ExtractedEdge[];        // NEW
  }

  interface ExtractedEntity {
    canonicalName: string;
    entityType: EntityType;       // from vocab registry
  }

  type EntityRef =
    | { kind: "file" }                                            // → current file entity
    | { kind: "symbol"; name: string }                            // → symbol in this file by name
    | { kind: "canonical"; canonicalName: string; entityType: EntityType };

  interface ExtractedEdge {
    source: EntityRef;
    target: EntityRef;
    relationType: RelationType;   // from vocab registry
    edgeKind: "observed" | "inferred" | "asserted";
    fact: string;
  }
  ```
- **Orchestrator pass**: after the existing symbols+imports passes, loop over `extraEntities` (upsert via existing `upsertEntity`) and `extraEdges` (resolve `EntityRef` to ULID via a lookup table built from this file's entities, then call `upsertEdge`). All writes use the same evidence chain as symbols from the same file
- **Backward compat**: existing extractors keep returning `{symbols, rawImports}` only. `extraEntities` / `extraEdges` default to empty arrays. Zero behavior change for current TS/Go/Python ingestion
- **dryRun**: counts of extra entities and edges included in `entitiesCreated` / `edgesCreated`
- **Storage**: no schema changes

## Integration Points

- `packages/engram-core/src/ingest/source/extractors/types.ts` — extend `ExtractedFile` and export new helper types
- `packages/engram-core/src/ingest/source/index.ts` — new orchestrator pass and `EntityRef` resolution
- `packages/engram-core/src/vocab/` — no change in this issue; children add entity/relation types as needed
- Unblocks: #237 (Bazel), #238 (kubebuilder RBAC), #239 (controller-runtime watches)

## Acceptance Criteria

- [ ] `ExtractedFile` type extended with optional `extraEntities` and `extraEdges`, with supporting `ExtractedEntity`, `ExtractedEdge`, `EntityRef` types exported
- [ ] Orchestrator materializes `extraEntities` (upsert by `canonicalName`/`entityType`) after file+symbol passes
- [ ] Orchestrator resolves `EntityRef` for edge sources/targets: `file` → current file entity, `symbol` → symbol by name in this file (returns error if name not extracted), `canonical` → lookup by `(canonicalName, entityType)`, upserting if not present
- [ ] All extra entity/edge writes share the same `EvidenceInput` as the file's symbols (one episode per file)
- [ ] `dryRun` counts extra entities and edges in result totals
- [ ] Existing TS/Go/Python ingestion behavior unchanged — snapshot test against self-ingest output shows identical entity/edge counts pre- and post-refactor
- [ ] Unit tests in `packages/engram-core/test/source/orchestrator-extras.test.ts` covering: single `extraEntity`, edge referencing `file`, edge referencing `symbol`, edge referencing `canonical` (cross-file), `symbol` ref that doesn't match any extracted symbol (error), upsert idempotency
- [ ] `verifyGraph()` passes after ingestion with extra entities/edges

## Documentation

- [ ] JSDoc on new types in `types.ts` — especially `EntityRef` resolution semantics

rnwolfe:
## Summary

- Extends `ExtractedFile` in `extractors/types.ts` with optional `extraEntities?: ExtractedEntity[]` and `extraEdges?: ExtractedEdge[]` fields, along with new `ExtractedEntity`, `EntityRef`, and `ExtractedEdge` types exported with full JSDoc on resolution semantics
- Adds `resolveEntityRef()` helper (exported for testing) that maps `EntityRef` variants (`file`, `symbol`, `canonical`) to ULIDs, upserting new canonical entities when needed using the same evidence chain as the file's symbols
- Orchestrator materializes `extraEntities` and `extraEdges` after the standard symbol pass; `dryRun` mode counts extras in result totals without DB writes
- Adds `test/ingest/source/orchestrator-extras.test.ts` covering all EntityRef resolution paths, idempotency, and regression guard for existing TS/Go/Python ingest behavior

## Acceptance Criteria

- [x] `ExtractedFile` extended with optional `extraEntities` and `extraEdges`, with supporting types exported
- [x] Orchestrator materializes `extraEntities` (upsert by `canonicalName`/`entityType`) after file+symbol passes
- [x] Orchestrator resolves `EntityRef` for edge sources/targets
- [x] All extra entity/edge writes share the same `EvidenceInput` as the file's symbols
- [x] `dryRun` counts extra entities and edges in result totals
- [x] Existing TS/Go/Python ingestion behavior unchanged (140 source tests pass)
- [x] Unit tests pass
- [x] `verifyGraph()` passes after ingestion with extra entities/edges
- [x] JSDoc on new types especially `EntityRef` resolution semantics

Closes #241


rnwolfe (commented):
## Code Review — PR #242

**Summary:** 1 blocking, 3 non-blocking findings.

---

### BLOCKING: Test isolation broken — `process.chdir` calls without CWD restore

The PR's premise is correct: `doctor.ts` no longer needs the caller's CWD because it derives `cwd` from the DB path. But the test file was only half-updated. The `origCwd` saves and `process.chdir(origCwd)` restores were removed, yet the `process.chdir(tmpDir)` calls themselves were left in **14 tests**:

- layout `it("fails when .engram is a flat file")` — line 123
- gitignore `it("fails when .gitignore contains .engram (flat-file entry)")` — line 193
- schema, fts_index, embedding_index, wal, evidence_integrity, fix-* tests — lines 236–720

Each of those tests mutates the global process CWD and never restores it. When `afterEach` deletes `tmpDir`, the process's CWD points to a now-deleted directory. Any subsequent test (or the test framework itself) that tries to do a relative-path FS operation will fail with `ENOENT`.

**Fix:** Either remove the `process.chdir(tmpDir)` calls from every test that no longer needs them (they don't, since all `--db` args are already absolute paths), or add a shared `beforeEach`/`afterEach` that saves and restores CWD. Removing the calls is cleaner given the doctor fix.

---

### Non-blocking: Go and Python import edges will always be empty

`resolveImport` is hardwired for TypeScript/JS resolution: it returns `null` for any specifier that doesn't start with `./` or `../`. Go import paths (`"fmt"`, `"github.com/pkg/foo"`) and Python dotted names (`os`, `pathlib`) will all hit that early-return and produce zero import edges. The `rawImports` are collected correctly but are silently dropped at the resolution pass.

This is consistent with the current limitation (only intra-repo relative imports are tracked for TS/JS), but it should be documented — right now nothing in the Go/Python extractors or the query files signals that import edge resolution is a no-op for those languages. A brief comment in the import resolution pass or in the extractor JSDoc would prevent future confusion.

---

### Non-blocking: `upsertEntity` matches only on `canonical_name`, ignores `entity_type`

In `index.ts`, `upsertEntity` calls `findEntities(graph, { canonical_name: input.canonical_name })` without filtering on `entity_type`. If two extractors produce a `canonical` EntityRef for the same name but different entity types, the first-inserted type wins silently. This pre-exists the PR but `resolveEntityRef(kind="canonical")` now gives arbitrary extractor code a direct path to trigger this mismatch. Worth a guard (or at minimum a `console.warn`) when the found entity's `entity_type` doesn't match the requested one.

---

### Non-blocking: Go query missing `var` declarations

`go.scm` captures `function_declaration`, `type_declaration`, and `const_declaration`, but not `var_declaration`. Package-level vars are idiomatic Go and commonly exported (e.g. `var ErrNotFound = errors.New(...)`). The KIND_MAP in `go.ts` doesn't have a `var` entry either, so extending this requires touching both files. Not blocking, but a gap worth tracking.

rnwolfe (commented):
## Review response

### BLOCKING: process.chdir without CWD restore

Fixed in d98c444. Removed all 14 `process.chdir(tmpDir)` calls from `packages/engram-cli/test/commands/doctor.test.ts`. Since every test already passes `--db` as an absolute path, the CWD is irrelevant to the command logic — the calls were vestigial. All 18 tests pass after the removal.

### Non-blocking #1: Go/Python import edges silently empty

Acknowledged. This is pre-existing behavior inherited from the TS/JS resolver: `resolveImport` returns `null` for any specifier that isn't `./`- or `../`-relative, so Go (`"fmt"`, `"github.com/pkg/foo"`) and Python (`os`, `pathlib`) imports are collected by the extractors but dropped at the resolution pass without any warning. Tracked for a follow-up — the right fix is a language-aware resolution pass or a no-op resolver per language that at minimum emits a debug-level log so the gap is visible in output.

### Non-blocking #2: upsertEntity ignores entity_type in lookup

Acknowledged as pre-existing, out of scope for this PR. The silent first-wins behavior when two extractors produce the same `canonical_name` with different `entity_type` values is a real risk worth a guard or `console.warn`. Will track separately.

### Non-blocking #3: Go query missing var_declaration

Acknowledged as pre-existing, out of scope for this PR. `var_declaration` at package level (e.g. `var ErrNotFound = errors.New(...)`) is idiomatic exported Go and should be captured. Both `go.scm` and the `KIND_MAP` in `go.ts` need to be updated together. Will track separately.

PR #244, opened 2026-04-21
feat(source): add Java tree-sitter parser

### LINKED ISSUE #233: feat(source): add Java tree-sitter parser
## Summary

Add Java source code parsing support to the tree-sitter ingestion layer, following the same pattern as the Go and Python parsers.

## Acceptance Criteria

- [ ] Download `tree-sitter-java.wasm` from https://github.com/tree-sitter/tree-sitter-java/releases (version compatible with web-tree-sitter 0.25.x)
- [ ] Add `java` to the `Language` type in `parser.ts` with `.java` extension mapping
- [ ] Create `queries/java.scm` capturing top-level `class`, `interface`, `enum`, `record`, `method_declaration`, `field_declaration` symbols and `import_declaration` imports
- [ ] Create `extractors/java.ts` with `extractJava()` — exported = has `public` modifier
- [ ] Wire into `index.ts` via `extractFile()`
- [ ] Update `grammars/MANIFEST.json` and `grammars/README.md`
- [ ] Add tests in `test/ingest/source/extractor-java.test.ts`
- [ ] All existing tests continue to pass

## Implementation Notes

Java visibility is modifier-based. The query should use capture name suffixes to distinguish public vs non-public symbols (e.g. `@symbol.class.exported`). Java import paths are dotted names without quotes.

Reference: `packages/engram-core/src/ingest/source/extractors/go.ts` for pattern.

rnwolfe:
## Summary

- Downloads `tree-sitter-java.wasm` v0.23.5 (compatible with web-tree-sitter 0.25.x) into `grammars/`
- Adds `java` to the `Language` type in `parser.ts` with `.java` extension mapping
- Creates `queries/java.scm` capturing top-level `class_declaration`, `interface_declaration`, `enum_declaration`, `record_declaration`, `method_declaration`, and `field_declaration` symbols plus `import_declaration` imports
- Creates `extractors/java.ts` with `extractJava()` — uses single-capture `.vis` pattern (same as Rust extractor) to detect `public` modifier; handles the `field_declaration → variable_declarator → identifier` nesting for field symbols
- Wires into `index.ts` via `extractFile()`
- Updates `grammars/MANIFEST.json` and `grammars/README.md`
- Adds 26 tests in `test/ingest/source/extractor-java.test.ts` covering all symbol kinds, visibility detection, imports, byte offsets, and nested-class exclusion

## Acceptance Criteria

- [x] `tree-sitter-java.wasm` v0.23.5 in `grammars/` (compatible with web-tree-sitter ^0.25.6)
- [x] `java` added to `Language` type with `.java` extension
- [x] `queries/java.scm` captures class, interface, enum, record, method, field symbols and imports
- [x] `extractors/java.ts` with `public` modifier → `exported: true`; all others → `exported: false`
- [x] Wired in `index.ts`
- [x] `MANIFEST.json` and `README.md` updated
- [x] 26 tests pass
- [x] All existing tests continue to pass (3 pre-existing failures unchanged)

## Test plan

- [x] `bun test packages/engram-core/test/ingest/source/extractor-java.test.ts` — 26/26 pass
- [x] `bun test` — no new failures introduced
- [x] `bun run build` — clean

Closes #233


rnwolfe (commented):
## Code Review — feat(source): add Java tree-sitter parser

Overall the implementation is solid and follows established extractor patterns well. I found one correctness bug (blocking), a couple of gaps in the tree-sitter query (non-blocking), and a few minor issues.

---

### BLOCKING

#### 1. `hasPublicModifier` splits on whitespace but `modifiers` is a tree-sitter subtree text, not a flat token list

`java.ts` line 16:
```ts
function hasPublicModifier(modifiersText: string): boolean {
  return modifiersText.split(/\s+/).includes("public");
}
```

In the Java tree-sitter grammar the `(modifiers)` node is the full text of the modifier block — e.g. `"@Override\npublic static final"` or `"@SuppressWarnings(\"all\")\npublic"`. Splitting on whitespace is almost correct, but **annotations with arguments** can introduce parenthesised text that includes arbitrary sub-tokens, such as `@SuppressWarnings("all")`. The split would produce tokens like `@SuppressWarnings("all")` which don't match `"public"`, so in practice the check still works for that case.

The real hazard is the opposite: **a symbol name or string literal containing the word `public`** could theoretically match. This is extremely unlikely in modifiers text but the correct approach — consistent with how `isBarePub` works in the Rust extractor — is to check with a word-boundary regex:

```ts
function hasPublicModifier(modifiersText: string): boolean {
  return /\bpublic\b/.test(modifiersText);
}
```

This is both simpler and more robust. I'm flagging this as blocking because the current implementation could silently misclassify visibility for annotated declarations in adversarial-ish code.

---

### NON-BLOCKING

#### 2. `visMap` key for field symbols traverses two parent levels but only one is stored

`java.ts` lines 58–64 walk up to `node.parent?.parent` for field symbols to account for the `variable_declarator` nesting. This is correct given the query shape:

```
field_declaration
  └─ variable_declarator
       └─ identifier   ← @symbol.field  (node)
```

And the vis capture is stored keyed by `field_declaration.startIndex` (parent of `variable_declarator`). The two-level walk is right. This is fine — just worth a clarifying comment that `node.parent` is `variable_declarator` and `node.parent.parent` is `field_declaration`.

#### 3. `seenNames` de-duplicates by bare name, not by qualified name

If a Java file has a class-level field named `value` and a different method named `value` (unlikely but legal in Java for fields vs methods), de-duplication by bare name would silently drop one. The other extractors share this limitation, so it's a known project-wide issue, not a regression introduced here.

#### 4. No test for `public static` / `public final` / annotated-public combinations

The test suite covers `public`, `private`, `protected`, and package-private, which is good. However there's no test case for compound modifiers like:

```java
public class MyClass {
  public static final int CONSTANT = 42;
  @Override
  public void annotatedMethod() {}
}
```

These are the most common real-world forms. Given the `hasPublicModifier` concern in item 1, add at least one test exercising a multi-modifier field or annotated public method to confirm `exported: true` is correctly detected.

#### 5. `java.scm` query comment is misleading for methods/fields

```scm
; --- Top-level method declarations (inside class body, depth 1) ---
```

These are not "top-level" in the Java sense — they are class members. The comment correctly says "depth 1" but calling them top-level conflicts with the Java terminology where "top-level" means "not inside another type". Consider rewording to "Class member method declarations (depth 1, not nested)" to avoid confusion with top-level class declarations.

#### 6. `export type { ExtractedFile, ExtractedSymbol }` in `java.ts` is unused

Lines 3–4 in `java.ts`:
```ts
export type { ExtractedFile, ExtractedSymbol };
```

None of the other extractors re-export these types (the rust and go extractors do the same, so this is a project-wide pattern, but it's dead surface area). Non-blocking — consistent with existing extractors.

#### 7. Import resolution pass for Java is no-op but not documented

`index.ts` feeds `rawImports` into `resolveImport()` (from the TypeScript extractor). Java imports like `java.util.List` will never resolve to a file path via that function, so all Java imports will silently produce no edges. This is acceptable for v1 but worth a comment in `index.ts` near the dispatch table (or in the Java extractor itself) noting that Java import resolution is not yet implemented and the raw import strings are retained for future use.

#### 8. Biome line-length compliance

`java.ts` line 49 (the `console.warn` call) fits within 80 chars after the string concatenation. Line 48–51 pattern is identical to the Rust extractor so no new issue here. No Biome violations detected.

---

### Summary

| 1 | **Blocking** | `hasPublicModifier` should use `/\bpublic\b/` regex, not `.split().includes()` |
| 2 | Non-blocking | Two-level parent walk for fields is correct but needs a clarifying comment |
| 3 | Non-blocking | `seenNames` deduplicates by bare name (pre-existing limitation) |
| 4 | Non-blocking | Missing tests for `public static`, `public final`, and annotated-public members |
| 5 | Non-blocking | Query file comments call class members "top-level" |
| 6 | Non-blocking | Re-exported types are unused surface area (matches existing pattern) |
| 7 | Non-blocking | Java import resolution silently no-ops — worth a TODO comment |

**One blocking issue** (item 1). The fix is a one-liner. Everything else is correct and the overall structure cleanly follows the established extractor pattern.

rnwolfe (commented):
Thanks for the thorough review. Fixed the blocking item and addressed the non-blocking ones:

**Item 1 (Blocking — FIXED):** Replaced `.split(/\s+/).includes("public")` with `/\bpublic\b/.test(modifiersText)` in `hasPublicModifier`. Commit dc52ee5.

**Item 4 (Non-blocking — FIXED):** Added a test case for an annotated public method (`@Override\npublic String toString()`) asserting `exported: true`. All 27 tests pass.

**Item 2 (Non-blocking — acknowledged):** The two-level parent walk is correct. I'll leave a clarifying comment in a follow-up if desired; it's a known-correct pattern consistent with the existing code.

**Items 3, 5, 6, 7 (Non-blocking — acknowledged):** Noted as pre-existing/project-wide patterns or future work. No action taken in this PR to keep the diff focused.

Build and lint clean on the changed files. Ready for re-review.

PR #245, opened 2026-04-21
feat(source): add Ruby, C, C++, and C# tree-sitter parsers

### LINKED ISSUE #234: feat(source): add Ruby, C/C++, and C# tree-sitter parsers
## Summary

Add source parsing for Ruby, C/C++, and C# following the established Go/Python pattern. These can be done together since they share the same implementation shape.

## Languages and Extensions

| Language | Extensions | Grammar repo |
|----------|------------|--------------|
| Ruby | `.rb` | `tree-sitter/tree-sitter-ruby` |
| C | `.c`, `.h` | `tree-sitter/tree-sitter-c` |
| C++ | `.cpp`, `.cc`, `.cxx`, `.hpp` | `tree-sitter/tree-sitter-cpp` |
| C# | `.cs` | `tree-sitter/tree-sitter-c-sharp` |

## Acceptance Criteria (per language)

- [ ] Download WASM from GitHub releases (web-tree-sitter 0.25.x compatible)
- [ ] Add language to `Language` type in `parser.ts` with extension mappings
- [ ] Create `queries/<lang>.scm` for top-level symbols and imports
- [ ] Create `extractors/<lang>.ts` with export detection appropriate to language
- [ ] Wire into `extractFile()` in `index.ts`
- [ ] Update `grammars/MANIFEST.json` and `grammars/README.md`
- [ ] Tests in `test/ingest/source/extractor-<lang>.test.ts`

## Notes

- C/C++: no formal export mechanism; treat all top-level declarations as exported.
- C#: exported = `public` modifier present.
- Ruby: exported = method/class is not inside `private`/`protected` block (heuristic: treat all module-level as exported for first pass).

Reference: `packages/engram-core/src/ingest/source/` for all existing patterns.

rnwolfe:
## Summary

- Adds four new tree-sitter language parsers: Ruby, C, C++, and C#, each following the established pattern (WASM grammar, `.scm` query, TypeScript extractor, colocated tests)
- Ruby: extracts `class`/`module`/`method`/`singleton_method` symbols and `require`/`require_relative` imports; all module-level symbols treated as exported
- C: extracts `function_definition`/`struct_specifier`/`enum_specifier`/`typedef_declaration` + `#include` paths; all top-level declarations are exported
- C++: extracts `function_definition`/`class_specifier`/`struct_specifier`/`enum_specifier`/`namespace_definition` + `#include` paths
- C#: extracts `class_declaration`/`interface_declaration`/`enum_declaration`/`struct_declaration`/`record_declaration`/`method_declaration` + `using` directives; exported = `public` modifier (same word-boundary regex as Java)

## Test plan

- [ ] 54 new extractor tests across 4 files (`extractor-ruby.test.ts`, `extractor-c.test.ts`, `extractor-cpp.test.ts`, `extractor-c-sharp.test.ts`)
- [ ] 10 new tests added to `parser.test.ts` covering `languageForPath` extension mappings and `SourceParser.parse()` root node assertions for all four languages
- [ ] All 267 source ingest tests pass (`bun test packages/engram-core/test/ingest/source/`)
- [ ] Build succeeds (`bun run build`)
- [ ] Lint clean in all new/modified files (`bun run lint`)

Closes #234


rnwolfe (commented):
## Code Review — PR #245: Ruby, C, C++, and C# tree-sitter parsers

Overall the PR follows the established extractor pattern well. Ruby, C, and C++ are correct. There is one **blocking bug** in the C# extractor affecting multi-modifier declarations, and a handful of non-blocking observations.

---

### BLOCKING

#### C# `visMap` overwrites on multi-modifier declarations

**File:** `packages/engram-core/src/ingest/source/extractors/c_sharp.ts` (lines 34–38)  
**Query:** `packages/engram-core/src/ingest/source/queries/c_sharp.scm` (line 3: `(modifier)* @symbol.class_declaration.vis`)

Unlike Java, which has a single aggregate `(modifiers)?` node, C#'s grammar uses individual `modifier` child nodes. A class like `public static class Foo {}` generates **three** captures all with `node.parent.startIndex === 0`:

```
symbol.class_declaration.vis: "public"  (startIndex 0, parent startIndex 0)
symbol.class_declaration.vis: "public"  (startIndex 0, parent startIndex 0)  ← duplicate
symbol.class_declaration.vis: "static"  (startIndex 7, parent startIndex 0)
```

The current extractor uses `visMap.set(node.parent.startIndex, node.text)` so each modifier **overwrites** the previous one. The final value stored is `"static"`, and `hasPublicModifier("static")` returns `false` — causing `public static class Foo` to be marked `exported: false`.

This bug affects any declaration with two or more modifiers where `public` is not the last one (e.g., `public static`, `public abstract`, `public sealed`, `public readonly struct`, `public override`).

**Fix:** Accumulate modifier text rather than overwrite. For example, use `Map<number, string[]>` and join on lookup, or concatenate with a space:

```typescript
// First pass — accumulate all modifier texts
const visMap = new Map<number, string>();
for (const capture of captures) {
  const { name: captureName, node } = capture;
  if (captureName.endsWith(".vis") && node.parent) {
    const key = node.parent.startIndex;
    const existing = visMap.get(key);
    visMap.set(key, existing ? `${existing} ${node.text}` : node.text);
  }
}
```

This matches the Java extractor's semantics (where `modText` is the full `modifiers` node text including all modifiers) without any query changes.

**Test gap:** No test covers `public static class`, `public abstract class`, or any other multi-modifier declaration. A test like `"exported = true for public static class"` should be added.

---

### NON-BLOCKING

#### C# MANIFEST.json has empty `commit` hashes

**File:** `packages/engram-core/src/ingest/source/grammars/MANIFEST.json`

The four new grammar entries have `"commit": ""`. The existing languages all have full SHA commits for reproducibility and auditability. This doesn't affect runtime behavior but is inconsistent with the established convention and reduces traceability.

---

#### Ruby test does not assert `Outer::Inner` extraction

**File:** `packages/engram-core/test/ingest/source/extractor-ruby.test.ts` (line 21–33)

The test source includes `class Outer::Inner` but only asserts that `"MyClass"` is present. The query correctly captures `scope_resolution` names as `"Outer::Inner"` (full text), but there is no explicit assertion on this. Worth adding `expect(names).toContain("Outer::Inner")` to document and lock in this behavior.

---

#### `using static` not tested for C#

**File:** `packages/engram-core/test/ingest/source/extractor-c-sharp.test.ts`

The `using_directive` query correctly handles `using static System.Math;` (the `qualified_name` node is captured), but this form is not covered by any test. Non-blocking since the query is structurally sound, but a test would prevent regression.

---

#### C `primitive_type` typedef query is correct but opaque

**File:** `packages/engram-core/src/ingest/source/queries/c.scm` (lines 19–21)

```scheme
(translation_unit (type_definition
  (primitive_type) @symbol.typedef_declaration .))
```

The `.` anchor (last sibling) correctly selects the alias name node (e.g., `uint32_t` for `typedef unsigned int uint32_t;`), which tree-sitter parses as a `primitive_type` node in this position. This is verified working. The comment above it is helpful but a brief note explaining _why_ capturing a `primitive_type` node yields the alias name (not the base type) would aid future maintainers.

---

### Summary

| Finding | Severity | File |
|---|---|---|
| `visMap` overwrites on multi-modifier C# declarations | **Blocking** | `extractors/c_sharp.ts`, `test/extractor-c-sharp.test.ts` |
| Empty `commit` hashes in MANIFEST.json | Non-blocking | `grammars/MANIFEST.json` |
| `Outer::Inner` Ruby class not asserted in tests | Non-blocking | `test/extractor-ruby.test.ts` |
| `using static` not tested for C# | Non-blocking | `test/extractor-c-sharp.test.ts` |
| `primitive_type` typedef query lacks explanation comment | Non-blocking | `queries/c.scm` |

The blocking bug (C# multi-modifier overwrite) will produce incorrect `exported: false` for all `public static`, `public abstract`, and `public sealed` types — a common pattern in real C# codebases. This must be fixed before merge.

rnwolfe:
Thanks for the thorough review. Here's how I've addressed each finding:

### BLOCKING — Fixed

**C# `visMap` overwrites on multi-modifier declarations** (commit 92f9af6)

Changed the first-pass accumulation from `visMap.set(node.parent.startIndex, node.text)` to accumulate with a template literal:

```typescript
const existing = visMap.get(node.parent.startIndex) ?? "";
visMap.set(node.parent.startIndex, `${existing} ${node.text}`);
```

This means `public static class Foo` now stores `" public static"` in `visMap`, and `hasPublicModifier()` correctly matches the `\bpublic\b` regex.

Added 3 new tests in `extractor-c-sharp.test.ts` covering:
- `exported = true` for `public static class`
- `exported = false` for `internal static class`
- `exported = true` for `public static method`

All 23 tests pass. Build and lint clean.

---

### NON-BLOCKING — Acknowledged

**Empty `commit` hashes in MANIFEST.json** — Noted. These were left empty because the grammars were copied from the node_modules WASM builds and the exact upstream commit wasn't tracked at copy time. I'll follow up in a separate PR to backfill the correct commit SHAs for auditability.

**Ruby `Outer::Inner` not asserted** — Good catch. I'll add `expect(names).toContain("Outer::Inner")` in a follow-up to lock in that behavior.

**`using static` not tested for C#** — Agreed, a test would prevent regression. Will add in follow-up.

**C `primitive_type` typedef query lacks explanation comment** — Will add a clarifying comment in a follow-up PR.

PR #246, opened 2026-04-21
feat(source): add controller-runtime SetupWithManager watches graph extraction

### LINKED ISSUE #239: feat(source): extend Go extractor to capture controller-runtime SetupWithManager watches graph
Part of #236. Depends on #241 (orchestrator generalization).

## Summary

Extend the Go tree-sitter extractor's query to capture the `.For(&T{})`, `.Owns(&U{})`, and `.Watches(&V{})` method-chain calls inside `SetupWithManager` functions from `controller-runtime`, emitting `controller_watches` and `controller_owns` edges from the controller `symbol` entity to a `k8s_resource_kind` entity representing the referenced resource type. This maps the runtime reconciliation topology — "who reacts to what" — without requiring any new grammar.

## Scope

- **In scope**: `.For()`, `.Owns()`, `.Watches()` calls within functions named `SetupWithManager`. Argument must be a composite literal (`&T{}` or `&pkg.T{}`) — bare type name `T` captured, package qualifier `pkg` ignored
- **Out of scope for v0.1**:
  - Raw `controller.NewUnmanaged` / `source.Kind` / `handler.EnqueueRequestsFromMapFunc` patterns outside the builder DSL
  - GVK resolution (resolving `appsv1` import alias to `k8s.io/api/apps/v1` to produce a fully-qualified group/kind target); v0.1 uses bare Go type name as the canonical `k8s_resource_kind` name, accepting that `Deployment` from `appsv1` and `extensionsv1beta1` would collide in the graph. If this becomes a problem in practice, a follow-up issue can add import-alias resolution
  - Builder-pattern variants like `.WithEventFilter()`, `.Named()`, `.Complete()`

- **No new grammar**: Extension of `go.scm` query and `go.ts` extractor only
- **Query strategy**: Match `function_declaration` or `method_declaration` whose `name` is `SetupWithManager`. Within its body, capture `call_expression` nodes whose `function` is a `selector_expression` with selector name `For` / `Owns` / `Watches` — argument captured as the `unary_expression` (`&T{}`) → `composite_literal` → `type`
- **Resource kind naming**: Use the bare Go type name from the argument as the canonical name. Package-qualified types (`appsv1.Deployment`) are normalized to the bare type name (`Deployment`)
- **Output via #241 generalization**: The Go extractor returns watches data through `extraEntities` / `extraEdges`:
  - one `extraEntity` per unique referenced type: `{ canonicalName: "<TypeName>", entityType: K8S_RESOURCE_KIND }`
  - one `extraEdge` per call: `{ source: {kind:"symbol", name:"<EnclosingReceiverTypeName>"}, target: {kind:"canonical", canonicalName:"<TypeName>", entityType:K8S_RESOURCE_KIND}, relationType: CONTROLLER_WATCHES | CONTROLLER_OWNS, edgeKind: "observed", fact: "..." }`
- **Mapping `.For` / `.Owns` / `.Watches` to relation types**: `.For` and `.Watches` both emit `controller_watches`; `.Owns` emits `controller_owns`. `.For` is semantically "primary watched type" but the edge distinction between primary-watch and secondary-watch is not modelled in v0.1
- **Controller symbol resolution**: The `source.kind: "symbol"` reference uses the enclosing receiver type (`func (r *FooReconciler) SetupWithManager(...)` → symbol `FooReconciler`). Top-level functions named `SetupWithManager` (no receiver) are skipped since they have no controller symbol to attach to
- **Vocabulary**: Add `K8S_RESOURCE_KIND: "k8s_resource_kind"` to `entity-types.ts`, and `CONTROLLER_WATCHES: "controller_watches"`, `CONTROLLER_OWNS: "controller_owns"` to `relation-types.ts`
- **Edge kind**: `observed`
- **Storage**: No schema changes

## Features / Scope

| Call pattern | Example | Output |
|---|---|---|
| `.For(&T{})` | `.For(&appsv1.Deployment{})` | `k8s_resource_kind:Deployment` entity + `controller_watches` edge |
| `.Owns(&T{})` | `.Owns(&corev1.ConfigMap{})` | entity + `controller_owns` edge |
| `.Watches(&T{})` | `.Watches(&corev1.Secret{}, handler.EnqueueRequestsForOwner(...))` | entity + `controller_watches` edge (only first arg captured) |
| Non-composite-literal argument | `.For(someVar)` | Skipped — v0.1 only handles inline `&T{}` |
| Top-level `SetupWithManager` (no receiver) | | Skipped — no controller symbol to attach |

## Integration Points

- `packages/engram-core/src/ingest/source/queries/go.scm` — new capture patterns for `SetupWithManager` method bodies and builder call chains
- `packages/engram-core/src/ingest/source/extractors/go.ts` — assembly of watches entries into `extraEntities` / `extraEdges`
- `packages/engram-core/src/ingest/source/extractors/types.ts` — extractor uses types from #241
- `packages/engram-core/src/ingest/source/index.ts` — no changes (orchestrator pass from #241 handles materialization)
- `packages/engram-core/src/vocab/` — one new entity type, two new relation types

## Acceptance Criteria

- [ ] `extractGo()` returns `extraEntities` (one per unique resource type) and `extraEdges` (one per call) for `.For()` / `.Owns()` / `.Watches()` inside `SetupWithManager` methods
- [ ] Calls outside of `SetupWithManager` methods are not captured
- [ ] Top-level `SetupWithManager` functions (no receiver) are skipped
- [ ] Package-qualified types (`appsv1.Deployment`) are normalized to the bare type name (`Deployment`) in the canonical name
- [ ] Non-composite-literal arguments (variable references) are silently skipped
- [ ] `.For` and `.Watches` both produce `controller_watches` edges; `.Owns` produces `controller_owns`
- [ ] `engram ingest source` on a controller-runtime controller file creates `k8s_resource_kind` entities and edges visible in `engram show <symbol>`
- [ ] Idempotent: re-running on unchanged source produces no duplicate entities or edges
- [ ] Unit tests in `packages/engram-core/test/source/go-watches.test.ts` covering: `.For` only, `.For` + `.Owns`, `.For` + `.Watches`, package-qualified type, bare type, top-level function (skipped), variable-ref argument (skipped), call chain across multiple lines
- [ ] `verifyGraph()` passes after ingestion

## Documentation

- [ ] `docs/internal/specs/vocabulary.md` — `k8s_resource_kind` entity type, `controller_watches` and `controller_owns` relation types

rnwolfe:
## Summary

- Extends the Go tree-sitter extractor to detect `SetupWithManager` method bodies and extract `.For()`, `.Owns()`, and `.Watches()` controller-runtime method-chain calls
- Emits `controller_watches` edges (for `.For` and `.Watches`) and `controller_owns` edges (for `.Owns`) from the controller `symbol` entity to `k8s_resource_kind` entities
- Adds `K8S_RESOURCE_KIND` entity type and `CONTROLLER_WATCHES` / `CONTROLLER_OWNS` relation types to the vocab registry

## Acceptance Criteria

- [x] `K8S_RESOURCE_KIND: "k8s_resource_kind"` added to `entity-types.ts`
- [x] `CONTROLLER_WATCHES: "controller_watches"` and `CONTROLLER_OWNS: "controller_owns"` added to `relation-types.ts`
- [x] `go.scm` query captures `setup.receiver`, `setup.name`, `setup.body` for `SetupWithManager` methods
- [x] Extractor processes captures, extracts receiver type, walks method body for For/Owns/Watches calls
- [x] Package-qualified types (`appsv1.Deployment`) normalized to bare type name (`Deployment`)
- [x] Top-level functions (no receiver) skipped
- [x] Non-composite-literal arguments skipped
- [x] `.For` and `.Watches` → `CONTROLLER_WATCHES`; `.Owns` → `CONTROLLER_OWNS`
- [x] Tests cover all specified scenarios: `.For` only, `.For` + `.Owns`, `.For` + `.Watches`, package-qualified type, bare type, top-level function skipped, variable-ref argument skipped, multi-line call chain
- [x] `verifyGraph()` passes after ingestion (integration test)
- [x] Vocabulary doc updated with new entity type and relation types
- [x] All tests pass, build succeeds, lint passes

## Test plan

- [x] Run `bun test packages/engram-core/test/ingest/source/go-watches.test.ts` — 15 tests, all pass
- [x] Run `bun test packages/engram-core/test/ingest/source/` — 215 tests, all pass
- [x] Run `bun run build` — successful
- [x] Run `bun run lint` — no errors

Closes #239

### DIFF
--- a/packages/engram-core/src/ingest/source/extractors/go.ts
+++ b/packages/engram-core/src/ingest/source/extractors/go.ts
@@ -1,5 +1,12 @@
+import type { SyntaxNode } from "web-tree-sitter";
+import { ENTITY_TYPES, RELATION_TYPES } from "../../../vocab/index.js";
 import type { QueryCapture } from "../parser";
-import type { ExtractedFile, ExtractedSymbol } from "./types.js";
+import type {
+  ExtractedEdge,
+  ExtractedEntity,
+  ExtractedFile,
+  ExtractedSymbol,
+} from "./types.js";
 
 export type { ExtractedFile, ExtractedSymbol };
 
@@ -9,22 +16,35 @@ const KIND_MAP: Record<string, ExtractedSymbol["kind"]> = {
   const: "const",
 };
 
+const WATCH_METHODS = new Set(["For", "Owns", "Watches"]);
+
 /**
  * Extract top-level symbols and import paths from tree-sitter-go query captures.
  *
  * In Go, exported symbols have names beginning with an uppercase letter.
  * Import paths are quoted string literals — quotes are stripped.
+ *
+ * Also extracts controller-runtime SetupWithManager watch/owns edges.
  */
 export function extractGo(captures: QueryCapture[]): ExtractedFile {
   const symbols: ExtractedSymbol[] = [];
   const rawImports: string[] = [];
   const seenNames = new Set<string>();
+  const extraEntities: ExtractedEntity[] = [];
+  const extraEdges: ExtractedEdge[] = [];
+  const seenResourceKinds = new Set<string>();
+
+  // Group setup.* captures by position so we can pair receiver + name + body.
+  // Each triplet shares the same method_declaration parent start index.
+  const setupGroups = new Map<
+    number,
+    { receiver?: SyntaxNode; name?: SyntaxNode; body?: SyntaxNode }
+  >();
 
   for (const capture of captures) {
     const { name: captureName, node } = capture;
 
     if (captureName === "import.source") {
-      // Go import paths are interpreted_string_literal — strip surrounding quotes
       const raw = node.text.slice(1, -1);
       rawImports.push(raw);
       continue;
@@ -43,7 +63,6 @@ export function extractGo(captures: QueryCapture[]): ExtractedFile {
         continue;
       }
 
-      // In Go, exported = name starts with uppercase letter
       const exported = /^[A-Z]/.test(symbolName);
 
       seenNames.add(symbolName);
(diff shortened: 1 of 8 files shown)


packages/engram-core/src/ingest/source/extractors/go.ts
```diff
@@ -54,8 +73,144 @@ export function extractGo(captures: QueryCapture[]): ExtractedFile {
...
+      }
+
+      const relationType =
+        selector === "Owns"
+          ? RELATION_TYPES.CONTROLLER_OWNS
+          : RELATION_TYPES.CONTROLLER_WATCHES;
+
+      extraEdges.push({
+        source: { kind: "symbol", name: receiverType },
+        target: {
```
rnwolfe, line 118: **[BLOCKING] Cross-file receiver type causes unhandled throw that aborts entire ingest**

The edge source is emitted as `{ kind: "symbol", name: receiverType }`. In the orchestrator's `resolveEntityRef`, a `"symbol"` ref is resolved by looking up `symbolEntityIds` — which only contains symbols extracted from the **current file**. If `FooReconciler` is defined in `foo.go` but `SetupWithManager` is defined in `bar.go` (a valid and common Go pattern — Go allows method bodies to be in different files than the struct definition as long as they are in the same package), `resolveEntityRef` will throw:

```
Error: EntityRef symbol "FooReconciler" not found among extracted symbols for this file
```

This throw is **not caught** by the orchestrator. The inner `try/catch` on lines 310–332 of `index.ts` only wraps the parse step. The extra-edges loop runs outside it, so the throw propagates to the outer `try/finally`, disposes the parser, and re-throws — **aborting the entire `ingestSource` call** and leaving the DB in a partially-written state (the new episode and file/symbol entities were already committed; the extra edges were not).

**Suggested fix**: Use a `{ kind: "canonical", canonicalName: receiverType, entityType: ENTITY_TYPES.SYMBOL }` ref, or emit a `"canonical"` ref scoped to the file path (e.g. `relPath::FooReconciler`). Alternatively, the orchestrator should catch and log `resolveEntityRef` errors for individual edges rather than letting them abort the file/ingest run. The extractor itself has no way to know the file path, so the cleanest fix is to add a per-edge try/catch in the orchestrator's extra-edges loop.

rnwolfe (commented):
Code review — 4 findings (1 blocking, 3 non-blocking).

packages/engram-core/src/ingest/source/extractors/go.ts
```diff
@@ -54,8 +73,144 @@ export function extractGo(captures: QueryCapture[]): ExtractedFile {
         startByte: node.startIndex,
         endByte: node.endIndex,
       });
+      continue;
+    }
+
+    if (captureName.startsWith("setup.")) {
+      // Group by the method_declaration's start index (grandparent of each capture)
+      const methodNode = node.parent;
```
rnwolfe, line 81: **[non-blocking] Comment says \"grandparent\" but `node.parent` is the direct parent**

The comment reads:
> Group by the method_declaration's start index (grandparent of each capture)

For all three captures in the query (`@setup.receiver` captures `parameter_list`, `@setup.name` captures `field_identifier`, `@setup.body` captures `block`), each captured node is a **direct child** of `method_declaration` in the Go grammar. So `node.parent` is `method_declaration` — the direct parent, not the grandparent. The code is correct; the comment is wrong and could mislead future readers.

Suggested text: `// Group by the method_declaration's start index (direct parent of each capture)`

packages/engram-core/src/ingest/source/extractors/go.ts
```diff
@@ -54,8 +73,144 @@ export function extractGo(captures: QueryCapture[]): ExtractedFile {
...
+          }
+        }
+      }
+    }
+  }
+
+  for (let i = 0; i < node.childCount; i++) {
+    const child = node.child(i);
+    if (child) results.push(...collectWatchCalls(child));
+  }
```
rnwolfe, line 184: **[non-blocking] Spread-based recursion is O(n²) in tree depth**

`results.push(...collectWatchCalls(child))` creates a new intermediate array for every child at every recursion level. For a method body with N AST nodes and depth D, this results in O(N × D) array allocations and copies. For typical controller-runtime `SetupWithManager` bodies this is trivially fast, but for pathologically large method bodies (e.g. a giant switch statement) it could degrade.

Consider accumulating into a shared array passed by reference, or using a generator/iterator pattern:

```ts
function* collectWatchCalls(node: SyntaxNode): Iterable<{selector: string; resourceKind: string}> {
  // ... check + yield ...
  for (let i = 0; i < node.childCount; i++) {
    const child = node.child(i);
    if (child) yield* collectWatchCalls(child);
  }
}
```

This avoids all intermediate array allocations. Not blocking given the bounded scope of `SetupWithManager` bodies, but worth noting for long-term maintainability.

packages/engram-core/test/ingest/source/go-watches.test.ts
```diff
@@ -0,0 +1,334 @@
...
+  });
+});
+
+// ---------------------------------------------------------------------------
+// Integration test — ingestSource + verifyGraph
+// ---------------------------------------------------------------------------
+
+describe("ingestSource — Go controller watches integration", () => {
+  let tmpDir: string;
+  let graphPath: string;
```
rnwolfe, line 276: **[non-blocking] No test for cross-file receiver type (struct in a different file)**

All tests define the struct (`type FooReconciler struct{}`) and the `SetupWithManager` method in the **same source string/file**. The blocking issue noted in `go.ts` — where `{ kind: \"symbol\", name: receiverType }` throws if the struct is defined in a different file — is not exercised here.

A test that writes the struct to `controllers/types.go` and the method to `controllers/foo_controller.go` and then calls `ingestSource` would expose the crash path. Even if the fix is \"catch and skip with a warning\", a regression test documenting that behavior would be valuable.

packages/engram-core/src/ingest/source/extractors/go.ts
rnwolfe: Fixed in 7b8582c. resolveEntityRef now returns null (with a warn log) instead of throwing when a symbol ref isn't found in the current file. The edge loop skips null-resolved refs. Added a test for the cross-file struct+method scenario.

packages/engram-core/src/ingest/source/extractors/go.ts
rnwolfe: Fixed in 7b8582c — changed 'grandparent' to 'parent' in the comment.

packages/engram-core/src/ingest/source/extractors/go.ts
rnwolfe: Acknowledged. The O(n²) behavior is a concern only for pathologically deep ASTs — method bodies are shallow in practice. Deferring optimization; can follow up if profiling shows it matters.

packages/engram-core/test/ingest/source/go-watches.test.ts
rnwolfe: Fixed in 7b8582c — added 'does not throw when SetupWithManager receiver type is defined in a different file' test covering the crash path.

PR #255, opened 2026-04-22
feat(ingest): google workspace scope grammar - folder and query discovery modes

### LINKED ISSUE #198: feat(ingest): Google Workspace scope grammar — folder: and query: discovery modes
## Summary

Extend `GoogleWorkspaceAdapter` (from #196) with discovery-based scope modes: `folder:<id>`
enumerates Docs within a Drive folder or shared drive, and `query:<drive-q>` runs an arbitrary
Drive search query. Both expand to a list of doc IDs that flow through the same revision-aware
ingest path landed in #196 — this is a **scope grammar extension**, not a new adapter.

Enables workflows like "ingest every Doc I own" (`query:'me' in owners`), "ingest our team's
docs folder" (`folder:1abc...`), and "ingest recently viewed Docs"
(`query:viewedByMeTime > '2026-01-01'`).

## Features / Scope

| Feature | Description |
|---------|-------------|
| `scope: folder:<folderId>` | Enumerate Docs in a Drive folder; ingest each |
| `scope: folder:<id>?recursive=true` | Recursively enumerate subfolders |
| `scope: folder:<sharedDriveId>` | Works with shared-drive IDs (same endpoint, `supportsAllDrives=true`) |
| `scope: query:<drive-q>` | Raw Drive search query; adapter AND-injects `mimeType='application/vnd.google-apps.document'` so only Docs are enumerated |

**Module location.** No new adapter — extends
`packages/engram-core/src/ingest/adapters/google-workspace.ts`. Drive discovery code split into
`google-workspace-discovery.ts`.

**Scope parsing.** Extend `scopeSchema.validate()`:
- `folder:<id>` and `folder:<id>?recursive=true|false` — validate folder ID shape.
  (Shared-drive IDs are structurally indistinguishable; Drive API handles both.)
- `query:<drive-q>` — accept any non-empty string after the prefix. Malformed Drive queries
  surface at call time as `400` → `data_error`.
- Unknown query-string keys (e.g. `folder:<id>?foo=bar`) → validation error listing valid keys.

**Drive API.**
- `GET https://www.googleapis.com/drive/v3/files?q=<query>&fields=files(id,modifiedTime),nextPageToken&pageSize=100&supportsAllDrives=true&includeItemsFromAllDrives=true`
- `folder:<id>` → query `mimeType='application/vnd.google-apps.document' and '<id>' in parents and trashed=false`.
- `folder:<id>?recursive=true` → BFS over subfolders
  (`mimeType='application/vnd.google-apps.folder' and '<id>' in parents`).
  Cycle guard required: shared drives can form cycles via shortcuts.
- `query:<drive-q>` → adapter AND-injects
  `mimeType='application/vnd.google-apps.document' and trashed=false and (<user-q>)`.

**Cursor semantics.** Discovery scopes need a scan cursor. Use `ingestion_runs.cursor` to
store the max `modifiedTime` seen in the Drive enumeration; on resume, Drive query is
augmented with `modifiedTime > <cursor>`. Deduplication still falls through to the
per-episode revision check from #196 — an enumerated-but-unchanged Doc is a silent skip.

**Recursion flag parsing.** `?key=value` syntax keeps the grammar extensible to future modes
(e.g. `?since=2026-01-01`). Parse with URL-like semantics; unknown keys = hard error.

**Rate limiting / pagination.** Drive list API is 1000 req/100 s/user. Paginate with
`nextPageToken`; call `opts.onProgress` on each page. Fan-out to the Docs API reuses #196's
429 handling.

**Performance guardrail.** A large folder may hold thousands of Docs.
- `--dry-run` with discovery scopes runs enumeration only (count preview, no content fetch).
  Lets users sanity-check scope before committing to long-running ingest.
- `onProgress` emits `phase: "enumerating"` during discovery, then `phase: "ingesting"`
  during content fetch.

## Integration Points

- **Hard dependency on #196.** Cannot start until #196 is merged — extends the same adapter
  class, scope schema, ingest helpers, and auth plumbing.
- **No new vocabulary.** Reuses `google_workspace` / `google_doc` / `document` from #196.
- **CLI**: the `engram ingest google-workspace` command grows help text and examples for
  folder/query scopes. No new subcommand.
- **Progress UI**: CLI shows "enumerating folder..." then "ingesting N docs..." with counts.

## Acceptance Criteria

**Scope grammar**
- [ ] `scopeSchema.validate()` accepts `folder:<id>`, `folder:<id>?recursive=true`,
      `folder:<id>?recursive=false`, and `query:<non-empty>`.
- [ ] Rejects unknown query-string keys (`folder:<id>?foo=bar`) with a message listing valid keys.
- [ ] Rejects empty query (`query:`) and unknown prefixes.

**Folder discovery**
- [ ] `folder:<id>` enumerates all non-trashed Docs whose parent is `<id>`; each flows through
      #196's ingest path.
- [ ] `folder:<id>?recursive=true` enumerates Docs in `<id>` and all subfolders; cycle guard
      prevents infinite loops on shared-drive shortcuts.
- [ ] Shared drives work: passing a shared drive ID enumerates top-level Docs.
- [ ] Empty folder → `episodesCreated: 0`, no error.

**Query discovery**
- [ ] `query:'me' in owners` enumerates user-owned Docs.
- [ ] `query:viewedByMeTime > '2026-01-01'` executes correctly.
- [ ] AND-injection wins: a user query of
      `query:mimeType='application/vnd.google-apps.spreadsheet'` returns zero results.
- [ ] Malformed Drive query → `EnrichmentAdapterError { code: 'data_error' }` carrying the
      400 message from Drive.

**Incremental re-ingest**
- [ ] Second run with same scope uses `ingestion_runs.cursor` to filter Drive enumeration
      to changed Docs only.
- [ ] New Docs appearing in scope since last run are picked up.
- [ ] Docs removed from scope (deleted, moved out, ownership change) do not trigger
      supersession — they just stop being re-enumerated. Existing episodes stay with their history.

**Error paths**
- [ ] 404 on folder ID → `data_error` with clear message.
- [ ] Missing `drive.readonly` scope → `auth_failure` with remediation hint.
- [ ] Partial fan-out failure (one doc 404s, rest succeed) → logged, counted, ingest continues.

**Dry run**
- [ ] `--dry-run` with `folder:` or `query:` scope enumerates + reports count; does NOT fetch
      doc content.

**Tests**
- [ ] Unit tests for scope parsing including recursion-flag parsing.
- [ ] Unit tests for folder enumeration with injected fetch: flat, nested, shared drive, empty,
      cycle guard.
- [ ] Unit test: query scope AND-injection of `mimeType` constraint.
- [ ] Unit test: incremental cursor with advanced `modifiedTime` skips unchanged docs.

## Documentation

- [ ] User-facing: extend the Google Workspace ingest docs (added in #196) with folder and
      query examples.
- [ ] Internal: extend the Google Workspace section in `docs/internal/specs/adapter-contract.md`
      with a scope-grammar table.
- [ ] CLAUDE.md: reference the `prefix:id[?opts]` scope-grammar pattern as the convention for
      future multi-mode adapters.

## Related

- Builds on #196 — cannot start until #196 is merged.
- Epic #111.

rnwolfe:
## Summary

- Extends `GoogleWorkspaceAdapter` with Drive-based doc discovery: `folder:<id>`, `folder:<id>?recursive=true`, and `query:<drive-q>` scopes
- Adds `google-workspace-discovery.ts` with `enumerateFolderDocs()` (flat + BFS recursive with cycle guard), `enumerateQueryDocs()` (mimeType AND-injection), `parseFolderScope()`, and `computeDiscoveryCursor()`
- Cursor semantics: stores `max(modifiedTime)` after each run; subsequent runs augment the Drive query with `modifiedTime > '<cursor>'` for incremental ingest
- Dry-run for discovery scopes runs Drive list calls but skips Docs content fetch, returning `episodesCreated: 0`

## Acceptance Criteria

- [x] `folder:<id>` enumerates all Docs in a flat folder via Drive files.list
- [x] `folder:<id>?recursive=true` traverses subfolders via BFS with cycle guard
- [x] `folder:<id>?foo=bar` rejected (unknown query param, error lists valid keys)
- [x] `query:<drive-q>` AND-injects `mimeType` and `trashed=false` constraints
- [x] `query:` (empty) rejected with `data_error`
- [x] 400 from Drive (malformed query) → `EnrichmentAdapterError { code: 'data_error' }`
- [x] Individual doc 404 in fan-out → logged, skipped, continues
- [x] Cursor stored as `max(modifiedTime)` across enumerated docs; next run augments query
- [x] Dry-run enumerates docs but skips Docs API calls
- [x] `verifyGraph()` passes after folder and query ingest
- [x] Existing `doc:` and `docs:` scopes unchanged
- [x] 39 unit + integration tests (injected fetch, real SQLite `:memory:`)
- [x] Build passes, lint clean, all tests pass

## Test plan

- Run `bun test packages/engram-core/test/ingest/google-workspace-discovery.test.ts` — 39 tests pass
- Run `bun test packages/engram-core/test/ingest/google-workspace.test.ts` — 32 existing tests pass
- Run `bun run build && bun run lint` — clean

Closes #198

### DIFF
--- a/packages/engram-core/src/ingest/adapters/google-workspace-discovery.ts
+++ b/packages/engram-core/src/ingest/adapters/google-workspace-discovery.ts
@@ -1,0 +1,347 @@
+/**
+ * google-workspace-discovery.ts — Drive discovery helpers for folder and query scopes.
+ *
+ * Provides:
+ * - `enumerateFolderDocs()` — list all Google Docs in a Drive folder (flat or recursive BFS)
+ * - `enumerateQueryDocs()` — execute a Drive search query, AND-injecting mimeType constraint
+ *
+ * These functions are called from the GoogleWorkspaceAdapter when scope is
+ * `folder:<id>` or `query:<q>`. They return arrays of doc-id / modifiedTime pairs
+ * that the adapter then fans out to the existing per-doc ingest logic.
+ *
+ * Cursor semantics:
+ *   Callers pass a `since` ISO8601 cursor. Discovery functions append
+ *   `modifiedTime > '<cursor>'` to the Drive query so only changed docs are returned.
+ *   After a run, the adapter stores `max(modifiedTime)` seen across all docs as the
+ *   new cursor.
+ */
+
+import { EnrichmentAdapterError } from "../adapter.js";
+import type { GWFetchFn } from "./google-workspace-helpers.js";
+import { apiGetGW } from "./google-workspace-helpers.js";
+
+// ---------------------------------------------------------------------------
+// Drive API types (only fields we use)
+// ---------------------------------------------------------------------------
+
+export interface DriveFileItem {
+  id: string;
+  modifiedTime?: string;
+}
+
+interface DriveFilesListResponse {
+  files?: DriveFileItem[];
+  nextPageToken?: string;
+}
+
+// ---------------------------------------------------------------------------
+// Parsed folder scope
+// ---------------------------------------------------------------------------
+
+export interface ParsedFolderScope {
+  folderId: string;
+  recursive: boolean;
+}
+
+/**
+ * Parse a `folder:<id>` or `folder:<id>?recursive=true` scope string.
+ * Throws on unrecognised query params.
+ */
+export function parseFolderScope(scope: string): ParsedFolderScope {
+  // Strip the "folder:" prefix
+  const rest = scope.slice("folder:".length);
+
+  const qIdx = rest.indexOf("?");
+  if (qIdx === -1) {
+    const folderId = rest.trim();
+    if (!folderId) {
+      throw new Error("folder scope must include a folder ID: folder:<id>");
+    }
+    return { folderId, recursive: false };
+  }
+
+  const folderId = rest.slice(0, qIdx).trim();
+  if (!folderId) {
+    throw new Error("folder scope must include a folder ID: folder:<id>");
+  }
+
+  const queryString = rest.slice(qIdx + 1);
+  const params = new URLSearchParams(queryString);
+
+  // Validate: only 'recursive' is a recognised key
+  for (const key of params.keys()) {
+    if (key !== "recursive") {
+      throw new Error(
+        `Unknown query parameter in folder scope: '${key}'. Valid keys: recursive`,
+      );
+    }
+  }
+
+  const recursiveVal = params.get("recursive");
+  let recursive = false;
+  if (recursiveVal !== null) {
+    if (recursiveVal !== "true" && recursiveVal !== "false") {
+      throw new Error(
+        `Invalid value for recursive in folder scope: '${recursiveVal}'. Use true or false`,
+      );
+    }
+    recursive = recursiveVal === "true";
+  }
+
+  return { folderId, recursive };
+}
+
+// ---------------------------------------------------------------------------
+// Drive list helper (one page)
+// ---------------------------------------------------------------------------
+
+/**
+ * Fetch one page of Drive files matching `query`.
+ * Handles 400 (malformed query) → data_error, 404 (folder not found) → data_error.
+ */
+async function driveListPage(
+  fetchFn: GWFetchFn,
+  token: string,
+  query: string,
+  pageToken: string | undefined,
+  fields: string,
+): Promise<DriveFilesListResponse> {
+  const params = new URLSearchParams({
+    q: query,
+    fields,
+    pageSize: "100",
+    supportsAllDrives: "true",
+    includeItemsFromAllDrives: "true",
+  });
+  if (pageToken) {
+    params.set("pageToken", pageToken);
+  }
+
+  const url = `https://www.googleapis.com/drive/v3/files?${params.toString()}`;
+
+  try {
+    return await apiGetGW<DriveFilesListResponse>(fetchFn, url, token);
+  } catch (err) {
+    if (err instanceof EnrichmentAdapterError) {
+      // apiGetGW maps unexpected statuses to data_error with the status code in the message
+      if (err.code === "data_error" && err.message.includes("404")) {
+        throw new EnrichmentAdapterError(
+          "data_error",
+          `folder not found: check the folder ID and Drive permissions`,
+        );
+      }
+      if (err.code === "data_error" && err.message.includes("400")) {
+        throw new EnrichmentAdapterError(
+          "data_error",
+          `Drive query error (400): ${err.message}`,
+        );
+      }
+    }
+    throw err;
+  }
+}
+
+// ---------------------------------------------------------------------------
+// Enumerate docs in a single folder (one level only)
+// ---------------------------------------------------------------------------
+
+async function enumerateDocsInFolder(
+  fetchFn: GWFetchFn,
+  token: string,
+  folderId: string,
+  since: string | null,
+): Promise<DriveFileItem[]> {
+  let q =
+    `mimeType='application/vnd.google-apps.document'` +
+    ` and '${folderId}' in parents` +
+    ` and trashed=false`;
+
+  if (since) {
+    q += ` and modifiedTime > '${since}'`;
+  }
+
+  const docs: DriveFileItem[] = [];
+  let pageToken: string | undefined;
+
+  do {
+    const resp = await driveListPage(
+      fetchFn,
+      token,
+      q,
+      pageToken,
+      "files(id,modifiedTime),nextPageToken",
+    );
+    for (const f of resp.files ?? []) {
+      if (f.id) docs.push(f);
+    }
+    pageToken = resp.nextPageToken;
+  } while (pageToken);
+
+  return docs;
+}
+
+// ---------------------------------------------------------------------------
+// Enumerate subfolders inside a folder (BFS, cycle-guarded)
+// ---------------------------------------------------------------------------
+
+async function enumerateSubfolders(
+  fetchFn: GWFetchFn,
+  token: string,
+  folderId: string,
+): Promise<string[]> {
+  const q =
+    `mimeType='application/vnd.google-apps.folder'` +
+    ` and '${folderId}' in parents` +
+    ` and trashed=false`;
+
+  const ids: string[] = [];
+  let pageToken: string | undefined;
+
+  do {
+    const resp = await driveListPage(
+      fetchFn,
+      token,
+      q,
+      pageToken,
+      "files(id),nextPageToken",
+    );
+    for (const f of resp.files ?? []) {
+      if (f.id) ids.push(f.id);
+    }
+    pageToken = resp.nextPageToken;
+  } while (pageToken);
+
+  return ids;
+}
+
+// ---------------------------------------------------------------------------
+// Public: enumerateFolderDocs
+// ---------------------------------------------------------------------------
+
+/**
+ * Enumerate all Google Docs inside `folderId`, optionally recursing into
+ * subfolders (BFS with cycle guard).
+ *
+ * @param fetchFn  - Injected fetch implementation (real or test stub).
+ * @param token    - Bearer token for Drive API.
+ * @param folderId - Drive folder ID to enumerate.
+ * @param recursive - Whether to recurse into subfolders.
+ * @param since    - ISO8601 cursor. If provided, only docs modified after
+ *                   this timestamp are returned.
+ * @returns Array of `{ id, modifiedTime }` tuples for matched docs.
+ */
+export async function enumerateFolderDocs(
+  fetchFn: GWFetchFn,
+  token: string,
+  folderId: string,
+  recursive: boolean,
+  since: string | null,
+): Promise<DriveFileItem[]> {
+  const allDocs: DriveFileItem[] = [];
+  const visitedFolders = new Set<string>();
+
+  // BFS queue — starts with the root folder
+  const queue: string[] = [folderId];
+
+  while (queue.length > 0) {
+    const currentId = queue.shift() as string;
+
+    if (visitedFolders.has(currentId)) {
+      // Cycle guard: skip already-visited folders (handles Drive shortcuts)
+      continue;
+    }
+    visitedFolders.add(currentId);
+
+    const docs = await enumerateDocsInFolder(fetchFn, token, currentId, since);
+    allDocs.push(...docs);
+
+    if (recursive) {
+      const subfolders = await enumerateSubfolders(fetchFn, token, currentId);
+      for (const subfolderId of subfolders) {
+        if (!visitedFolders.has(subfolderId)) {
+          queue.push(subfolderId);
+        }
+      }
+    }
+  }
+
+  return allDocs;
+}
+
+// ---------------------------------------------------------------------------
+// Public: enumerateQueryDocs
+// ---------------------------------------------------------------------------
+
+/**
+ * Execute a Drive search query, always AND-injecting `mimeType` and `trashed`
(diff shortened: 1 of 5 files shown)


packages/engram-core/src/ingest/adapters/google-workspace-discovery.ts
```diff
@@ -0,0 +1,347 @@
...
+
+async function enumerateDocsInFolder(
+  fetchFn: GWFetchFn,
+  token: string,
+  folderId: string,
+  since: string | null,
+): Promise<DriveFileItem[]> {
+  let q =
+    `mimeType='application/vnd.google-apps.document'` +
+    ` and '${folderId}' in parents` +
```
rnwolfe, line 166: **[BLOCKING] folderId injection into Drive query — potential query manipulation**

In `enumerateDocsInFolder` (line 155-157 of the new file):
```ts
let q =
  `mimeType='application/vnd.google-apps.document'` +
  ` and '${folderId}' in parents` +
  ` and trashed=false`;
```

The `folderId` is interpolated directly into the Drive query string without any sanitisation. A folder ID containing a single-quote can break the query structure, e.g. `folder:abc'and trashed=false or '1'='1`. `parseFolderScope` validates the query param keys and the `recursive` value but imposes **no character-set restriction on `folderId` itself** — it only checks `!folderId` (empty string).

Drive folder IDs from Google are alphanumeric (base62-ish) in practice, but the contract does not enforce that. The same pattern appears in `enumerateSubfolders`.

**Fix**: add a regex guard in `parseFolderScope` before returning:
```ts
if (!/^[A-Za-z0-9_-]+$/.test(folderId)) {
  throw new Error(`Invalid folder ID: '${folderId}'. Must be alphanumeric (A-Za-z0-9_-).`);
}
```
This keeps the public scope grammar aligned with Google's actual ID format and eliminates the injection surface.

rnwolfe (commented):
Code review by forge-loop sub-agent.

**Summary**: 3 blocking findings, 3 non-blocking findings.

The implementation is well-structured and covers all acceptance criteria from #198. The major concerns are a cursor regression in dry-run mode, folderId injection into Drive queries (security), and the `computeDiscoveryCursor` using lexicographic string comparison for ISO8601 timestamps (technically correct but brittle). The test suite is comprehensive and correctly uses `verifyGraph()` and injected fetch.

packages/engram-core/src/ingest/adapters/google-workspace.ts
```diff
@@ -210,23 +223,38 @@ export class GoogleWorkspaceAdapter implements EnrichmentAdapter {
...
+            refreshFn,
+            counts,
+            opts.dryRun,
+          );
+        }
 
-      if (!opts.dryRun && run) {
-        completeIngestionRun(graph, runId, null, {
-          episodes: counts.episodesCreated,
-          entities: counts.entitiesCreated,
```
rnwolfe, line 227: **[BLOCKING] Cursor not read in dry-run mode — second dry-run always re-enumerates everything**

In `enrichDiscovery` (the new method, line ~292 of the file):
```ts
const since = readIsoCursor(graph, INGESTION_SOURCE, scope);
```

The cursor is read from the **last completed run**. But when `dryRun: true`, `createIngestionRun` is skipped entirely (`run = null`, `runId = ""`), and `completeIngestionRun` / `writeCursor` are never called. That is correct for dry-run (no writes). However, the cursor lookup queries `ingestion_runs … WHERE status = 'completed'` — which means a dry-run following a dry-run will also correctly fall back to the last real run's cursor. This part is fine.

The actual regression: after the dry-run path returns early (line ~324 in the method), the call site in `enrich()` falls through to the outer `catch/return` with no cursor write — **that is correct**. But `enrichDiscovery` is called with `runId = ""` when `dryRun`. Later, inside `enrichDiscovery`, after the fan-out, it calls:
```ts
if (runId) {
  completeIngestionRun(graph, runId, newCursor, ...);
}
```
The `if (runId)` guard is the right intent, but the early `return` for dry-run happens **before** this point, so this path is unreachable in dry-run — no write occurs. This is actually correct.

**However**: the `since` cursor is read **before** the dry-run guard. When dry-run with an empty `runId` and then the full fan-out is skipped, the enumerated `discoveredDocs` still contains items filtered by `since`. If the intent is that `--dry-run` should show the full count of what *would* be enumerated on a fresh run (not cursor-filtered), then the cursor should not be applied during dry-run. Currently the dry-run count reflects *cursor-filtered* docs, not the total. The acceptance criterion says "enumerates + reports count" — ambiguous, but the cursor-filtered count is misleading to users doing a first sanity-check.

**Recommendation**: either document clearly that dry-run respects the cursor (and update help text), or pass `since = null` when `dryRun === true`.

packages/engram-core/src/ingest/adapters/google-workspace-discovery.ts
```diff
@@ -0,0 +1,347 @@
...
+// ---------------------------------------------------------------------------
+// Cursor: compute max modifiedTime from a set of discovered docs
+// ---------------------------------------------------------------------------
+
+/**
+ * Given a list of enumerated Drive file items, return the maximum
+ * `modifiedTime` seen as an ISO8601 string — to be stored as the run cursor.
+ * Returns `null` if the list is empty or no item has a `modifiedTime`.
+ */
+export function computeDiscoveryCursor(items: DriveFileItem[]): string | null {
```
rnwolfe, line 347: **[BLOCKING] `computeDiscoveryCursor` uses lexicographic string comparison — will silently produce wrong cursor if timestamps use different ISO8601 formats**

```ts
if (max === null || item.modifiedTime > max) {
  max = item.modifiedTime;
}
```

Lexicographic comparison works correctly **only** if all Drive `modifiedTime` strings are in the same format (e.g. `2024-06-01T12:00:00.000Z`). The Drive API is consistent in practice (RFC3339 with milliseconds and Z suffix), so this will work today. However:

1. It is undocumented that this assumes a fixed format — a future reader might refactor without realising the ordering dependency.
2. The Drive API spec does not guarantee millisecond precision in all responses; a timestamp without millis sorts differently from one with millis of the same second.

**Recommendation**: parse with `new Date(a).getTime() > new Date(b).getTime()` for correctness and clarity, and add a comment explaining why the max is used as the cursor. Alternatively, add an `// invariant: Drive modifiedTime is always RFC3339 UTC with ms precision` comment to make the assumption explicit.

packages/engram-core/src/ingest/adapters/google-workspace-discovery.ts
```diff
@@ -0,0 +1,347 @@
...
+  if (pageToken) {
+    params.set("pageToken", pageToken);
+  }
+
+  const url = `https://www.googleapis.com/drive/v3/files?${params.toString()}`;
+
+  try {
+    return await apiGetGW<DriveFilesListResponse>(fetchFn, url, token);
+  } catch (err) {
+    if (err instanceof EnrichmentAdapterError) {
```
rnwolfe, line 135: **[NON-BLOCKING] 404 error message loses the original folder ID**

```ts
throw new EnrichmentAdapterError(
  "data_error",
  `folder not found: check the folder ID and Drive permissions`,
);
```

The error message tells the user to "check the folder ID" but does not say which folder ID was not found. In recursive mode with many subfolders this makes it hard to diagnose which folder caused the 404 (a subfolder can also 404 if it is a shortcut to a deleted location).

**Recommendation**: include the folder ID in the message:
```ts
`folder not found (${folderId}): check the folder ID and Drive permissions`
```
Note: `driveListPage` does not currently receive `folderId` as a parameter, so you would need to pass it through or re-throw with context at the `enumerateDocsInFolder` / `enumerateSubfolders` call sites.

packages/engram-core/src/ingest/adapters/google-workspace.ts
```diff
@@ -175,8 +190,6 @@ export class GoogleWorkspaceAdapter implements EnrichmentAdapter {
 
     googleWorkspaceScopeSchema.validate(scope);
 
-    const docIds = parseScope(scope);
-
     // Resolve token from auth credential
     const auth = opts.auth;
```
rnwolfe, line 194: **[NON-BLOCKING] `supportsCursor = false` is stale after this PR**

```ts
supportsCursor = false;
```

The adapter now reads and writes an ISO cursor for `folder:` and `query:` scopes via `readIsoCursor` / `writeCursor`. The `supportsCursor` field advertised on the adapter interface should be `true` (or at minimum conditionally true based on scope type). Callers or tooling that inspect this flag to decide whether to offer resume UI will show incorrect information.

**Recommendation**: set `supportsCursor = true` and update the JSDoc if any exists on the `EnrichmentAdapter` interface for this field.

packages/engram-core/test/ingest/google-workspace-discovery.test.ts
```diff
@@ -0,0 +1,831 @@
...
+        status: 200,
+        headers: { "Content-Type": "application/json" },
+      });
+    };
+
+    await enumerateQueryDocs(
+      fetchFn as typeof fetch,
+      TOKEN,
+      "name contains 'spec'",
+      null,
```
rnwolfe, line 473: **[NON-BLOCKING] Cursor integration test (second run) does not assert the cursor value stored in `ingestion_runs`**

The test at line ~640 (`second run uses cursor`) captures the Drive query string and verifies it contains `modifiedTime >` and the expected cursor value — good. But it does not verify the cursor that was *written* to `ingestion_runs` after run 1 is correct. If `computeDiscoveryCursor` returns `null` (e.g. because `modifiedTime` fields are absent from the mock), `readIsoCursor` on run 2 returns `null` and the cursor filter would never appear in the query — the test would still pass.

The `driveFiles` mock for run 1 does include `modifiedTime`, so in practice this works. But a direct assertion like:
```ts
const storedCursor = readIsoCursor(graph, INGESTION_SOURCE_TYPES.GOOGLE_WORKSPACE, `folder:${FOLDER}`);
expect(storedCursor).toBe("2024-06-01T00:00:00Z");
```
between the two adapter runs would make the test immune to mock changes silently breaking the cursor chain. Note: you would need to import `readIsoCursor` and `INGESTION_SOURCE_TYPES` in the test file.

rnwolfe:
All review findings addressed in commit b119015.

**Fix 1 (BLOCKING — folder ID injection):** Added a regex guard `!/^[A-Za-z0-9_-]+$/` in both code paths of `parseFolderScope` (plain ID and ID-with-query-string). Any folder ID containing a single quote or other special character now throws immediately with a clear error message before it can reach the Drive query string.

**Fix 2 (BLOCKING — dry-run cursor skipping):** In `enrichDiscovery`, the `since` variable is now set to `null` when `dryRun === true`, so `--dry-run` enumerates all in-scope docs rather than only those past the stored cursor. This gives an accurate "what would be ingested" count for first-time sanity checks.

**Fix 3 (BLOCKING — lexicographic timestamp comparison):** `computeDiscoveryCursor` now uses `new Date(ts).getTime()` for comparison instead of raw string `>`. This is robust against differing RFC3339 precision (whole-second vs millisecond suffixes).

**Fix 4 (non-blocking — `supportsCursor`):** Corrected `supportsCursor = false` → `true` on `GoogleWorkspaceAdapter`.

**Fix 5 (non-blocking — cursor assertion in test):** Added `readIsoCursor` and `INGESTION_SOURCE_TYPES` imports to the test file. The cursor integration test now explicitly asserts the stored cursor equals `"2024-06-01T00:00:00Z"` (the max modifiedTime from the first run) before proceeding to the second run.

All 41 discovery tests pass. Build is clean.

PR #256, opened 2026-04-22
refactor(ingest): move google-workspace adapter to packages/plugins/

rnwolfe:
## Summary

- Moves `GoogleWorkspaceAdapter` from `engram-core/src/ingest/adapters/` to
  `packages/plugins/google-workspace/`, following the in-repo plugin model
  established by the gerrit port
- Adds `@engram/plugin-google-workspace` workspace package with `package.json`,
  `manifest.json`, and migrated source (`index.ts`, `helpers.ts`, `discovery.ts`)
- Removes `GoogleWorkspaceAdapter` and `googleWorkspaceScopeSchema` exports from
  `engram-core/src/index.ts`
- Updates `engram-cli` to import `GoogleWorkspaceAdapter` from
  `@engram/plugin-google-workspace` instead of `engram-core`
- Moves all Google Workspace tests to `packages/plugins/google-workspace/test/`
  with imports updated to use `engram-core` and relative plugin paths

Addresses ADR-006 decision in DECISIONS.md.

All 1379 tests pass, build succeeds, lint clean.

## Test plan

- [x] `bun test` — 1379 tests pass (0 failures), including the migrated
  Google Workspace adapter and discovery tests in the new plugin package
- [x] `bun run lint` — exit 0

### DIFF
--- a/packages/engram-cli/package.json
+++ b/packages/engram-cli/package.json
@@ -13,6 +13,7 @@
   "dependencies": {
     "@clack/prompts": "^0.9.0",
     "@engram/engram-web": "workspace:*",
+    "@engram/plugin-google-workspace": "workspace:*",
     "commander": "^13.0.0",
     "engram-core": "workspace:*",
     "google-auth-library": "^9.0.0",
--- a/packages/engram-cli/src/commands/ingest.ts
+++ b/packages/engram-cli/src/commands/ingest.ts
@@ -11,7 +11,7 @@
 import * as fs from "node:fs";
 import * as path from "node:path";
 import { intro, log, outro, spinner } from "@clack/prompts";
-
+import { GoogleWorkspaceAdapter } from "@engram/plugin-google-workspace";
 // git and markdown ingest use execFileSync / synchronous SQLite — the event
 // loop is blocked for the duration, so spinner intervals cannot fire.
 // For those commands we log a "starting" line and print results when done.
@@ -26,7 +26,6 @@ import {
   closeGraph,
   EnrichmentAdapterError,
   GitHubAdapter,
-  GoogleWorkspaceAdapter,
   ingestGitRepo,
   ingestMarkdown,
   ingestSource,
--- a/packages/engram-core/src/index.ts
+++ b/packages/engram-core/src/index.ts
@@ -156,10 +156,6 @@ export {
   GitHubAuthError,
   githubScopeSchema,
 } from "./ingest/adapters/github.js";
-export {
-  GoogleWorkspaceAdapter,
-  googleWorkspaceScopeSchema,
-} from "./ingest/adapters/google-workspace.js";
 export type {
   PluginReferencePattern,
   ReferencePattern,
--- a/packages/plugins/google-workspace/manifest.json
+++ b/packages/plugins/google-workspace/manifest.json
@@ -1,0 +1,15 @@
+{
+  "name": "google-workspace",
+  "version": "0.1.0",
+  "contract_version": 1,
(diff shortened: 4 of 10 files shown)


rnwolfe (commented):
Code review by forge-loop sub-agent

## Summary

This is a clean refactor with no behaviour changes. All core logic is intact, imports are correctly migrated, and the plugin contract is correctly implemented. **2 low-severity observations** — no blocking issues.

### Migration correctness

- All relative `../../` imports replaced with `engram-core` — zero missed paths across `src/index.ts`, `src/helpers.ts`, `src/discovery.ts`
- Test files updated from deep relative `../../src/ingest/adapters/...` paths to `engram-core` and local `../src/` paths
- `engram-core/src/index.ts` cleanly drops the four-line `GoogleWorkspaceAdapter`/`googleWorkspaceScopeSchema` export block

### No new behaviour

Pure port — logic in all three source files is byte-for-byte identical to the original except for the import rewrites and the inline dynamic import cleaned up to a named import (`DriveFileItem` from `./discovery.js`).

### Plugin contract

- `manifest.json` `supported_auth: ["oauth2", "bearer"]` matches `supportedAuth` on the class ✓
- `manifest.json` `supports_cursor: true` matches `supportsCursor = true` on the class ✓
- `src/index.ts` exports `default new GoogleWorkspaceAdapter()` as required by js-module contract ✓
- Scope schema description and regex pattern in `manifest.json` match the four scope prefixes handled by `validateScope()` ✓

### Dependency hygiene

- `engram-core/package.json` has no `google-auth-library` — correct ✓
- `google-auth-library` present in both plugin `package.json` (for type resolution) and CLI `package.json` (for ADC) — correct ✓
- `ulid` declared in plugin `package.json` — correct (no longer ambient from `engram-core`) ✓

### Tests

- Both test files import from `engram-core` and `../src/` — no stale paths ✓
- `GoogleWorkspaceAdapter` imported from `../src/index.js` in tests — exercises plugin code, not core ✓

### No leftover references

- Zero hits for `google-workspace` or `GoogleWorkspace` in `packages/engram-core/src/` ✓
- Only surviving reference is `INGESTION_SOURCE_TYPES.GOOGLE_WORKSPACE` in `packages/engram-core/src/vocab/source-types.ts` — that is the vocab registry and is correct to remain there ✓

---

## Observations (non-blocking)

### [Low] Dual adapter instances — default export vs CLI direct instantiation

**File:** `packages/plugins/google-workspace/src/index.ts` (bottom of file)

The CLI imports `GoogleWorkspaceAdapter` as a named class and calls `new GoogleWorkspaceAdapter()` at `ingest.ts:657`. The `default` export at the bottom of `src/index.ts` also creates an instance, but it is never used by the CLI. Two adapter instances exist when the CLI runs; neither path uses the other.

This is harmless today (`this.fetchFn` defaults to `fetch`, class is otherwise stateless), but if the constructor ever gains required parameters the two code paths will diverge silently. Consider either:
- Having the CLI import and use the default export (`import adapter from '@engram/plugin-google-workspace'`), or
- Documenting clearly that the CLI always constructs its own instance independent of the js-module contract default.

### [Low] No `exports` map in `package.json`

**File:** `packages/plugins/google-workspace/package.json`

Only `"main": "src/index.ts"` is declared. The gerrit plugin has the same pattern, so this is consistent with existing in-repo convention. Bun workspace resolution works correctly today. When a build step is introduced, add an `exports` map:
```json
"exports": { ".": { "bun": "./src/index.ts", "default": "./dist/index.js" } }
```
No action required for this PR.

PR #257, opened 2026-04-22
feat(plugins): plugin docs contract — manifest description/docs fields and plugin info command

rnwolfe:
## Summary

- Adds optional `description` and `docs` (summary, auth_setup, scope_examples) fields to `PluginManifest` type and `loadManifest()` in `engram-core`
- Surfaces `description` as a new column in `plugin list` output
- Adds `engram plugin info <name>` subcommand that renders a formatted capabilities + docs card
- Prints `docs.auth_setup` hint after successful `plugin install`
- Adds `README.md` to gerrit and google-workspace bundled plugins
- Documents the contract in `docs/internal/specs/plugin-docs.md`

## Test plan

- [ ] `bun run build` passes (all packages)
- [ ] `bun test` passes (1379 tests, 0 failures)
- [ ] `bun run lint` passes (no errors)
- [ ] `engram plugin list` shows DESCRIPTION column when plugins have `description` in their manifest
- [ ] `engram plugin info gerrit` renders capabilities block + overview + auth setup + examples
- [ ] `engram plugin info google-workspace` renders correctly with all docs sections
- [ ] `engram plugin install <name>` shows auth_setup hint after success for plugins with docs.auth_setup
- [ ] `engram plugin info <nonexistent>` exits with error and helpful message

### DIFF
--- a/packages/engram-cli/src/commands/plugin.ts
+++ b/packages/engram-cli/src/commands/plugin.ts
@@ -67,6 +67,28 @@ function copyDir(src: string, dest: string): void {
   }
 }
 
+/**
+ * Word-wrap a string to a given column width. Returns an array of lines.
+ */
+function wrapText(text: string, width: number): string[] {
+  const words = text.split(/\s+/);
+  const lines: string[] = [];
+  let current = "";
+  for (const word of words) {
+    if (!word) continue;
+    if (!current) {
+      current = word;
+    } else if (current.length + 1 + word.length <= width) {
+      current += ` ${word}`;
+    } else {
+      lines.push(current);
+      current = word;
+    }
+  }
+  if (current) lines.push(current);
+  return lines;
+}
+
 // ---------------------------------------------------------------------------
 // plugin list
 // ---------------------------------------------------------------------------
@@ -134,6 +156,7 @@ function registerList(plugin: Command): void {
         scope: string;
         source: string;
         status: string;
+        description: string;
       }> = [];
 
       for (const pd of discovered) {
@@ -146,6 +169,7 @@ function registerList(plugin: Command): void {
             scope: pd.scope,
             source: pd.source,
             status: "OK",
+            description: manifest.description ?? "",
           });
         } catch (err) {
           rows.push({
@@ -158,6 +182,7 @@ function registerList(plugin: Command): void {
               err instanceof ManifestValidationError
                 ? `FAILED: ${err.message}`
                 : `FAILED: ${err instanceof Error ? err.message : String(err)}`,
+            description: "",
           });
         }
       }
@@ -170,6 +195,7 @@ function registerList(plugin: Command): void {
         "SCOPE",
         "SOURCE",
         "STATUS",
+        "DESCRIPTION",
       ];
       const colWidths = headers.map((h, i) => {
         const key = [
@@ -179,6 +205,7 @@ function registerList(plugin: Command): void {
           "scope",
           "source",
           "status",
+          "description",
         ][i] as keyof (typeof rows)[0];
         return Math.max(h.length, ...rows.map((r) => r[key].length));
       });
@@ -200,6 +227,7 @@ function registerList(plugin: Command): void {
             r.scope,
             r.source,
             r.status,
+            r.description,
           ]),
         ),
       ];
@@ -286,6 +314,19 @@ function registerInstall(plugin: Command): void {
       log.success(
         `Installed plugin '${name}' → ${dest} (${method === "symlink" ? "symlinked" : "copied"})`,
       );
+
+      // Show auth_setup hint if present in the manifest
+      try {
+        const manifest = loadManifest(src);
+        if (manifest.docs?.auth_setup) {
+          const wrapped = wrapText(manifest.docs.auth_setup, 78);
+          log.info(
+            `Before first use:\n${wrapped.map((l) => `    ${l}`).join("\n")}`,
+          );
+        }
+      } catch {
+        // Non-fatal: manifest already validated above, ignore re-read errors
+      }
     });
 }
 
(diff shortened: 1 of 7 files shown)


packages/engram-core/src/plugins/manifest.ts
```diff
@@ -183,5 +196,13 @@ export function loadManifest(pluginDir: string): PluginManifest {
...
         ? (obj.vocab_extensions as PluginVocabExtensions)
         : undefined,
+    description:
+      typeof obj.description === "string" ? obj.description : undefined,
+    docs:
+      typeof obj.docs === "object" &&
+      obj.docs !== null &&
+      !Array.isArray(obj.docs)
+        ? (obj.docs as PluginDocs)
+        : undefined,
```
rnwolfe: [Medium] Unsound cast: `obj.docs` inner fields are not validated.

The outer type check (`object`, not `null`, not `Array`) is correct, but `(obj.docs as PluginDocs)` is unsound — the inner fields (`scope_examples`, `summary`, `auth_setup`) are completely unvalidated. If a manifest ships `"scope_examples": "bad-string"`, `plugin info` crashes with `TypeError: manifest.docs.scope_examples.map is not a function` when it tries to compute `maxScope`.

At minimum, `scope_examples` should be validated as either `undefined` or an `Array`. Example fix:

```ts
docs: (() => {
  if (typeof obj.docs !== 'object' || obj.docs === null || Array.isArray(obj.docs)) return undefined;
  const d = obj.docs as Record<string, unknown>;
  return {
    summary: typeof d.summary === 'string' ? d.summary : undefined,
    auth_setup: typeof d.auth_setup === 'string' ? d.auth_setup : undefined,
    scope_examples: Array.isArray(d.scope_examples)
      ? (d.scope_examples as PluginDocs['scope_examples'])
      : undefined,
  };
})()
```

packages/engram-cli/src/commands/plugin.ts
```diff
@@ -67,6 +67,28 @@ function copyDir(src: string, dest: string): void {
   }
 }
 
+/**
+ * Word-wrap a string to a given column width. Returns an array of lines.
+ */
+function wrapText(text: string, width: number): string[] {
+  const words = text.split(/\s+/);
```
rnwolfe, line 77: [Low] `wrapText` silently overflows `width` when a single word exceeds the limit.

A word longer than `width` is placed on its own line without splitting, so the output can exceed `width`. The google-workspace `auth_setup` URL token (`https://www.googleapis.com/auth/documents.readonly,...`) is ~100 chars and will overflow the 78-char limit. Consider either force-breaking long tokens or documenting this as a known limitation in the JSDoc.

packages/engram-cli/src/commands/plugin.ts
```diff
@@ -286,6 +314,19 @@ function registerInstall(plugin: Command): void {
       log.success(
         `Installed plugin '${name}' → ${dest} (${method === "symlink" ? "symlinked" : "copied"})`,
       );
+
+      // Show auth_setup hint if present in the manifest
+      try {
+        const manifest = loadManifest(src);
+        if (manifest.docs?.auth_setup) {
+          const wrapped = wrapText(manifest.docs.auth_setup, 78);
+          log.info(
```
rnwolfe, line 326: [Low] Spec says wrap at 80 chars; code uses 78.

`plugin-docs.md` section 3 states "word-wrapped at 80 chars" for the install hint. The code calls `wrapText(manifest.docs.auth_setup, 78)`. Adding the 4-space indent prefix brings effective column width to 82 chars — which may be intentional — but conflicts with the stated spec. Either update the spec to clarify "78 chars plus 4-space indent" or use 76 so that `4 + 76 = 80`.

packages/engram-cli/src/commands/plugin.ts
```diff
@@ -385,6 +426,100 @@ function registerUninstall(plugin: Command): void {
...
+    .command("info <name>")
+    .description("Show detailed information about a discovered plugin")
+    .option("--project <path>", "check project-local plugins too")
+    .option("--bundled-root <path>", "override bundled plugins root (testing)")
+    .action(async (name: string, opts: PluginInfoOpts) => {
+      const projectRoot = path.resolve(opts.project ?? ".");
+      const root = opts.bundledRoot ?? bundledPluginsRoot();
+
+      const discovered = discoverPlugins(projectRoot, root ?? undefined);
+      const pd = discovered.find((p) => p.name === name);
```
rnwolfe: [Low] `plugin info <name>` cannot find uninstalled bundled plugins.

`discoverPlugins()` only enumerates installed plugins (XDG data home and `.engram/plugins/`). Running `engram plugin info gerrit` before `engram plugin install gerrit` fails with "not found in any plugin directory" even though gerrit is bundled. `plugin list --available` correctly uses `listBundledPlugins(root)` for this case. `plugin info` should do the same fallback: if `discoverPlugins` returns no match but `listBundledPlugins(root).includes(name)`, call `loadManifest(path.join(root, name))` directly from the bundled source path.

rnwolfe (commented):
Code review by forge-loop sub-agent. Found 4 issues (1 medium, 3 low) described in inline comments.

packages/engram-core/src/plugins/manifest.ts
rnwolfe: Fixed in 06bd6b8. Added validation in `loadManifest()` that checks `docs.scope_examples` is an array if present, throwing `ManifestValidationError` with a descriptive message if not. The `docs` assignment now uses an IIFE to encapsulate the check before casting to `PluginDocs`.

packages/engram-cli/src/commands/plugin.ts
rnwolfe: Fixed in 06bd6b8. Added a JSDoc comment to `wrapText` documenting that long tokens (e.g. URLs) exceeding `width` are placed on their own line without splitting — hyphenation is not implemented. This is a documentation fix only; the existing behaviour is already the best we can do without hyphenation.

packages/engram-cli/src/commands/plugin.ts
rnwolfe: Fixed in 06bd6b8. Updated `plugin-docs.md` §3 to say "word-wrapped at 78 chars" to match the code (`wrapText(text, 78)`). The code was already correct; the spec was the one that needed updating. Also updated the §1 description of manifest loader behaviour to reflect the new `scope_examples` array validation.

packages/engram-cli/src/commands/plugin.ts
rnwolfe: Fixed in 06bd6b8. After `discoverPlugins()` returns no match, `plugin info` now falls back to `listBundledPlugins(root)`. If the name is found there, it constructs a synthetic `PluginDirectory` pointing at `path.join(root, name)` with `source: 'bundled'`. Search order is now: project-local > user (XDG) > bundled. The error message was also updated to suggest `--available` for discoverability.

PR #265, opened 2026-04-23
feat(ingest): monorepo-aware source exclusions — vendor heuristics + .engramignore

### LINKED ISSUE #258: feat(ingest): monorepo-aware source exclusions — vendor heuristics + .engramignore
## Summary

Large monorepos routinely commit third-party code that has no business in the knowledge
graph: vendor trees, codegen output, test fixtures, copied proto libraries. The current
walker only excludes paths listed in `.gitignore` and a short hardcoded denylist
(`node_modules`, `dist`, etc.). Committed vendor directories like `vendor/`,
`third_party/`, or `proto/gen/` pass right through, bloating the entity graph and
poisoning retrieval. This issue adds two complementary layers of exclusion targeted at
monorepo use: (1) an expanded denylist that covers the most common vendor/generated
directory names with zero config, and (2) a `.engramignore` file that lets individual
repos express repo-specific exclusions using familiar glob syntax.

- **Module location**: `packages/engram-core/src/ingest/source/walker.ts`
- **Storage**: No schema changes. `.engramignore` is read at walk-time from the
  filesystem, not stored in the database.
- **Expanded `DENY_DIRS`**: Add `vendor`, `third_party`, `_vendor`, `extern`,
  `generated`, `gen`, `pb`, `proto_gen`, `testdata` to the hardcoded set. These
  names are unambiguous in virtually every codebase and safe to exclude by default.
  No new config or CLI flags needed.
- **`.engramignore` parsing**: Use the existing `ignore` library (already a dep) to
  parse `.engramignore` at the repo root. Follow the same chain-traversal pattern
  already in place for `.gitignore` — check each directory level for a `.engramignore`
  file, build an `Ignore` instance, and apply it to relative paths. The file uses
  standard gitignore glob syntax so developers need no new documentation.
- **Precedence**: denylist → `.engramignore` → `.gitignore` → user `--exclude` flags.
  A path excluded by denylist cannot be re-included by `.engramignore` (denylist is
  a hard floor). A path excluded by `.gitignore` can be re-included by a negation
  pattern in `.engramignore` (`!vendor/my-patch/`).
- **`respectEngramignore` option**: Add to `WalkOptions` alongside the existing
  `respectGitignore` flag. Default `true`. Allows callers to opt out (useful in tests).
- **Security**: `.engramignore` paths are resolved relative to the repo root and
  normalized before use — no path traversal possible since the `ignore` library
  operates purely on relative strings.
- **Performance**: `.engramignore` files are cached per directory in the same
  `gitignoreCache` map pattern already used for `.gitignore`. No additional I/O per
  file.

## Integration Points

- `walker.ts` is the only file that changes for core logic. The public `WalkOptions`
  interface gains one optional field (`respectEngramignore?: boolean`).
- `ingestSource()` in `packages/engram-core/src/ingest/source/index.ts` passes
  `WalkOptions` through — no changes needed there unless we want to surface a
  `--no-engramignore` flag on the CLI.
- Existing walker tests in `packages/engram-core/test/` need new fixture cases.
- No changes to the database schema, CLI surface, or plugin system.

## Acceptance Criteria

- [ ] `DENY_DIRS` in `walker.ts` includes `vendor`, `third_party`, `_vendor`, `extern`,
  `generated`, `gen`, `pb`, `proto_gen`, `testdata`
- [ ] A directory named `vendor/` at repo root is never yielded, even with
  `respectGitignore: false`
- [ ] A directory named `testdata/` nested inside `src/` is never yielded
- [ ] Walker checks for `.engramignore` at each directory level, using the same
  chain-traversal logic as `.gitignore`
- [ ] A pattern in root `.engramignore` (e.g. `proto/gen/`) excludes all files under
  that path
- [ ] A subdirectory `.engramignore` excludes paths relative to that subdirectory
- [ ] Negation patterns in `.engramignore` (`!proto/gen/custom.ts`) re-include files
  excluded by a broader pattern in the same file
- [ ] `respectEngramignore: false` in `WalkOptions` bypasses `.engramignore` but keeps
  the denylist (mirrors existing `respectGitignore` behavior)
- [ ] `.engramignore` is NOT itself yielded as a `FileEntry` (add to `DENY_FILE_PATTERNS`)
- [ ] Unit tests cover: new denylist entries, `.engramignore` at root, subdirectory
  `.engramignore`, negation pattern, `respectEngramignore: false` opt-out
- [ ] `bun test`, `bun run lint`, `bun run build` all pass

## Documentation

- [ ] Add `.engramignore` usage note to `engram ingest source --help` text
- [ ] CLAUDE.md: update walker description under File Organization to mention `.engramignore`

rnwolfe:
## Summary

- Expands `DENY_DIRS` in the file walker with 9 additional vendor/generated directory names commonly found in monorepos: `vendor`, `third_party`, `_vendor`, `extern`, `generated`, `gen`, `pb`, `proto_gen`, `testdata`
- Adds `.engramignore` file support: gitignore glob syntax, checked at each directory level in the same chain-traversal pattern as `.gitignore`
- New `WalkOptions.respectEngramignore` field (default `true`) and `--no-engramignore` CLI flag
- `.engramignore` is never itself yielded as a `FileEntry`
- Denylist is a hard floor — it cannot be overridden by `.engramignore`

## Acceptance Criteria

- [x] `DENY_DIRS` includes: `vendor`, `third_party`, `_vendor`, `extern`, `generated`, `gen`, `pb`, `proto_gen`, `testdata`
- [x] A directory named `vendor/` at repo root is never yielded, even with `respectGitignore: false`
- [x] A directory named `testdata/` nested inside `src/` is never yielded
- [x] Walker checks for `.engramignore` at each directory level (same chain-traversal as `.gitignore`)
- [x] A root `.engramignore` pattern (e.g. `proto/output/`) excludes all files under that path
- [x] A subdirectory `.engramignore` excludes paths relative to that subdirectory
- [x] Negation patterns in `.engramignore` (e.g. `*.pb.ts` + `!custom.ts`) re-include files excluded by broader pattern in same file (follows standard gitignore semantics: negation works for file globs, not for directories)
- [x] `respectEngramignore: false` bypasses `.engramignore` but keeps the denylist
- [x] `.engramignore` is NOT yielded as a `FileEntry`
- [x] Unit tests cover: new denylist entries (all 9), `.engramignore` at root, subdirectory `.engramignore`, negation pattern, `respectEngramignore: false` opt-out, independence from `.gitignore`, chain traversal
- [x] `bun test`, `bun run lint`, `bun run build` all pass

## Files Changed

- `packages/engram-core/src/ingest/source/walker.ts` — expanded `DENY_DIRS`, `.engramignore` loader/cache/check, `WalkOptions.respectEngramignore`
- `packages/engram-core/src/ingest/source/index.ts` — `SourceIngestOptions.respectEngramignore`, pass-through to `walk()`
- `packages/engram-core/test/ingest/source/walker.test.ts` — new test suites for vendor heuristics and `.engramignore`
- `packages/engram-cli/src/commands/ingest.ts` — `--no-engramignore` flag, updated help text with exclusion precedence documentation
- `CLAUDE.md` — updated walker description to mention `.engramignore`

Closes #258

### DIFF
--- a/packages/engram-cli/src/commands/ingest.ts
+++ b/packages/engram-cli/src/commands/ingest.ts
@@ -242,14 +242,16 @@ Examples:
 
 Exclusion precedence (highest to lowest):
   1. Built-in denylist dirs (node_modules, vendor, generated, testdata, …)
-  2. .engramignore  — per-directory ignore file (gitignore glob syntax).
+  2. --exclude flags
+  3. .engramignore  — per-directory ignore file (gitignore glob syntax).
      Place a .engramignore at the repo root or any subdirectory.
-     Supports negation patterns to re-include paths:
+     Supports negation patterns to re-include paths excluded by .engramignore:
        proto/gen/        # exclude everything under proto/gen/
        !proto/gen/custom.ts  # …except this one file
+     Note: negation only works within .engramignore itself — it cannot
+     re-include a path that .gitignore has excluded (they are independent gates).
      Use --no-engramignore to bypass .engramignore (denylist still applies).
-  3. .gitignore      — use --no-gitignore to bypass
-  4. --exclude flags
+  4. .gitignore      — use --no-gitignore to bypass
 
 When to use:
   Run after engram ingest git to add symbol-level entities (functions,
--- a/packages/engram-cli/src/commands/ingest.ts
+++ b/packages/engram-cli/src/commands/ingest.ts
@@ -54,6 +54,7 @@ interface IngestMdOpts {
 interface IngestSourceOpts {
   exclude?: string[];
   gitignore: boolean;
+  engramignore: boolean;
   dryRun?: boolean;
   verbose?: boolean;
   db: string;
@@ -239,6 +240,17 @@ Examples:
   # Verbose per-file output
   engram ingest source --verbose
 
+Exclusion precedence (highest to lowest):
+  1. Built-in denylist dirs (node_modules, vendor, generated, testdata, …)
+  2. .engramignore  — per-directory ignore file (gitignore glob syntax).
+     Place a .engramignore at the repo root or any subdirectory.
+     Supports negation patterns to re-include paths:
+       proto/gen/        # exclude everything under proto/gen/
+       !proto/gen/custom.ts  # …except this one file
+     Use --no-engramignore to bypass .engramignore (denylist still applies).
+  3. .gitignore      — use --no-gitignore to bypass
+  4. --exclude flags
+
 When to use:
   Run after engram ingest git to add symbol-level entities (functions,
   classes, modules) for code navigation queries.
@@ -255,6 +267,10 @@ See also:
     .option(
       "--no-gitignore",
       "skip .gitignore application (denylist still applies)",
+    )
+    .option(
+      "--no-engramignore",
+      "skip .engramignore application (denylist still applies)",
     )
     .option("--dry-run", "walk and report counts without writing")
     .option("-v, --verbose", "emit per-file progress output")
@@ -318,6 +334,7 @@ See also:
           root: resolvedSource,
           exclude: opts.exclude?.length ? opts.exclude : undefined,
           respectGitignore: opts.gitignore,
+          respectEngramignore: opts.engramignore,
           dryRun: opts.dryRun,
           onProgress,
         });
--- a/packages/engram-core/src/ingest/source/walker.ts
+++ b/packages/engram-core/src/ingest/source/walker.ts
@@ -16,6 +16,7 @@ export interface WalkOptions {
   root: string;
   exclude?: string[];
   respectGitignore?: boolean; // default true
+  respectEngramignore?: boolean; // default true
   maxFileBytes?: number; // default 1_048_576
 }
 
@@ -27,6 +28,18 @@ const DENY_DIRS = new Set([
   "out",
   "target",
   "coverage",
+  // vendor / third-party
+  "vendor",
+  "third_party",
+  "_vendor",
+  "extern",
+  // generated / proto output
+  "generated",
+  "gen",
+  "pb",
+  "proto_gen",
+  // test fixtures
+  "testdata",
 ]);
 
 /** File glob patterns that are always excluded */
@@ -49,6 +62,7 @@ const DENY_FILE_PATTERNS = [
   /^\.gitignore$/,
   /^\.gitattributes$/,
   /^\.gitmodules$/,
+  /^\.engramignore$/,
 ];
 
 const DEFAULT_MAX_FILE_BYTES = 1_048_576; // 1MB
@@ -89,6 +103,21 @@ function loadGitignore(dirPath: string): Ignore | null {
 }
 
 /**
+ * Load .engramignore file and return an `ignore` instance, or null if not found.
+ */
+function loadEngramignore(dirPath: string): Ignore | null {
+  const engramignorePath = path.join(dirPath, ".engramignore");
+  try {
+    const content = fs.readFileSync(engramignorePath, "utf8");
+    const ig = ignore();
+    ig.add(content);
+    return ig;
+  } catch {
+    return null;
+  }
+}
+
+/**
  * Walk a directory tree, yielding FileEntry objects for each ingestable file.
  */
 export async function* walk(opts: WalkOptions): AsyncIterable<FileEntry> {
@@ -96,6 +125,7 @@ export async function* walk(opts: WalkOptions): AsyncIterable<FileEntry> {
     root,
     exclude = [],
     respectGitignore = true,
+    respectEngramignore = true,
     maxFileBytes = DEFAULT_MAX_FILE_BYTES,
   } = opts;
 
@@ -112,6 +142,8 @@ export async function* walk(opts: WalkOptions): AsyncIterable<FileEntry> {
 
   // gitignore instances per directory path (keyed by absPath)
   const gitignoreCache = new Map<string, Ignore | null>();
+  // engramignore instances per directory path (keyed by absPath)
+  const engramignoreCache = new Map<string, Ignore | null>();
 
   function getGitignore(dirPath: string): Ignore | null {
     const cached = gitignoreCache.get(dirPath);
@@ -120,6 +152,16 @@ export async function* walk(opts: WalkOptions): AsyncIterable<FileEntry> {
     }
     const ig = respectGitignore ? loadGitignore(dirPath) : null;
     gitignoreCache.set(dirPath, ig);
+    return ig;
+  }
+
+  function getEngramignore(dirPath: string): Ignore | null {
+    const cached = engramignoreCache.get(dirPath);
+    if (cached !== undefined) {
+      return cached;
+    }
+    const ig = respectEngramignore ? loadEngramignore(dirPath) : null;
+    engramignoreCache.set(dirPath, ig);
     return ig;
   }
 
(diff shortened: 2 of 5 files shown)


packages/engram-core/src/ingest/source/walker.ts
```diff
@@ -149,6 +191,42 @@ export async function* walk(opts: WalkOptions): AsyncIterable<FileEntry> {
...
   }
 
+  /**
+   * Check if a path (relative to root, posix) is ignored by any .engramignore
+   * in the chain from root to the file's parent directory.
+   *
+   * Precedence: denylist (hard floor) → .engramignore → .gitignore → user excludes
+   *
+   * A negation in .engramignore can re-include a path excluded by .gitignore
+   * because .engramignore is checked independently. The caller applies both
```
rnwolfe: **[BLOCKING] False claim: `.engramignore` negation cannot re-include `.gitignore`-excluded paths**

The docblock states:

> A negation in `.engramignore` can re-include a path excluded by `.gitignore` because `.engramignore` is checked independently.

This is incorrect. The two filters are independent sequential gates, not a single merged ignore chain:

```ts
// walker.ts — file branch
if (respectEngramignore && isEngramignored(relPosix)) continue;  // gate 1
if (respectGitignore    && isGitignored(relPosix))    continue;  // gate 2
```

For a file excluded by `.gitignore`:
- `isEngramignored()` returns `false` (no `.engramignore` pattern matches it positively, so negation has nothing to cancel)
- The file then hits gate 2, `isGitignored()` returns `true`, and it is skipped

Adding `!path/to/file` to `.engramignore` does nothing for a file excluded only by `.gitignore`. The `ignore` library negation operates within a single ruleset — it can negate a prior positive pattern in the same file, but it cannot negate a match in a different filter.

**What negation actually does here:** if `.engramignore` contains `proto/output/*.pb.ts` AND `!proto/output/custom.ts`, then `custom.ts` is not caught by gate 1 and (if not gitignored) passes through. That is the only working cross-pattern negation — entirely within `.engramignore` itself.

The same false claim appears as an inline comment at the walk loop call site (positions 149-150 in the walker.ts diff hunk).

**Fix:** Remove the cross-layer negation claim from the docblock. The accurate statement is: negation patterns within `.engramignore` follow standard gitignore semantics — they can re-include files matched by an earlier positive pattern in the same `.engramignore` file. They cannot override `.gitignore` exclusions.

packages/engram-cli/src/commands/ingest.ts
```diff
@@ -239,6 +240,17 @@ Examples:
...
+  1. Built-in denylist dirs (node_modules, vendor, generated, testdata, …)
+  2. .engramignore  — per-directory ignore file (gitignore glob syntax).
+     Place a .engramignore at the repo root or any subdirectory.
+     Supports negation patterns to re-include paths:
+       proto/gen/        # exclude everything under proto/gen/
+       !proto/gen/custom.ts  # …except this one file
+     Use --no-engramignore to bypass .engramignore (denylist still applies).
+  3. .gitignore      — use --no-gitignore to bypass
+  4. --exclude flags
+
```
rnwolfe, line 255: **[BLOCKING] Help text precedence order does not match the actual code**

The help text documents:
```
1. Built-in denylist dirs
2. .engramignore
3. .gitignore
4. --exclude flags  <-- listed last
```

But in `walker.ts` the actual check order in the file branch is:
```ts
if (isDeniedFile(entryName)) continue;                       // 1. denylist
if (exclude.length > 0 && rootUserIg.ignores(relPosix)) continue;  // 2. --exclude (SECOND, not last)
if (respectEngramignore && isEngramignored(relPosix)) continue;     // 3. .engramignore
if (respectGitignore && isGitignored(relPosix)) continue;           // 4. .gitignore
```

`--exclude` patterns are applied before `.engramignore` and `.gitignore`, not after. This means a file excluded by `--exclude` is never evaluated by `.engramignore` — fine semantically, but the documented order is wrong and will confuse users trying to reason about flag interactions.

**Fix:** Update the help text to:
```
1. Built-in denylist dirs (node_modules, vendor, generated, testdata, ...)
2. --exclude flags
3. .engramignore  -- per-directory ignore file (gitignore glob syntax).
   ...
4. .gitignore     -- use --no-gitignore to bypass
```

rnwolfe (commented):
## Review: feat(ingest): monorepo-aware source exclusions — vendor heuristics + .engramignore

Overall the implementation is solid. DENY_DIRS expansion, caching, self-exclusion of `.engramignore`, `respectEngramignore: false` semantics, and the test suite all check out. Two blocking issues found — one incorrect claim in docs/comments that misrepresents what the code actually does, and one precedence order bug in the help text. Two minor non-blocking observations follow.

### Findings by severity

**Blocking (2):** Incorrect cross-layer negation claim; incorrect help text precedence order
**Non-blocking (2):** Triple-redundant flag check; no directory-level .engramignore pruning

rnwolfe (commented):
Non-blocking observations (style and documentation).

packages/engram-core/src/ingest/source/walker.ts
```diff
@@ -199,6 +277,13 @@ export async function* walk(opts: WalkOptions): AsyncIterable<FileEntry> {
           continue;
         }
 
+        // Apply .engramignore chain
+        // Checked before .gitignore so that a negation in .engramignore can
+        // re-include a path that .gitignore would otherwise exclude.
+        if (respectEngramignore && isEngramignored(relPosix)) {
+          continue;
```
rnwolfe, line 283: **[NON-BLOCKING] Triple-redundant `respectEngramignore` guard**

The flag is checked in three places for the same code path:

1. Walk loop call site: `if (respectEngramignore && isEngramignored(relPosix))`
2. Inside `isEngramignored`: `if (!respectEngramignore) return false;`
3. Inside `getEngramignore`: `const ig = respectEngramignore ? loadEngramignore(dirPath) : null;`

Compare `isGitignored`, which only has the early-return inside the function plus the outer guard — two checks. The extra layer in `getEngramignore` is the odd one out; the early-return in `isEngramignored` already makes it unreachable.

No correctness impact, but the inconsistency with the gitignore pattern is worth cleaning up for symmetry. Removing either the early-return in `isEngramignored` or the conditional in `getEngramignore` would match the existing style.

packages/engram-core/src/ingest/source/walker.ts
```diff
@@ -149,6 +191,42 @@ export async function* walk(opts: WalkOptions): AsyncIterable<FileEntry> {
     return false;
   }
 
+  /**
+   * Check if a path (relative to root, posix) is ignored by any .engramignore
```
rnwolfe, line 195: **[NON-BLOCKING] Directories are not pruned by `.engramignore` — only files are filtered**

The `.engramignore` check runs in the `stat.isFile()` branch only. If a user adds `large-generated-dir/` to `.engramignore`, the walker still descends into that directory and checks each file individually against the `.engramignore` chain.

This is consistent with the existing `.gitignore` behaviour in this walker (same limitation), so it is not a regression introduced by this PR. However, it is a known performance caveat worth documenting, particularly since the `.engramignore` feature is specifically aimed at monorepos with large vendor/generated trees — the very trees where directory-level pruning matters most.

Suggested addition to the `isEngramignored` docblock:

```
 * Note: this check runs at the file level only. Directories are not pruned early
 * by .engramignore (unlike DENY_DIRS). A pattern like "large-dir/" will correctly
 * exclude all files in that directory but will not prevent the walker from
 * descending into it.
```

rnwolfe:
Fixed both blocking issues: corrected the cross-layer negation claim in the docblock (negation only works within .engramignore itself, not across .gitignore), and updated the help text precedence order to match actual code execution (denylist → --exclude → .engramignore → .gitignore).

PR #266, opened 2026-04-23
feat(cli): engram sync — config-driven multi-source orchestration

### LINKED ISSUE #203: feat(cli): engram sync — config-driven multi-source orchestration
## Summary

A single `engram sync` command that runs all configured ingesters in dependency order,
driven by a per-project `.engram.config.json` file. Replaces the need to remember
`ingest git` + `ingest source` + `ingest enrich github` + `ingest enrich gerrit` + ...
as adapters multiply. The config file is the persistent handle for endpoint URLs,
scope values, and auth env var names that today live in shell history or per-developer
aliases.

## Problem

With GitHub (built-in), Gerrit (plugin, #191), Google Workspace (plugin, #196), and
GitLab/Jira/Linear/Buganizer (#193–195, #192) adapters in or arriving, the
one-command-per-source pattern is unusable. Dependency order (git entities → source
code → enrichments → cross-ref resolver #197) is implicit and fragile when run by
hand. And cross-ref resolution only pays off when all sources for a project have been
ingested before it runs — which is hard to guarantee without orchestration.

## Features / Scope

| Feature | Description |
|---------|-------------|
| `engram sync` | Runs all sources in config declaration order, then cross-refs |
| `--config <path>` | Override default config discovery |
| `--db <path>` | Target graph (defaults to `.engram` resolution order) |
| `--only <name>[,<name>...]` | Run a subset of sources by `name` |
| `--continue-on-error` | Run remaining sources after a failure (default: fail-fast) |
| `--no-cross-refs` | Skip the cross-ref resolver step |
| `--format json` | Structured output for CI; default is human-readable |
| `--dry-run` | Validate config + print plan without executing |

## Config schema (`.engram.config.json`)

```json
{
  "version": 1,
  "sources": [
    { "name": "repo-git",     "type": "git",       "path": "." },
    { "name": "repo-source",  "type": "source",    "root": "packages/" },
    { "name": "engram-gh",    "type": "github",    "scope": "rnwolfe/engram",
                              "auth": { "kind": "bearer", "tokenEnv": "GITHUB_TOKEN" } },
    { "name": "chromium-cl",  "type": "gerrit",    "scope": "chromium/src",
                              "endpoint": "https://chromium-review.googlesource.com" },
    { "name": "design-docs",  "type": "google_workspace", "scope": "docId1,docId2",
                              "auth": { "kind": "oauth2", "tokenEnv": "GOOGLE_TOKEN" } }
  ]
}
```

Schema rules:
- `version` is required and must be `1` (bump on breaking schema changes).
- Every source requires `name` (unique within the file), `type`, and whatever scope
  shape the adapter declares via its `scopeSchema.validate()`.
- `auth` block uses the existing `AuthCredential` union shape from the v2 adapter
  contract. Sources that declare `supportedAuth: ['none']` (`git`, `source`) may omit
  `auth`.
- `name` is required even for single-source-of-a-type configs — it is the stable
  handle for `--only`, log messages, and per-source exit status.
- Unknown fields at the top level or per-source level cause validation failure
  (fail-closed — do not silently drop).

Config discovery order (first hit wins):
1. `--config <path>` flag
2. `$ENGRAM_CONFIG` env var
3. `.engram.config.json` in the current working directory
4. `.engram.config.json` alongside the resolved `.engram` db file

Absence is an error: `engram sync` with no discoverable config exits non-zero with a
message pointing at `engram init` (future) or the example in `docs/`.

- **Split — core owns orchestration, CLI owns discovery/parsing:**
  - `packages/engram-core/src/sync/` — new package-internal module
    - `run.ts` → `runSync(graph, config, opts): SyncResult` (the orchestrator)
    - `types.ts` → `SyncConfig`, `SyncSource`, `SyncResult`, `SourceResult`
    - `errors.ts` → `SyncConfigValidationError`, `SyncSourceError`
    - `index.ts` → barrel; re-exported from `packages/engram-core/src/index.ts`
  - `packages/engram-cli/src/commands/sync.ts` — new command
    - Config discovery (flag → env → cwd → db-adjacent)
    - JSON parse + hand-rolled schema validation (mirrors
      `packages/engram-core/src/plugins/manifest.ts` pattern — no new dep)
    - Env-var resolution for `auth.tokenEnv`
    - Calls `runSync()`, formats output
- **No new dependencies.** Validation is hand-rolled with a `SyncConfigValidationError`
  class that includes a field path (e.g. `sources[2].auth.tokenEnv`) and a reason. This
  matches how `ManifestValidationError` is structured in the plugins module.
- **Source resolution is plugin-aware.** For each source, `runSync` resolves the
  adapter by `type` in this order: built-in adapters (`GitHubAdapter`, `ingestGitRepo`,
  `ingestSource`) → discovered plugins (via `discoverPlugins()` from
  `packages/engram-core/src/plugins/`). Unknown `type` fails with an actionable error
  listing available types.
- **Dependency order = declaration order.** The config order IS the dependency order.
  `runSync` does not reorder. The canonical order (`git → source → enrichments`) is
  documented but not enforced — users may override it deliberately.
- **Cross-refs run last, unconditionally on success.** After all sources complete
  (or after the first failure in fail-fast mode — in which case cross-refs are
  skipped), `resolveReferences` runs once globally over the union of episodes written
  by this run. The `--no-cross-refs` flag disables it for debugging. There is no
  config-level disable: the whole point of sync is a connected graph.
- **Storage**: Config is on disk only. No tables added to `.engram`. Per-run summary is
  ephemeral (stdout); durable history lives in existing `ingestion_runs`.
- **Security**: Tokens never stored in config — only `tokenEnv` env var names. If the
  named env var is unset at sync time, validation fails before any network calls.

## Integration Points

- **v2 adapter contract** (`packages/engram-core/src/ingest/adapter.ts`): `runSync`
  builds `EnrichOptions` using `AuthCredential` + `scope` + `scopeSchema.validate()`
  exactly as the current `ingest enrich` CLI does (see `packages/engram-cli/src/
  ingest/auth.ts` → `buildAuthCredential`). No adapter-side changes required.
- **Plugin discovery** (`packages/engram-core/src/plugins/`): Uses `discoverPlugins()`
  + `loadManifest()` + `loadExecutablePlugin` / `loadJsModulePlugin` to resolve plugin
  sources. If a config names a plugin `type` that isn't installed, the error names the
  missing plugin and exits before any adapter runs.
- **Cross-ref resolver** (`packages/engram-core/src/ingest/cross-ref/`, landed in
  #197): Invoked once at end of run via its public `resolveReferences` entry point.
- **Individual `ingest` subcommands are unchanged.** `engram ingest git`,
  `ingest source`, `ingest enrich <adapter>` continue to work for one-off / scripting
  use. `sync` is a thin orchestrator over the same APIs.
- **CI usage**: `engram sync --format json` emits a summary object; exit code
  reflects aggregate status (`0` = all ok, `1` = one or more source failures, `2` =
  config validation error).

## Acceptance Criteria

- [ ] `runSync(graph, config, opts)` implemented in
      `packages/engram-core/src/sync/run.ts` and re-exported from the package index
- [ ] `engram sync` command registered in `packages/engram-cli/src/commands/sync.ts`;
      wired into the root `commander` program
- [ ] Config discovery resolves in documented order (flag → env → cwd → db-adjacent);
      missing config fails with a pointer to the example
- [ ] JSON schema validation via hand-rolled `SyncConfigValidationError` with field
      path + reason (no new dependency); unknown fields cause failure
- [ ] `version: 1` enforced; mismatched version fails with an upgrade-path message
- [ ] `name` uniqueness enforced across sources within a config
- [ ] Built-in `type` values resolve directly (`git`, `source`, `github`); plugin
      `type` values resolve via `discoverPlugins()`; unknown types list the available
      set in the error message
- [ ] Auth env vars resolved and checked before any network calls; missing env var
      fails validation for that source
- [ ] Sources run in declaration order; cross-ref resolver runs exactly once at end
      (unless `--no-cross-refs`)
- [ ] Fail-fast default: first source failure aborts the run, skips remaining sources
      and cross-refs, exits non-zero
- [ ] `--continue-on-error`: runs all sources, reports aggregate, exits non-zero if
      any failed; cross-refs still run over the subset that succeeded
- [ ] `--only <name>[,<name>...]`: runs only the named subset in declaration order;
      cross-refs run over that subset unless disabled
- [ ] `--dry-run`: prints the resolved plan (source names, types, adapter resolution,
      env var presence) without executing; exits zero on valid config
- [ ] `--format json`: emits a `SyncResult` object with per-source counts
      (episodes/entities/edges), elapsed ms, status, and top-level exit reason
- [ ] Exit codes: `0` success, `1` at least one source failed, `2` config validation
      error, `3` no discoverable config
- [ ] Default human-readable output is a table of per-source results + aggregate line

## Edge Cases

- **Config missing.** Exit `3` with a message pointing at `docs/` example and
  `--config` flag. No partial work.
- **Config JSON parse error.** Exit `2` with line/column from the parser.
- **Schema validation error.** Exit `2` with field path (e.g.
  `sources[2].auth.tokenEnv: expected string`) — all errors reported, not just first.
- **Duplicate `name`.** Validation error naming both source indices.
- **Unknown `type`.** Validation error listing known built-in types and currently
  discovered plugin types.
- **Plugin declared but not installed.** Error names the `type`, the expected plugin
  id, and (if possible) a hint from the plugin registry.
- **Auth env var unset.** Validation error before execution; names the source and the
  env var.
- **`--only` names a source not in config.** Exit `2` with the set of known names.
- **All sources skipped (empty `--only` match after filter).** Exit `2` — the user
  asked for work that doesn't exist.
- **First source fails, fail-fast.** Remaining sources skipped, cross-refs skipped,
  exit `1`; JSON output shows completed/skipped/failed status per source.
- **Partial failure, `--continue-on-error`.** Cross-refs still run over the episodes
  the successful sources wrote. Exit `1`.
- **Cross-ref step itself fails.** Treated as a source-level failure; sources already
  committed remain committed. Exit `1`.
- **Two `github` sources against different scopes.** Both run; their
  `ingestion_runs` rows are distinct by `(source_type, scope)` — already true.
- **Config on disk but graph has never been opened.** `openGraph()` creates it; the
  run proceeds.
- **Concurrent `sync` runs.** Not supported in v0.1. SQLite locking handles
  correctness; competing runs may fail noisily on write contention. Document, don't
  engineer around.

## Tests

Core (`packages/engram-core/test/sync/`):
- [ ] Unit: `runSync` executes sources in declaration order
- [ ] Unit: fail-fast aborts on first failure; remaining sources report `skipped`
- [ ] Unit: `--continue-on-error` runs all sources; aggregate status reflects failures
- [ ] Unit: cross-ref step runs on success; skipped on fail-fast failure
- [ ] Unit: `SyncResult` totals match sum of per-source counts
- [ ] Integration: sync over a fixture repo + fake GitHub adapter produces the same
      graph as running `ingest git` + `ingest enrich github` by hand

CLI (`packages/engram-cli/test/commands/sync.test.ts`):
- [ ] Config discovery order (flag > env > cwd > db-adjacent)
- [ ] Missing config → exit `3` with pointer message
- [ ] JSON parse error → exit `2`
- [ ] Schema errors report all violations with field paths
- [ ] Duplicate `name` rejected
- [ ] Unknown `type` lists known types
- [ ] Missing auth env var → exit `2` before execution
- [ ] `--only` selects subset; unknown name → exit `2`
- [ ] `--dry-run` prints plan, exits `0`, writes nothing
- [ ] `--format json` output matches documented `SyncResult` shape
- [ ] `--no-cross-refs` skips the resolver step
- [ ] Plugin `type` resolved via `discoverPlugins()` fixture

## Documentation

- [ ] `docs/internal/specs/sync-orchestration.md` (new) — config schema, discovery
      order, adapter resolution, failure semantics, cross-ref integration, exit codes
- [ ] `docs/` example `.engram.config.json` committed at repo root or under
      `docs/examples/` — referenced by the error message for missing config
- [ ] CLAUDE.md "File Organization" — add `packages/engram-core/src/sync/` row
- [ ] CLAUDE.md "Key Files" — add row for `packages/engram-core/src/sync/run.ts`
- [ ] CLAUDE.md "Ingestion Architecture" — note that `engram sync` is the preferred
      entry point; individual `ingest` subcommands remain for scripting
- [ ] JSDoc on `runSync`, `SyncConfig`, `SyncResult`, `SyncConfigValidationError`

## Labels

`feature`, `area/cli`, `area/ingest`, `spec` (after the spec doc lands), `priority/medium`,
`backlog/ready` (after this refinement is merged into the issue body)

## Related

- **Unblocked by** #197 (cross-ref resolver — **closed**), #196 (Google Workspace —
  **closed**), #191 (Gerrit — **closed**). The "defer until 2–3 adapters have landed"
  gate has tripped.
- **Integrates with** #204 (plugin loading spec) — plugin adapters are resolved via
  `discoverPlugins()`.
- **Complements** #205 (CLI consumes v2 adapter options — **closed**). `sync` reuses
  `buildAuthCredential` and the same option-building logic.
- **Benefits from future** #193 (GitLab), #194 (Jira), #195 (Linear), #192 (Buganizer)
  — each becomes a one-line addition to the user's config.

rnwolfe:
Closes #203

## Summary

- Adds `engram sync` CLI command that reads `.engram.config.json` and runs all configured ingesters in declaration order, then the cross-ref resolver
- Adds `runSync()` + `validateSyncConfig()` to `engram-core` public API
- Introduces `packages/engram-core/src/sync/` with `types.ts`, `errors.ts`, `run.ts`, `index.ts`
- Adds example config at `docs/examples/.engram.config.json` and spec at `docs/internal/specs/sync-orchestration.md`
- Updates `CLAUDE.md` with file organization, key files, and preferred entry point note

## Acceptance Criteria

| Criterion | Status |
|-----------|--------|
| `runSync()` implemented and exported from `packages/engram-core/src/index.ts` | ✅ |
| `engram sync` command wired into root commander program | ✅ |
| Config discovery resolves in documented order; missing config exits `3` with pointer message | ✅ |
| `version: 1` enforced; mismatch fails with upgrade-path message | ✅ |
| Unknown fields at any level cause validation failure (fail-closed) | ✅ |
| `name` uniqueness enforced across sources | ✅ |
| Built-in types resolve directly; plugin types via `discoverPlugins()`; unknown types list available set | ✅ |
| Auth env vars checked before any network calls; missing env var = exit `2` naming source + var | ✅ |
| Sources run in declaration order; cross-refs run exactly once at end (unless `--no-cross-refs`) | ✅ |
| Fail-fast: first failure aborts, skips remaining, exits `1` | ✅ |
| `--continue-on-error`: all sources run, cross-refs over successful subset, exits `1` if any failed | ✅ |
| `--only`: runs named subset in declaration order; unknown name → exit `2` listing known names | ✅ |
| `--dry-run`: prints resolved plan, exits `0`, writes nothing | ✅ |
| `--format json`: emits `SyncResult` with per-source counts (episodes/entities/edges), elapsed ms, status | ✅ |
| Default output: table of per-source results + aggregate line | ✅ |
| Tests: unit tests for `runSync` ordering, fail-fast, continue-on-error, cross-ref skip; validation error tests | ✅ (29 tests) |
| `docs/internal/specs/sync-orchestration.md` — config schema, discovery, failure semantics, exit codes | ✅ |
| `docs/examples/.engram.config.json` — example config | ✅ |
| CLAUDE.md updated: file organization, key files, ingestion architecture note | ✅ |

## Test plan

- [ ] `bun test packages/engram-core/test/sync/` — all 29 tests pass
- [ ] `bun run build` — builds cleanly
- [ ] `bun run lint` — no errors
- [ ] `engram sync --help` — shows full help with discovery order, examples, exit codes
- [ ] `engram sync --dry-run --config docs/examples/.engram.config.json` — prints plan, exits 0
- [ ] Missing config → exits 3 with pointer message
- [ ] Invalid config (unknown field) → exits 2 with all field errors listed

### DIFF
--- a/packages/engram-core/src/sync/run.ts
+++ b/packages/engram-core/src/sync/run.ts
@@ -1,0 +1,570 @@
+/**
+ * sync/run.ts — config-driven multi-source ingestion orchestrator.
+ *
+ * Runs all configured sources in declaration order, then runs the cross-ref
+ * resolver. Replaces hand-crafted shell sequences of individual ingest commands.
+ */
+
+import * as path from "node:path";
+import type { EngramGraph } from "../format/index.js";
+import type { AuthCredential, EnrichmentAdapter } from "../ingest/adapter.js";
+import { resolveReferences } from "../ingest/cross-ref/index.js";
+import { ingestGitRepo } from "../ingest/git.js";
+import { ingestSource } from "../ingest/source/index.js";
+import {
+  bundledPluginsRoot,
+  discoverPlugins,
+  loadExecutablePlugin,
+  loadJsModulePlugin,
+} from "../plugins/index.js";
+import { SyncConfigValidationError, SyncSourceError } from "./errors.js";
+import type {
+  RunSyncOpts,
+  SourceResult,
+  SyncAuthConfig,
+  SyncConfig,
+  SyncResult,
+  SyncSource,
+} from "./types.js";
+import { resolveSyncAuth } from "./types.js";
+
+// ---------------------------------------------------------------------------
+// Built-in adapter types
+// ---------------------------------------------------------------------------
+
+const BUILT_IN_TYPES = ["git", "source", "github", "google_workspace"] as const;
+
+// ---------------------------------------------------------------------------
+// Config validation
+// ---------------------------------------------------------------------------
+
+const ALLOWED_AUTH_KINDS = [
+  "none",
+  "bearer",
+  "basic",
+  "service_account",
+  "oauth2",
+] as const;
+
+const ALLOWED_SOURCE_FIELDS = new Set([
+  "name",
+  "type",
+  "scope",
+  "path",
+  "root",
+  "auth",
+]);
+
+const ALLOWED_AUTH_FIELDS: Record<string, Set<string>> = {
+  none: new Set(["kind"]),
+  bearer: new Set(["kind", "tokenEnv"]),
+  basic: new Set(["kind", "usernameEnv", "secretEnv"]),
+  service_account: new Set(["kind", "keyJsonEnv"]),
+  oauth2: new Set(["kind", "tokenEnv", "scopesEnv"]),
+};
+
+/**
+ * Validate a raw JSON object against the SyncConfig schema.
+ * Collects ALL errors before throwing, so the user can fix everything at once.
+ *
+ * Rules (fail-closed):
+ * - `version: 1` required
+ * - Every source: `name` (unique), `type` — no unknown fields
+ * - Auth uses `SyncAuthConfig` union — no unknown fields per kind
+ */
+export function validateSyncConfig(raw: unknown): SyncConfig {
+  if (typeof raw !== "object" || raw === null || Array.isArray(raw)) {
+    throw new SyncConfigValidationError([
+      { field: "(root)", reason: "must be a JSON object" },
+    ]);
+  }
+
+  const obj = raw as Record<string, unknown>;
+  const failures: Array<{ field: string; reason: string }> = [];
+
+  // Check for unknown top-level fields
+  const ALLOWED_TOP_FIELDS = new Set(["version", "sources"]);
+  for (const key of Object.keys(obj)) {
+    if (!ALLOWED_TOP_FIELDS.has(key)) {
+      failures.push({
+        field: key,
+        reason: `unknown top-level field (allowed: ${Array.from(ALLOWED_TOP_FIELDS).join(", ")})`,
+      });
+    }
+  }
+
+  // Validate version
+  if (!("version" in obj)) {
+    failures.push({ field: "version", reason: "required field missing" });
+  } else if (obj.version !== 1) {
+    failures.push({
+      field: "version",
+      reason: `must be 1 (got ${JSON.stringify(obj.version)}). If you have a newer config format, upgrade engram to the matching version.`,
+    });
+  }
+
+  // Validate sources
+  if (!("sources" in obj)) {
+    failures.push({ field: "sources", reason: "required field missing" });
+  } else if (!Array.isArray(obj.sources)) {
+    failures.push({ field: "sources", reason: "must be an array" });
+  } else {
+    const seenNames = new Set<string>();
+
+    for (let i = 0; i < obj.sources.length; i++) {
+      const src = obj.sources[i];
+      const prefix = `sources[${i}]`;
+
+      if (typeof src !== "object" || src === null || Array.isArray(src)) {
+        failures.push({ field: prefix, reason: "must be an object" });
+        continue;
+      }
+
+      const srcObj = src as Record<string, unknown>;
+
+      // Check for unknown source fields
+      for (const key of Object.keys(srcObj)) {
+        if (!ALLOWED_SOURCE_FIELDS.has(key)) {
+          failures.push({
+            field: `${prefix}.${key}`,
+            reason: `unknown field (allowed: ${Array.from(ALLOWED_SOURCE_FIELDS).join(", ")})`,
+          });
+        }
+      }
+
+      // name
+      if (!("name" in srcObj)) {
+        failures.push({ field: `${prefix}.name`, reason: "required" });
+      } else if (typeof srcObj.name !== "string" || !srcObj.name) {
+        failures.push({
+          field: `${prefix}.name`,
+          reason: "must be a non-empty string",
+        });
+      } else if (seenNames.has(srcObj.name)) {
+        failures.push({
+          field: `${prefix}.name`,
+          reason: `duplicate source name '${srcObj.name}' — names must be unique`,
+        });
+      } else {
+        seenNames.add(srcObj.name as string);
+      }
+
+      // type
+      if (!("type" in srcObj)) {
+        failures.push({ field: `${prefix}.type`, reason: "required" });
+      } else if (typeof srcObj.type !== "string" || !srcObj.type) {
+        failures.push({
+          field: `${prefix}.type`,
+          reason: "must be a non-empty string",
+        });
+      }
+
+      // scope (optional string)
+      if ("scope" in srcObj && typeof srcObj.scope !== "string") {
+        failures.push({ field: `${prefix}.scope`, reason: "must be a string" });
+      }
+
+      // path (optional string)
+      if ("path" in srcObj && typeof srcObj.path !== "string") {
+        failures.push({ field: `${prefix}.path`, reason: "must be a string" });
+      }
+
+      // root (optional string)
+      if ("root" in srcObj && typeof srcObj.root !== "string") {
+        failures.push({ field: `${prefix}.root`, reason: "must be a string" });
+      }
+
+      // auth (optional)
+      if ("auth" in srcObj) {
+        const authFailures = validateAuthConfig(srcObj.auth, `${prefix}.auth`);
+        failures.push(...authFailures);
+      }
+    }
+  }
+
+  if (failures.length > 0) {
+    throw new SyncConfigValidationError(failures);
+  }
+
+  return obj as unknown as SyncConfig;
+}
+
+function validateAuthConfig(
+  auth: unknown,
+  fieldPrefix: string,
+): Array<{ field: string; reason: string }> {
+  const failures: Array<{ field: string; reason: string }> = [];
+
+  if (typeof auth !== "object" || auth === null || Array.isArray(auth)) {
+    failures.push({ field: fieldPrefix, reason: "must be an object" });
+    return failures;
+  }
+
+  const authObj = auth as Record<string, unknown>;
+
+  if (!("kind" in authObj)) {
+    failures.push({ field: `${fieldPrefix}.kind`, reason: "required" });
+    return failures;
+  }
+
+  const kind = authObj.kind;
+  if (
+    !ALLOWED_AUTH_KINDS.includes(kind as (typeof ALLOWED_AUTH_KINDS)[number])
+  ) {
+    failures.push({
+      field: `${fieldPrefix}.kind`,
+      reason: `must be one of [${ALLOWED_AUTH_KINDS.join(", ")}], got ${JSON.stringify(kind)}`,
+    });
+    return failures;
+  }
+
+  const kindStr = kind as string;
+  const allowedFields = ALLOWED_AUTH_FIELDS[kindStr];
+  if (allowedFields) {
+    for (const key of Object.keys(authObj)) {
+      if (!allowedFields.has(key)) {
+        failures.push({
+          field: `${fieldPrefix}.${key}`,
+          reason: `unknown field for auth kind '${kindStr}' (allowed: ${Array.from(allowedFields).join(", ")})`,
+        });
+      }
+    }
+  }
+
+  // Kind-specific required fields
+  switch (kindStr) {
+    case "bearer":
+      if (typeof authObj.tokenEnv !== "string" || !authObj.tokenEnv) {
+        failures.push({
+          field: `${fieldPrefix}.tokenEnv`,
+          reason:
+            "required for bearer auth — name of the env var holding the token",
+        });
+      }
+      break;
+    case "basic":
+      if (typeof authObj.usernameEnv !== "string" || !authObj.usernameEnv) {
+        failures.push({
+          field: `${fieldPrefix}.usernameEnv`,
+          reason: "required for basic auth",
+        });
+      }
+      if (typeof authObj.secretEnv !== "string" || !authObj.secretEnv) {
+        failures.push({
+          field: `${fieldPrefix}.secretEnv`,
+          reason: "required for basic auth",
+        });
+      }
+      break;
+    case "service_account":
+      if (typeof authObj.keyJsonEnv !== "string" || !authObj.keyJsonEnv) {
+        failures.push({
+          field: `${fieldPrefix}.keyJsonEnv`,
+          reason: "required for service_account auth",
+        });
+      }
+      break;
+    case "oauth2":
+      if (typeof authObj.tokenEnv !== "string" || !authObj.tokenEnv) {
+        failures.push({
+          field: `${fieldPrefix}.tokenEnv`,
+          reason: "required for oauth2 auth",
+        });
+      }
+      break;
+  }
+
+  return failures;
+}
+
+// ---------------------------------------------------------------------------
+// Auth env var pre-flight check
+// ---------------------------------------------------------------------------
+
+/**
+ * Check that all env vars referenced in auth configs are present.
+ * Returns a list of {sourceName, envVar} pairs that are missing.
+ */
+function checkAuthEnvVars(
+  sources: SyncSource[],
+): Array<{ sourceName: string; envVar: string }> {
+  const missing: Array<{ sourceName: string; envVar: string }> = [];
+
+  for (const src of sources) {
+    if (!src.auth) continue;
+    const result = resolveSyncAuth(src.auth);
+    if ("missing" in result) {
+      missing.push({ sourceName: src.name, envVar: result.missing });
+    }
+  }
+
+  return missing;
+}
+
+// ---------------------------------------------------------------------------
+// Adapter resolution
+// ---------------------------------------------------------------------------
+
+async function resolveAdapter(
+  src: SyncSource,
+  cwd: string,
+): Promise<EnrichmentAdapter> {
+  // Only network adapters (non-git, non-source) need a loader
+  const type = src.type;
+
+  if (type === "github") {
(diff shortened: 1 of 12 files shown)


packages/engram-core/src/sync/types.ts
```diff
@@ -0,0 +1,158 @@
...
+    case "bearer": {
+      const token = process.env[authConfig.tokenEnv];
+      if (!token) return { missing: authConfig.tokenEnv };
+      return { credential: { kind: "bearer", token } };
+    }
+
+    case "basic": {
+      const username = process.env[authConfig.usernameEnv];
+      const secret = process.env[authConfig.secretEnv];
+      if (!username) return { missing: authConfig.usernameEnv };
```
rnwolfe: **[B2] `basic` auth only reports first missing env var**

```ts
case "basic": {
  const username = process.env[authConfig.usernameEnv];
  const secret = process.env[authConfig.secretEnv];
  if (!username) return { missing: authConfig.usernameEnv };
  if (!secret) return { missing: authConfig.secretEnv }; // never reached if username is also missing
```

If both `usernameEnv` and `secretEnv` are absent, the function returns `{ missing: usernameEnv }` and `secretEnv` is never checked. The user fixes the first var, reruns, and hits a second error.

Options:
1. Change the return type to `{ missing: string[] }` and return all missing vars together.
2. Add a separate `checkPresence(authConfig)` function that collects all missing vars without resolving values, and call that from `checkAuthEnvVars` instead of reusing `resolveSyncAuth`.

packages/engram-core/src/sync/run.ts
```diff
@@ -0,0 +1,570 @@
...
+
+// ---------------------------------------------------------------------------
+// Auth env var pre-flight check
+// ---------------------------------------------------------------------------
+
+/**
+ * Check that all env vars referenced in auth configs are present.
+ * Returns a list of {sourceName, envVar} pairs that are missing.
+ */
+function checkAuthEnvVars(
```
rnwolfe, line 47: **[B3] `SyncSourceError` constructed with empty `sourceType`**

```ts
throw new SyncSourceError(
  missingEnvVars[0].sourceName,
  "",            // ← empty string produces "source 'X' () failed: …"
  new Error(...)
);
```

The source type is available — add it to the `{ sourceName, envVar }` objects returned by `checkAuthEnvVars` and pass it here. Alternatively, use a plain `Error` here since this is a pre-flight config error, not a source execution error, and \`SyncSourceError\` semantics ("source failed during execution") don't quite fit.

packages/engram-core/src/sync/run.ts
```diff
@@ -0,0 +1,570 @@
...
+async function resolveAdapter(
+  src: SyncSource,
+  cwd: string,
+): Promise<EnrichmentAdapter> {
+  // Only network adapters (non-git, non-source) need a loader
+  const type = src.type;
+
+  if (type === "github") {
+    const { GitHubAdapter } = await import("../ingest/adapters/github.js");
+    return new GitHubAdapter();
```
rnwolfe, line 76: **[B4] `dryRun` silently dropped for `git` built-in**

```ts
const result = await ingestGitRepo(graph, repoPath, {
  dryRun,
} as Parameters<typeof ingestGitRepo>[2]);
```

`GitIngestOpts` does not have a `dryRun` field — the cast hides this from the type system but the flag has no effect. A programmatic caller using `runSync(graph, config, { dryRun: true })` will trigger a live git ingest against the repo.

Fix options:
- Add `dryRun?: boolean` to `GitIngestOpts` and early-return in `ingestGitRepo` when set.
- Or use the dry-run shortcut already in `runSync` (the `dryRun ? { status: "success", ... }` branch) and remove the `dryRun` pass-through to the adapter entirely for built-ins.

packages/engram-core/src/sync/run.ts
```diff
@@ -0,0 +1,570 @@
...
+  if (manifest.transport === "executable") {
+    return loadExecutablePlugin(
+      found.dir,
+      manifest,
+    ) as unknown as EnrichmentAdapter;
+  }
+
+  throw new Error(
+    `Plugin '${type}' has unsupported transport '${manifest.transport}'`,
+  );
```
rnwolfe, line 124: **[N2] Scope not validated when absent**

```ts
if (adapter.scopeSchema && src.scope) {
  adapter.scopeSchema.validate(src.scope);
}
```

Calling `validate` only when `src.scope` is truthy means a missing required scope is silently accepted here. The adapter will fail later (e.g. GitHub adapter will throw when constructing the API URL) with a less user-friendly error.

Consider:
```ts
if (adapter.scopeSchema) {
  adapter.scopeSchema.validate(src.scope); // let adapter decide if undefined is valid
}
```
This lets each adapter's `validate()` throw a clear error for missing scope.

rnwolfe (commented):
## Code Review: `engram sync` orchestration (PR #266)

**Overall**: The architecture is solid — fail-closed validation, env-var-only auth refs, ordered execution, and comprehensive happy-path tests. Four issues need fixing before merge (2 blocking, 2 non-blocking but important). Four informational notes follow.

### Blocking

**B1** — `run.ts` is 570 lines, exceeding the 500-line limit in CLAUDE.md. Split `validateSyncConfig` + `validateAuthConfig` into `validate.ts` (they're already self-contained) and keep `run.ts` for the orchestration logic.

**B2** — `resolveSyncAuth` returns `{ missing: usernameEnv }` for `basic` auth without ever checking `secretEnv`. If both vars are absent the second missing var is silently dropped. The pre-flight report will say "set MYUSER_ENV" and the user will set it, rerun, and get a *second* error about `MYSECRET_ENV`. All required env vars for a source should be returned together. `resolveSyncAuth` returns a single `missing` string by design — either change it to `missing: string[]` or add a separate `checkAuthEnvVars` pass that collects them independently of `resolveSyncAuth`.

**B3** — `SyncSourceError` is constructed with empty string for `sourceType` in the pre-flight block in `runSync`. This produces the message `source 'X' () failed: …`. Either pass the actual source type (available from the `missingEnvVars` array — add it there) or use a different error class for the pre-flight case.

**B4** — `GitIngestOpts` has no `dryRun` field. The cast `{dryRun} as Parameters<typeof ingestGitRepo>[2]` silently drops the flag — the git adapter runs a live ingest during `runSync({ dryRun: true })` programmatic calls. Either add `dryRun` to `GitIngestOpts` and honour it in the adapter, or document that the programmatic `dryRun` flag is a no-op for the `git` built-in (and adjust the dry-run `SourceResult` shortcut in `runSync` to cover it instead).

### Non-blocking

**N1** — Missing auth env var pre-flight throws `SyncSourceError` and exits 2 — correct. But test coverage for this is only in the core layer (one test). The CLI's exit-2 path for pre-flight, the `--format json` output shape, the `--dry-run` printed plan, and the `--only` unknown-name exit-2 branch are all untested. These are in the CLI layer so they'd need a process-spawn integration test or at least a test of the helper functions (`discoverConfigPath`, `printTable`).

**N2** — Adapter scope is validated only when `src.scope` is truthy (`if (adapter.scopeSchema && src.scope) { adapter.scopeSchema.validate(src.scope) }`). For adapters that *require* a scope (e.g. GitHub requires `owner/repo`), a missing `scope` field is silently accepted by the orchestrator and the adapter itself will fail later with a less clear error. Consider calling `adapter.scopeSchema.validate(src.scope)` unconditionally and letting the adapter's `validate()` throw on `undefined`.

### Informational

**I1** — The `import("@engram/plugin-google-workspace" as string)` cast bypasses TypeScript's module resolution type-checking. This is intentional for a dynamic optional dependency, but it also means the `GoogleWorkspaceAdapter` return type is `any`, and any interface drift between the plugin and `EnrichmentAdapter` will surface at runtime, not build time. A thin `// @ts-expect-error dynamic optional plugin` comment would make the intent clearer than the cast.

**I2** — `checkAuthEnvVars` iterates sources and calls `resolveSyncAuth`, which *resolves* the credential (reads env vars) just to discard the result if present. For `service_account` this means the entire JSON key blob is read and then thrown away during the pre-flight pass. It's not a correctness issue, but it's wasteful and couples the pre-flight check to the auth resolver. A dedicated `missingAuthEnvVars(authConfig)" helper that only *checks* presence (not reads value) would be cleaner.

**I3** — `root` and `path` are both accepted for both the `git` and `source` adapters with different precedence (`root` wins for `source`, `path` wins for `git`). The spec doc says `root` is an alias for `path` specifically for `source`. This cross-contamination is unlikely to bite in practice but could confuse users who put `root` in a `git` source entry and wonder why it worked.

rnwolfe:
Fixed all 4 blocking issues: extracted validateSyncConfig into validate.ts (B1), fixed basic auth to collect all missing env vars before throwing (B2), fixed SyncSourceError to use src.type instead of empty string (B3), fixed --dry-run to skip adapter calls entirely at orchestrator level (B4). Also addressed N2: scope presence now validated for adapters that require it.

PR #273, opened 2026-04-24
fix(viz): sidebar toggle for mobile + layout fix for revealed nodes

rnwolfe:
## Summary

- **Sidebar toggle**: adds a ☰ button to the toolbar. On desktop it collapses the sidebar to 0 width (smooth CSS transition). On mobile (≤768px) the sidebar becomes a fixed overlay that slides in from the left, with a semi-transparent backdrop to dismiss it. Mobile starts with the sidebar closed so the full graph is immediately visible.
- **CoSE layout stacking fix**: `nodeRepulsion` was 2048 — less than half the algorithm's default of 4500 — so nodes didn't push apart at all. Raised to 400000 along with tuning `idealEdgeLength`, `numIter`, `initialTemp`, and `coolingFactor`.
- **Re-layout on reveal**: nodes in `DEFAULT_HIDDEN_ENTITY_TYPES` (file, module, symbol) are hidden before the initial layout, so they never get positions assigned. When a user checks a previously-hidden node type, the layout now re-runs immediately so newly-visible nodes are spread out instead of stacking at (0,0).

## Test plan

- [ ] Desktop: ☰ collapses/expands sidebar without disturbing graph
- [ ] Mobile: sidebar starts closed; ☰ slides it in; tapping backdrop closes it
- [ ] Initial graph load: nodes are spread out, not stacked
- [ ] Toggling "file" or "module" on: nodes appear spread out (not piled at origin)
- [ ] Toggling a type off: graph stays in place, no unnecessary re-layout
- [ ] `bun run build` passes


rnwolfe (commented):
## Response to review comments — commit 5c1b3cd

**CSS — collapsed sidebar still keyboard-focusable**
Added `visibility: hidden` and `pointer-events: none` to `.collapsed`. Also managing `inert` and `aria-hidden` in JS (`setSidebarAccessibility`) so the collapsed sidebar is fully out of the tab order and accessibility tree.

**JS — `openSidebar` class leak across breakpoints**
`openSidebar` now only adds `sidebar-open` on mobile / removes `collapsed` on desktop. `closeSidebar` always removes `sidebar-open` (cleans up regardless of breakpoint). Added a `resize` listener that strips both classes and reconciles state when crossing 768px.

**JS — focus trap in collapsed/closed sidebar**
`setSidebarAccessibility(open)` toggles `aria-hidden` and `inert`/removes `inert` on the sidebar element. Called from `openSidebar`, `closeSidebar`, the initial-state block, and the resize handler.

**HTML — `aria-expanded` defaults to `true` but sidebar starts closed on mobile**
Changed to `aria-expanded="false"` in the HTML. JS sets it to `"true"` at init on desktop via `setSidebarAccessibility(true)`.

**filters.ts — `onReveal` fires for any checkbox check, not just default-hidden types**
Gate is now `if (defaultHidden.has(value)) onReveal?.()` so re-layout only triggers when a type that was hidden by default is first revealed.

**graph.ts — CoSE work scales poorly on large graphs**
Extracted `COSE_BASE` constants and made `runCoseLayout` accept `{ animate? }`. `numIter` now scales as `Math.min(2500, Math.max(500, Math.round(5000 / n)))` — large graphs get fewer iterations. Filter-triggered re-layouts pass `animate: false` for instant repositioning; the manual "Re-layout" button still animates.

PR #274, opened 2026-04-24
feat: repo-lifecycle health — doctor, whats-new, update, freshness

rnwolfe:
## Summary

Four coordinated additions that help users (and their agents) understand whether their engram install and their graph are current:

- **`engram doctor`** gains three checks: `freshness` (days + commits-behind HEAD), `engine_version_drift` (nudge when the user upgraded but hasn't reviewed what changed), and `update_available` (GitHub releases, cached 24h).
- **`engram whats-new`** renders user-facing highlights from `docs/whats-new.json` filtered to versions the user hasn't acknowledged; bumps `last_seen_engine_version` on render so the drift nudge clears.
- **`engram update`** is the matching self-updater: `--check` probes the release API; the default mode downloads the matching OS/arch binary and atomic-renames it into `process.execPath`. Refuses dev installs.
- **`engram status`** surfaces per-source "N commits behind, X days ago" inline.

Underpinned by:
- `ENGINE_VERSION` synced to `0.2.0` across every workspace `package.json`, guarded by `scripts/check-versions.ts` wired into `bun run verify`.
- New `computeFreshness()` helper in `engram-core/health` — combines days-since with `git rev-list <cursor>..HEAD --count` for git sources.
- New `last_seen_engine_version` metadata + `markEngineVersionSeen()` helper (only `whats-new` bumps it — drift nudges persist until explicitly acknowledged).

The `/release` skill is updated to require the new discipline: pre-tag `ENGINE_VERSION` sync, a matching `docs/whats-new.json` entry, mandatory `Migration:` sub-bullets on breaking entries.

## Test plan

- [x] `bun run check:versions` passes
- [x] `bun run build` across all packages
- [x] `bun test` (1608 pass; 1 pre-existing flake in `project.test.ts` reproduces on `main`, unrelated)
- [x] `bun run lint` clean after `lint:fix`
- [x] Manual smoke: init graph → ingest git → `doctor` shows fresh → set `last_seen_engine_version=0.1.0` → `doctor` warns on drift → `whats-new` renders 0.2.0 highlights → `doctor` clears
- [x] `engram update --check` against live GitHub API
- [x] `engram update` refuses dev install (running via `bun src/cli.ts`) with exit 2

## Commits

1. `chore: sync ENGINE_VERSION across workspace and add drift guard`
2. `feat: track engine version drift and ingestion freshness in core`
3. `feat: surface freshness, version drift, and updates in doctor/status`
4. `feat: add engram whats-new and engram update commands`
5. `docs: require version sync and whats-new entries in release skill`


rnwolfe:
Thanks for the review — addressed all 8 findings in e753c63.

| # | Issue | Resolution |
|---|-------|------------|
| 1 | `compareSemver` prerelease comparison | Reimplemented per semver.org spec item 11 — numeric identifiers compared numerically, alphanumeric per ASCII, numeric < alphanumeric, fewer ids < more. Added 3 regression tests. |
| 2 | Update header claimed verification that wasn't happening | Corrected the docstring and the `--help` text to say authenticity rests on HTTPS/TLS only (no checksum/signature today). |
| 3 | Supply-chain risk from unverified self-update | Agreed — SHA256/signature verification is the right fix, but it needs release-workflow changes to publish checksums. Tracked as follow-up #275 so this PR stays scoped. The in-code comment now references it. |
| 4 | Git source with null cursor reported severity `unknown` | Fall back to day-only severity when cursor is absent; `severityFromCommits(null)` no longer poisons a fresh day verdict. Added a regression test. |
| 5 | `cursorLost` doc contradicted the implementation | Aligned the comment to reality — `cursorLost` is true when the SHA is missing or unreachable. |
| 6 | Dead `${severity === "fresh" ? "" : ""}` ternary in `buildReason` | Removed; dropped the now-unused `severity` parameter. |
| 7 | `status.ts` Map keyed on `sourceType` dropped scope, last (oldest) entry won | Rewrote `collectIngestion` to build the view directly from `freshness.sources`: pick the freshest per canonical type, and render non-canonical scopes one-per-line so plugin sources aren't collapsed. Added `cursor` to `SourceFreshness` to avoid a second DB query. |
| 8 | `whats-new` marked `last_seen` in JSON mode too (side-effects for scripts) | Gated marking on `opts.format === "text"` alongside the existing `--since`/`--all`/`--no-mark` opt-outs. Updated the JSON test to assert no mutation. |

`bun test` 1611 pass / 1 pre-existing flake (reproduces on `main`, unrelated). Build + lint clean.

